Issue #2, opened 2015-11-07
Database fails if not on the same server


groob:
Hi, 

I tried building zentral today with separate docker containers for Zentral and Postgres, and ran into an issue with connection. 
I believe this is because the connection string in postgres.py only accepts a username and password, so psycopg2 is defaulting to a unix socket connection, instead of TCP.

```python
    def __init__(self, config_d):
        super(EventStore, self).__init__(config_d)
        self._conn = psycopg2.connect("dbname=%(db_name)s user=%(user)s" % config_d)
        self._test_table()
```

https://github.com/zentralopensource/zentral/blob/master/zentral/core/stores/backends/postgres.py

Please change the connection string to accept host, port and password parameters in addition to name and user. 

Here's the full trace:

```
Environment:


Request Method: GET
Request URL: http://dev.groob.io:8000/

Django Version: 1.8.6
Python Version: 3.4.3
Installed Applications:
('django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'bootstrapform',
 'configuration',
 'zentral.contrib.inventory',
 'zentral.contrib.osquery',
 'zentral.contrib.santa')
Installed Middleware:
('django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'django.middleware.security.SecurityMiddleware')


Traceback:
File "/usr/local/lib/python3.4/dist-packages/django/core/handlers/base.py" in get_response
  119.                 resolver_match = resolver.resolve(request.path_info)
File "/usr/local/lib/python3.4/dist-packages/django/core/urlresolvers.py" in resolve
  365.             for pattern in self.url_patterns:
File "/usr/local/lib/python3.4/dist-packages/django/core/urlresolvers.py" in url_patterns
  401.         patterns = getattr(self.urlconf_module, "urlpatterns", self.urlconf_module)
File "/usr/local/lib/python3.4/dist-packages/django/core/urlresolvers.py" in urlconf_module
  395.             self._urlconf_module = import_module(self.urlconf_name)
File "/usr/lib/python3.4/importlib/__init__.py" in import_module
  109.     return _bootstrap._gcd_import(name[level:], package, level)
[...]

PR #3, opened 2015-11-07
Modular containers

groob:
Hi, 

I've made a few changes to the Dockerfile.
The main issue, is that a container should be as modular as possible(ideally, one process per container). 
This allows to compose, reuse and scale each component individually. 

I've made an attempt to split the django application from the rest - intending to run Postgres and Django as two separate containers. I ran into issue #2, but this container should work once that is resolved. 

We should strive to separate all services(elastic, kibana, redis etc) into independent services. 

Here are a few reference to more modular deployments of django apps
https://github.com/groob/docker-munkiwebadmin/tree/alpine
https://github.com/macadmins/sal


groob:
Hi @np5, 

The [redis](https://github.com/zentralopensource/zentral/blob/9cc3494e54da16a125b8eb41ae8b675d3c3532b3/zentral/core/queues/backends/redisq.py) backend is still hardcoded to localhost, making it not possible to run with redis outside the Docker container. 

Here too: https://github.com/zentralopensource/zentral/blob/9cc3494e54da16a125b8eb41ae8b675d3c3532b3/zentral/contrib/inventory/backends/base.py

np5:
In the [base inventory backend](https://github.com/zentralopensource/zentral/blob/9cc3494e54da16a125b8eb41ae8b675d3c3532b3/zentral/contrib/inventory/backends/base.py#L17-L19) and in the [redis queue backend](https://github.com/zentralopensource/zentral/blob/33d6918c69a10f65baef7dd172dc15a0645a64e3/zentral/core/queues/backends/redisq.py#L8-L10), we build an instance of the inventory or the store by passing them the relevant config part (config_d). We use the get method on the dict and only provide default values. You can override them by providing other values in the conf file.

So, you could use 

```json
"queues": {
  "backend": "zentral.core.queues.backends.redisq",
  "host": "10.0.17.3",
  "port": 7654,
  "db": 23
}
```

for the redis queue backend, and

```json
"zentral.contrib.inventory": {
  "backend": "zentral.contrib.inventory.backends.dummy",
  "redis_host": "10.0.17.3",
  "redis_post": 7654,
  "redis_db": 24
}
```

for the inventory app.

This is the standard construction for the different elements of zentral. Same with the actions. They receive the config dict when they are instantiated.

groob:
I'll try that, thank you. 

Should I edit the Dockerfile to remove all external dependencies and resend the pull request?

np5:
We want to maintain a docker container with everything in it so that everyone can start using zentral without having to manage multiple containers with the right connections between the containers and the proper start-up order. This is of course not acceptable in a production environment.

We could maybe rename this dockerfile as allinone.dockerfile and keep the bare minimum in the main Dockerfile.

The question is where do we stop dividing the containers. The backgroup zentral processes should run in a different container for example, if we want to be able to scale and deploy them independently from the django apps.

np5:
There is a new dev env based on docker-machine in master now.

PR #14, opened 2016-09-09
Changing OS version to set 'patch' to 0 if there is a ValueError. Thi…

jbaker10:
…s resolves a bug that prevented machines running 10.12 from enrolling successfully.


np5:
Thanks

PR #16, opened 2016-10-12
Update osquery enrollment package postinstall script to fix permissions on plist.

goobersmooch:
The current osquery enrollment package installs the "/Library/LaunchDaemons/com.facebook.osqueryd.plist" file with invalid permissions. This causes launchctl to fail to load "/Library/LaunchDaemons/com.facebook.osqueryd.plist" also causing the following error message in the system log: Caller specified a plist with bad ownership/permissions: path = /Library/LaunchDaemons/com.facebook.osqueryd.plist, caller = launchctl.89355.

I've added two lines of code to fix the permission issue.


goobersmooch:
After further investigation and much help from Henry, the root of the cause was the `umask`. To resolve this, run `umask 0022` just before a git checkout.

PR #17, opened 2016-10-20
Support request signing for Amazon IAM requests against AWS' ElasticSearch service

zbuc:
Add an `auth` dictionary for ElasticSearch to the `base.json` to enable signing.


np5:
Hi,

First of all, thanks for your contributions!

I did a quick review of your patches.

About the style:

* We order the imports by scope. First come the the standard library imports, then the external libraries imports, and then the Zentral imports.
* When the commits are about the same functionality (like 3ba1d8e and ece59f9), we like to have them rebased in a single commit.
* We like to have different pull requests when they address different issues (bc9e566 has nothing to do with Elasticsearch on AWS).

Then, a few remarks:

I could be wrong, but isn't it  possible to connect to the Elasticsearch service if an adhoc IAM role has been attached to the instance running zentral ? In that case, no need to store secrets in the zentral configuration. But maybe some situations require this kind of setup...

And, about the permissions on the plist. I really don't want to solve a umask problem in a Dockerfile. Open an issue and we can start a discussion about that and find the root of the problem.

np5:
I checked the AWS documentation and I was wrong. I will merge this if you prepare a new pull request with one commit, the correct import order and more explicit config attribute names.

```python
"elasticsearch": {
    "backend": "zentral.core.stores.backends.elasticsearch",     
    "kibana_base_url": "https://zentral.apfelwerk.io/kibana/",
    "index": "zentral-events",
    "servers": [
        "http://elastic:9200"
    ],
    "aws_auth": {
        "access_id": "AAAAAAAAAAAA",
        "access_key": "KKKKKKKKKKKK",
        "region": "eu-central-1"
    },
    "frontend": true
}
```
Sorry to be so picky!

zbuc:
I found this in the AWS docs, which I think might work, actually:
http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#instance-metadata-security-credentials

But it still may be worth exposing both methods of passing credentials.

np5:
See a24ea514ef8227f7

PR #29, opened 2017-04-10
Added password mask for JSS password input field

samuarl:
Added a mask to the password input field for the JSS Instance Form so its not visible whilst typed.


np5:
Thank you !

PR #33, opened 2017-05-15
Initial support for PuppetDB inventory

hjuutilainen:
So this is a start. There are a number of things I'm thinking about refactoring but thought it might be a good time to show the code and ask for feedback :)

Some questions:
- What's the proper way to create migration files after changing the model? I've been currently running `makemigrations` with docker but not sure if that's the way to do it.
- There's support for creating business units and groups from Puppet facts (nested or not) and this is entirely configurable via base.json but any opinions on what's the best way to describe these in config?
- There's support for defining additional custom facts to be displayed in the inventory detail view and these too are configurable via base.json. Same question for these as the above...
- This is currently doing just one call to the PuppetDB inventory API and then yielding nodes for everything in the database. Is this really the best way to go about this (other contrib clients seem to be doing the same thing)? PuppetDB has a pretty extensive query language so if contrib clients were given the serial numbers Zentral already knows about, they could request inventory for only those clients. Or am I just overthinking this?

Required configuration would look something like this:

```
{
    "backend": "zentral.contrib.inventory.clients.puppetdb",
    "puppetdb_host": "puppetdb.example.com",
    "puppetdb_ssl_key": "/zentral/conf/zentral-server.example.com.key.pem",
    "puppetdb_ssl_cert": "/zentral/conf/zentral-server.example.com.cert.pem",
    "puppetdb_ssl_ca": "/zentral/conf/puppet_ca.pem",
    "puppetdb_port": 8081,
    "puppetdb_protocol": "https",
    "puppetboard_enable": true,
    "puppetboard_url": "https://puppetboard.example.com",
    "group_fact_names": [
        { "fact_path": "my_custom_fact_site_id" },
        { "fact_path": "my_custom_fact_customer_environment" }
    ],
    "business_unit_fact_name": "my_custom_fact_customer",
    "facts": [
        {
            "fact_key": "os.hardware",
            "fact_display_name": "Hardware"
        },
        {
            "fact_key": "system_uptime.uptime",
            "fact_display_name": "Uptime"
        },
        {
            "fact_key": "kernel",
            "fact_display_name": "Kernel"
        }
    ]
}
```


np5:
I think we should use the [http report module](https://docs.puppet.com/puppet/latest/report.html#http) in puppet to post the reports to Zentral. We can then feed them in the storage/processing pipeline as Zentral events attached to a certain machine. This can be useful. We could trigger a notification in case of an error for example.

As the events arrive, we would call the puppetDB API to get all the machine facts and update the zentral inventory. We do [this](https://github.com/zentralopensource/zentral/blob/8ab1d9d11a65d2efdcc9d00d155ce338e506a669/zentral/contrib/jamf/workers.py#L117) in the zentral.contrib.jamf app. This way, we could avoid endlessly looping on all the machines in puppetDB.

np5:
Merged: 8ab1d9d

PR #41, opened 2017-06-06
Add support for loading feeds from disk

zbuc:
Spoke with Henry about this on MacAdmins Slack.

This adds support for loading feeds from disk as well as URL.

By default it configures the path to load feeds from as `/zentral/conf/feeds`:

```
class Feed(models.Model):
    url = models.URLField(unique=True, blank=True, null=True, default=None)
    path = models.FilePathField(unique=True, blank=True, null=True, default=None, path='/zentral/conf/feeds', recursive=True, match='*.json')
```

They are specified on the command-line using just the filename of the feed. Introspection on the model class is used to expand to the full path:

```
zentral@3dbbb4b7eae7:/zentral$ python server/manage.py add_probe_feed osquery-osx-attacks.json
2017-06-06 20:07:50,161 PID693 feeds WARNING Feed serializer <class 'zentral.contrib.osquery.feeds.PackSerializer'> errors
2017-06-06 20:07:50,162 PID693 feeds WARNING {'queries': ['This field is required.']}
2017-06-06 20:07:50,184 PID693 feeds WARNING Feed serializer <class 'zentral.contrib.osquery.feeds.PackSerializer'> errors
2017-06-06 20:07:50,185 PID693 feeds WARNING {'queries': ['This field is required.']}
Feed osquery-osx-attacks.json synced.
No changes.
```


np5:
I have read your code. It works. But I can't see how it would be useful.

Feeds, even when loaded from the disk, are not exactly like osquery packs. They cannot be used directly to update probes. They contain only the part of the probes that can be shared between Zentral deployments (the osquery queries, the event type filters, the santa rules, …). Once you import one of the probe they contain, you still eventually need to add an inventory filter, or an action to finish the probe configuration. These parts are not in the feeds because they are specific to a deployment (your tags, your actions, …)

Maybe it is only what you are interested in, and maybe your zentral instance is deployed on a network without unrestricted internet access, or without access to a web server where you could put those files.

What I think would be more useful, is a way of saving and loading the probes, with their actions and the filters, from a set of files. This way, the zentral configuration could be managed with git for example, and a zentral deployment could be totally automatized with puppet, chef or ansible.

zbuc:
> Maybe it is only what you are interested in, and maybe your zentral instance is deployed on a network without unrestricted internet access, or without access to a web server where you could put those files.

Yes, that is what we are dealing with right now. 

> What I think would be more useful, is a way of saving and loading the probes, with their actions and the filters, from a set of files. This way, the zentral configuration could be managed with git for example, and a zentral deployment could be totally automatized with puppet, chef or ansible.

This is our end goal. I think this PR is a useful step for now, at least for us, as we already have Chef syncing our feed json documents from Git to our Zentral servers.

mikemcdonald:
Any update to this potentially getting merged to master? I would prefer not having my feeds on a public endpoint so using files is the most straightforward solution without dealing with github deploy keys, etc.

Issue #22, opened 2017-02-20
Add support for Watchman Monitoring API v 2.5


YesThatAllen:
Watchman Monitoring's API has been upgraded, adding the ability to write [Asset ID & Descriptions](https://api.watchmanmonitoring.com/#update_computer) to Computer Records. It should be noted that Asset ID information written to the Computer Record via the API will become available on the disk of the monitored computer after its next hourly checkin. 

Also of interest for maintaining Group sync, a [Find or Create a Group](https://api.watchmanmonitoring.com/#find_or_create_group) endpoint has been added.

To access these features, the version in use by Zentral should be moved has been bumped to 2.5. The [Change Log](https://api.watchmanmonitoring.com/change-log/) describes the differences in v2.5, and the adjustments which are needed to complete the transition.

PR #43, opened 2017-06-16
Replace Watchman API 2.2 with 2.5

nasa42:
Closes https://github.com/zentralopensource/zentral/issues/22.


np5:
Thanks. Some other things were needed as well.

PR #50, opened 2017-09-11
osquery now uses plist (compliance queries)

mikemcdonald:
osquery> select filename, key, value from preferences p, file f WHERE ((p.path = '/Library/Preferences/com.apple.loginwindow.plist')) and f.path = p.path;
Error: no such column: p.path

preferences no longer has `path` key. osquery uses `plist` now


np5:
Thank you!

PR #57, opened 2018-02-13
update prometheus.yml to support prometheus 2

devx:
Currently the docker-compose image does not pin the prometheus version.  The current `prometheus.yml` is not compatible with the prometheus version.  This PR fixes the problem and ensure that everything comes up when you execute the command `docker-compose up -d`.


np5:
Thanks. This looks good. Could you fix the commit message, so that it shows that the problem was with the command line flags, not with prometheus.yml ?

PR #60, opened 2018-04-09
Airwatch Integration first version

janheise:
This is a first version to integrate Airwatch into Zentral


np5:
Thank you for your contribution.

Some quick notes.

* Since you have the airwatch accounts as models, no need to duplicate the configuration in `base.json` and use this kind of deprecated way to declare inventory workers. Declare your workers in `workers.py` in your contrib module. You could query the db in this module, to know which credentials to use.

* Airwatch [is providing events via Webhooks](https://my.air-watch.com/help/9.1/en/Content/Expert_Guides/System_Settings/System_Advanced_API_Notif.htm). You could build an endpoint, have it queue the events for a "preprocessor" (like [in the jamf module](https://github.com/zentralopensource/zentral/blob/4291cce542601c0ad36025f2092f4d1214bd5e7c/zentral/contrib/jamf/workers.py#L17)). The preprocessor could then query the Airwatch API to get a full device record if it is not already included in the webhook payload, generate the inventory update events, and some other important events, like "Device Compliance Status Change".

PR #61, opened 2018-05-02
Resolve: XSS bug in TAG field

nishitm:
Add whitelisting characters only to prevent XSS bug in "create new tag"
field. New tag can be created by going into the list of machines and
then select any perticular machine and click on "manage tags" or by
clicking direclty into "Inventories" and go to "Tags" fields


np5:
Hi @nishitm !

Thanks a lot for reporting this very embarrassing issue. aed1219 fixes it. I should be more careful when using the [Django `mark_safe`](https://docs.djangoproject.com/en/2.0/ref/utils/#django.utils.safestring.mark_safe) function. I have decided to escape the value of the tag name, instead of introducing some restriction at the form or model level.

PR #64, opened 2018-06-14
Fixes changes related to libssl-dev

samueljon:
* Changed libssl-dev to libssl1.0-dev
* Moved source building runs after libssl1.0-dev libraries have been installed.


np5:
Thanks

PR #81, opened 2019-05-09
Increase Cipher strength and protocol support.

johnmikep:
I use the Google-all-in-one deployment so the location is a bit different, but I think the tls.conf is the same as /etc/nginx/ssl/common.conf on the all in one. There was a note on the original commit that it was reverted to intermediate for 10.12 but at this point I think its safe to move forward from that. 

Also, added an addition to easily grab the serial number of a machine that triggered an event probe. This was useful for one-off backend actions that needed access to just the serial number, not the full event probe at zentral/core/events/templates/base_body.txt. This may be a lazy approach instead of parsing the probe, but for actions that don't deserve a full module build out it was helpful. A simple example is included below. 

```
class Action(BaseAction):
    def trigger(self, event, probe, action_config_d):
        action_config_d = action_config_d or {}
        serialNumber = event.get_notification_serial(probe)
        args = {"updateAction":"add","tags":"Quarantine","serials":serialNumber}
        url = self.config_d['webhook']
        payload = ""
        headers = {
            'API-Key': "",
            'Content-Type': "application/json",
            'cache-control': "no-cache",
            }
        
         requests.request("PUT", url, data=payload, headers=headers, params=args)
```


np5:
Thanks for your contribution. 

You can get the machine serial number from the event metadata: [`event.metadata.machine_serial_number`](https://github.com/zentralopensource/zentral/blob/17e80f0684f957b8b7086161e6ee69c46de722e7/zentral/core/events/base.py#L165).

johnmikep:
I swear I looked up and down for that exact line -__- 
Thank you for pointing that out.

PR #123, opened 2020-12-01
FOR DISCUSSION: add chrome principal user to machine snapshot

arubdesu:
I didn't flesh out the view server-side, may not be the best implementation (original has open TODO's) but worth a shot to at least discuss


np5:
It works. As it is, it will first try to get a Company portal user (stronger machine ↔ user signal), then a Google Chrome user. For Google Chrome, it will pick the most recently used profile among all the users profiles (not only `Default`).

We could implement a filter based on a domain name, we could even activate or deactivate some methods based on configuration:

```json
{
  "principal_user_detection": {
    "sources": ["google_chrome"],
    "domain": "zentral.pro"
  }
}
```

The postflight script could [get this bit of configuration at every run](https://github.com/zentralopensource/zentral/blob/0c38a6a3a3640f112b94b55afce1c27386c6a6d5/zentral/contrib/munki/osx_package/build.tmpl/root/usr/local/zentral/munki/zentral_postflight#L421-L423), and do the principal user detection based on it.

We could even make it configurable in the GUI, but that would be much more work.

PR #185, opened 2021-03-06
[Build Fix]Update docker copy commands

gavinelder:
Small Fix following on from https://github.com/zentralopensource/zentral/pull/171 which added PiP constraints and attempted to copy an additional file to the path of `.` sadly this is not supported and gives an error when trying to do the local docker-compose workflows. 

Error from linter. 

```
When using COPY with more than one source file, the destination must be a directory and end with a / or a \dockerfile-utils(23)
```


np5:
Thanks!

Somehow these errors are not triggered when using `DOCKER_BUILDKIT=1` and `COMPOSE_DOCKER_CLI_BUILD=1`, like in the [github workflow](https://github.com/zentralopensource/zentral/blob/e3202033e2f173aa2b738b6bad3921514947dcf1/.github/workflows/tests.yml#L21-L22). Probably an undefined BuildKit behaviour. These options also speed up the standard builds, by skipping the unecessary stages.

PR #219, opened 2021-03-16
Report munki full apps info & locally save tags

np5:
Add `apps_full_info_shard` for the munki app. This value if present is
sent to the postflight script, and will be use to decide if the full
application info (std + bundle display name and bundle version) will be
uploaded. If the option is absent, those info will not be uploaded. A
value between 1 and 100 will only activate this on some machines (%).

The job details API call has been moved to the preflight. The job
details are temporarily saved in the `/usr/local/zentral/munki` dir for
the postflight (to avoid a second API call). The preflight script looks
for the `tags` key in the job details and save them as `zentral_tags` in
the default `ConditionalItems.plist`. It is done in the preflight to
allow the tags to be used as soon as possible in during a munki run.


np5:
Writing directly to `ConditionalItems.plist` was a bad idea. We know save the tags fetched during preflight in `/usr/local/zentral/munki/facts.plist`. These are read either via a module for munki-facts or a conditional items script. The postinstall script [decides which one to install](https://github.com/zentralopensource/zentral/pull/219/commits/6a3dcabcd68ddd181b77eaba018295df0965064d#diff-16c0681f0e782021280f07e919e47fad13ef096b14d9512c5d9e42fd01467df1R139-R140).

PR #292, opened 2021-09-21
Add "excluded_*" attributes to santa rules

np5:
* `excluded_primary_users`
* `excluded_serial_numbers`
* `excluded_tags`

These take precedence over `primary_users`, `serial_numbers` and `tags`.


np5:
@sierra-hotel Maybe this will help with the "unique target rule per configuration" constraint. You could for example have one rule blocking a binary except for some users.

sierra-hotel:
Ooh I like this! Particularly the addition of "tags" and "excluded_tags" -- that's going to be very useful

PR #350, opened 2022-01-12
Create permissions.md

arubdesu:
I'm probably missing 'dry-run' santa endpoint perms, I can add them to this PR if you'd like, naming and anything else content-wise is open to discussion as well


np5:
Really sorry that I didn't merged this when we need it. It has become out of date now, but many thanks for your efforts.

PR #376, opened 2022-03-02
Added simple GELF input

janheise:
A simple GELF input like the syslog input to add message into graylog (https://www.graylog.org)


np5:
Many many thanks for your contribution. I am really sorry to not have merged this earlier, but we never had requests for this, so it would require extra efforts to make sure that it works. Also, we are now refactoring the stores to move their configuration and management into the database, so this is now out-of-date.

PR #470, opened 2022-06-02
Force Santa clean sync if bad rule counts

np5:
If the reported rule counts in the Santa preflight request are not equal to the number of machine rules stored on the server, force a clean sync.


np5:
Not really the best option.

PR #545, opened 2022-09-08
Specify dependency on apple enterprise developer program

zhumo:
I think this is a requirement in order for you to sign your own APNS certs with a vendor cert? You have to be an approved MDM vendor. 

I was not able to get this to work otherwise. When I log into my apple developer account, I get no link to manage my certificates, as per the instructions. 
image

### DIFF
--- a/docs/apps/mdm.md
+++ b/docs/apps/mdm.md
@@ -4,7 +4,7 @@ Zentral can be used as MDM server for Apple devices.
 
 ## Push certificates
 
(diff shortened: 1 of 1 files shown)


np5:
You need a enterprise developer account and you also probably need to request the option to sign MDM CSR.

np5:
Could you maybe change the wording ? It currently reads as if you could get the certificate under the enterprise developer portal. Maybe a separate sentence indicating that an enterprise developer account is required, with the MDM CSR option ?

np5:
Thanks!

headmin:
@zhumo As a follow up activity, please sign our individual CLA or would you need the Corporate Contributor License since your employer is involved?

zhumo:
@headmin I'm not sure honestly. When do I do one or the other? Can you send them over?

PR #605, opened 2023-01-19
Better handle inactive names in sub manifests

np5:
- Display a red background for PkgInfoNames without active PkgInfos in sub manifests.
- Do not display the edit button for PkgInfoNames without active PkgInfos in sub manifests.


soysauceconspiracy:
P

Coveralls ***@***.***> schrieb am Sa. 28. Sept. 2024 um 01:17:

PR #617, opened 2023-01-31
Add atcs api endpoints

xarbit:


tests/osquery/test_api_views.py
```diff
@@ -143,6 +162,253 @@ def put_json_data(self, url, data, include_token=True):
...
+            "description": atc.description,
+            "path": "/home/yolo",
+            "name": atc.name
+        }])
+
+    def test_get_atcs_by_configuration(self):
+        configuration, atc = self.force_configuration(force_atc=True)
+        configuration2, atc2 = self.force_configuration(force_atc=True)
+        self.set_permissions("osquery.view_automatictableconstruction")
+        response = self.get(reverse('osquery_api:atcs'), data={"configuration": configuration.id})
```
np5, line 198: Use `configuration_id` that we use everywhere else instead of `configuration`

tests/osquery/test_api_views.py
```diff
@@ -143,6 +162,253 @@ def put_json_data(self, url, data, include_token=True):
...
+            "platforms": ["darwin", "windows"],
+            "path": "/home/yolo"
+        })
+        self.assertEqual(response.status_code, 201)
+        atc = AutomaticTableConstruction.objects.first()
+        self.assertEqual(response.json(), {
+            "platforms": ["darwin", "windows"],
+            "updated_at": atc.updated_at.isoformat(),
+            "columns": ["un", "deux"],
+            "id": response.json()["id"],
```
np5, line 374: Testing that the response id is the response id is always true.

tests/osquery/test_api_views.py
```diff
@@ -143,6 +162,253 @@ def put_json_data(self, url, data, include_token=True):
...
+            "columns": ["un", "deux", "trois"],
+            "id": atc.id,
+            "created_at": atc.created_at.isoformat(),
+            "table_name": "yolo_new",
+            "query": "select new from yolo;",
+            "description": "yolo changed",
+            "path": "/home/yolo/new",
+            "name": atc.name
+        })
+        self.assertEqual(response.data, {
```
np5, line 317: testing response.data and response.json is a bit the same. The response and the object must be tested.

np5 (changes requested):

zentral/contrib/osquery/api_views.py
```diff
@@ -25,6 +25,11 @@
 from .tasks import export_distributed_query_results
 
 
+class AutomaticTableConstructionFilter(filters.FilterSet):
+    configuration_id = filters.NumberFilter(field_name='configuration__id')
```
np5, line 29: Probably better to user a [ModelChoiceFilter](https://github.com/zentralopensource/zentral/blob/12ae493bfc831ab024c42e447bafec432b51521a/zentral/contrib/santa/api_views.py#L170). Also maybe test if the config doesn't exist, the kind of error you are getting.

np5 (changes requested):

PR #620, opened 2023-02-02
Add Pack & ConfigurationPack endpoints

xarbit:


docs/apps/osquery.md
```diff
@@ -36,7 +36,201 @@ Zentral will parse the body of the request based on the `Content-Type` HTTP head
...
+Response:
+
+```json
+{
+    "id": 2,
+    "name": "Example",
+    "slug": "example",
+    "description": "description of the example",
+    "discovery_queries": [
+        "select * from example"
```
np5, line 123: Use a more meaningful [discovery query](https://osquery.readthedocs.io/en/stable/deployment/configuration/#discovery-queries).

zentral/contrib/osquery/api_views.py
```diff
@@ -416,6 +416,35 @@ def post(self, request, *args, **kwargs):
...
+class PackList(generics.ListCreateAPIView):
+    queryset = Pack.objects.all()
+    permission_classes = [DefaultDjangoModelPermissions]
+    serializer_class = PackSerializer
+    filter_backends = (filters.DjangoFilterBackend,)
+    filterset_class = PackFilter
+
+    def get(self, request, *args, **kwargs):
+        configuration_id = request.GET.get("configuration_id")
+        if configuration_id:
```
np5, line 436: Why do you need an exception here?

np5 (changes requested):

zentral/contrib/osquery/api_views.py
xarbit: @np5 
When the configuration exist, but it is not linked to any pack, the configuration_id filter will return an empty queryset, which is fine. 

But in the case, where the the configuration does not exist at all, we receive a default error that is not very useful. The message is: Select a valid choice. That choice is not one of the available choices. Id think we want to return a more explicit error message or at least always return an empty queryset, here I went for a default message.

That is why I created this check before calling the super method. 
So if the configuration does not exist and is not linked to any pack, we raise a ValidationError with a more clear message.

np5: I like the default behaviour better.

 * Valid configuration_id, no packs found → empty result
 * Invalid configuration_id, no configuration exists with the id → error saying that the choice is invalid, which is the case.

We already have the same error message when filtering by configuration_id on some other endpoints.

np5 (changes requested):

zentral/contrib/osquery/serializers.py
```diff
@@ -220,3 +221,37 @@ def update(self, instance, validated_data):
...
+
+
+class ConfigurationPackSerializer(serializers.ModelSerializer):
+    configuration = serializers.PrimaryKeyRelatedField(queryset=Configuration.objects.all())
+
+    class Meta:
+        model = ConfigurationPack
+        fields = "__all__"
+
+    def validate_configuration(self, value):
```
np5, line 233: Multiple packs can be added to the configuration. You should test that this is possible. This validation method is probably not necessary.

np5 (changes requested):

tests/osquery/test_api_views.py
```diff
@@ -1540,7 +1557,351 @@ def test_delete_enrollment(self):
...
+        pack = self.force_pack()
+        self.set_permissions("osquery.change_pack")
+        data = {"name": ""}
+        response = self.put_json_data(reverse("osquery_api:pack", args=(pack.pk,)), data)
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.json(), {
+            "name": ["This field may not be blank."]
+        })
+
+    def test_update_pack_not_found(self):
```
np5, line 1745: This test exists twice.

tests/osquery/test_api_views.py
```diff
@@ -1540,7 +1557,351 @@ def test_delete_enrollment(self):
...
+        self.assertEqual(pack.shard, 1)
+        self.assertEqual(pack.event_routing_key, "pack_created")
+
+    # delete pack <int:pk>
+
+    def test_delete_pack_unauthorized(self):
+        response = self.delete(reverse("osquery_api:pack", args=(1,)), include_token=False)
+        self.assertEqual(response.status_code, 401)
+
+    def test_delete_pack_permission_denied(self):
```
np5, line 1888: This test is duplicated

tests/osquery/test_api_views.py
```diff
@@ -1540,7 +1557,351 @@ def test_delete_enrollment(self):
...
+        response = self.delete(reverse("osquery_api:pack", args=(1,)))
+        self.assertEqual(response.status_code, 403)
+
+    def test_delete_pack_not_found(self):
```
np5, line 1892: This test is duplicated

tests/osquery/test_api_views.py
```diff
@@ -1540,7 +1557,351 @@ def test_delete_enrollment(self):
...
+
+    def test_update_pack_unauthorized(self):
+        response = self.put_json_data(reverse("osquery_api:pack", args=(1,)), {}, include_token=False)
+        self.assertEqual(response.status_code, 401)
+
+    def test_update_pack_permission_denied(self):
+        response = self.put_json_data(reverse("osquery_api:pack", args=(1,)), {})
+        self.assertEqual(response.status_code, 403)
+
+    def test_update_pack_not_found(self):
```
np5, line 1692: This test is duplicated

tests/osquery/test_api_views.py
```diff
@@ -1540,7 +1557,351 @@ def test_delete_enrollment(self):
...
+    def test_delete_pack_permission_denied(self):
+        response = self.delete(reverse("osquery_api:pack", args=(1,)))
+        self.assertEqual(response.status_code, 403)
+
+    def test_delete_pack_not_found(self):
+        self.set_permissions("osquery.delete_pack")
+        response = self.delete(reverse("osquery_api:pack", args=(9999,)))
+        self.assertEqual(response.status_code, 404)
+
+    def test_delete_pack(self):
```
np5, line 1897: This test is duplicated

tests/osquery/test_api_views.py
```diff
@@ -2401,3 +2764,283 @@ def test_delete_query_permission_denied(self):
...
+        response = self.get(reverse("osquery_api:configuration_packs"), include_token=False)
+        self.assertEqual(response.status_code, 401)
+
+    def test_list_configurationpacks_permission_denied(self):
+        response = self.get(reverse("osquery_api:configuration_packs"))
+        self.assertEqual(response.status_code, 403)
+
+    def test_list_configurationpacks_filter_configuration_id_not_found(self):
+        self.set_permissions("osquery.view_configurationpack")
+        response = self.get(reverse("osquery_api:configuration_packs"), {"configuration_id": 9999})
```
np5, line 2780: Why are we not getting a `Select a valid choice. That choice is not one of the available choices.` like in the other part of the API?

tests/osquery/test_api_views.py
```diff
@@ -2401,3 +2764,283 @@ def test_delete_query_permission_denied(self):
...
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json(), [])
+
+    def test_list_configurationpacks_filter_pack_id_not_found(self):
+        self.set_permissions("osquery.view_configurationpack")
+        response = self.get(reverse("osquery_api:configuration_packs"), {"pack_id": 9999})
```
np5, line 2786: Why are we not getting a `Select a valid choice. That choice is not one of the available choices.` like in the other part of the API?

tests/osquery/test_api_views.py
```diff
@@ -2401,3 +2764,283 @@ def test_delete_query_permission_denied(self):
...
+                "pack": pack
+            })
+            self.assertEqual(configuration_pack.pack.pk, pack)
+            self.assertEqual(configuration_pack.configuration.pk, configuration.pk)
+        result = ConfigurationPack.objects.filter(configuration_id=configuration.pk).count()
+        self.assertEqual(result, 3)
+
+    def test_create_configurationpack_with_pack_id_used_in_another_configuration_pack(self):
+        self.set_permissions("osquery.add_configurationpack")
+        _configuration_pack = self.force_configuration_pack()
```
np5, line 2986: do not prefix the variables with a `_`

tests/osquery/test_api_views.py
```diff
@@ -2401,3 +2764,283 @@ def test_delete_query_permission_denied(self):
...
+                "configuration": configuration.pk,
+                "tags": [],
+                "pack": pack
+            })
+            self.assertEqual(configuration_pack.pack.pk, pack)
+            self.assertEqual(configuration_pack.configuration.pk, configuration.pk)
+        result = ConfigurationPack.objects.filter(configuration_id=configuration.pk).count()
+        self.assertEqual(result, 3)
+
+    def test_create_configurationpack_with_pack_id_used_in_another_configuration_pack(self):
```
np5, line 2984: Maybe find a better name for the test, to describe that you want to make sure that adding the same pack to different configurations is possible.

np5 (changes requested):
Make sure you run a linter on your code, it would have caught the redefinitions of the tests (duplicated tests).

tests/osquery/test_api_views.py
xarbit: this is fixed, i am using ModelChoiceFilter in the api view

tests/osquery/test_api_views.py
xarbit: this is fixed, i am using ModelChoiceFilter in the api view

PR #621, opened 2023-02-03
Add PackQuery endpoints

xarbit:


zentral/contrib/osquery/serializers.py
```diff
@@ -220,3 +221,45 @@ def update(self, instance, validated_data):
...
+        slug = slugify(value.name)
+        pq_qs = PackQuery.objects.all()
+        if self.instance:
+            pq_qs = pq_qs.exclude(id=self.instance.id)
+        if pq_qs.filter(slug=slug).exists():
+            raise serializers.ValidationError(["PackQuery with this Query already exists."])
+        return value
+
+    def validate_log_removed_actions(self, value):
+        if value and self.initial_data.get("snapshot_mode"):
```
np5, line 244: If you need to validate values that depend on each others, you cannot use the `validate_` methods. You see this since you need to get the `initial_data` that might be changed by the data actually used by the serializer. See [the DRF docs](https://www.django-rest-framework.org/api-guide/serializers/#object-level-validation).

zentral/contrib/osquery/serializers.py
```diff
@@ -220,3 +221,45 @@ def update(self, instance, validated_data):
...
+            raise serializers.ValidationError(["'log_removed_actions' and 'snapshot_mode' are mutually exclusive"])
+        return value
+
+    def validate_snapshot_mode(self, value):
+        query = Query.objects.get(id=self.initial_data.get("query"))
```
np5, line 249: Same here

zentral/contrib/osquery/serializers.py
```diff
@@ -220,3 +221,45 @@ def update(self, instance, validated_data):
...
+class PackQuerySerializer(serializers.ModelSerializer):
+    pack = serializers.PrimaryKeyRelatedField(queryset=Pack.objects.all())
+    query = serializers.PrimaryKeyRelatedField(queryset=Query.objects.all())
+
+    class Meta:
+        model = PackQuery
+        fields = "__all__"
+
+    def validate_query(self, value):
+        slug = slugify(value.name)
```
np5, line 235: Can you confirm that the framework doesn't already catch this error?

np5 (changes requested):

zentral/contrib/osquery/serializers.py
xarbit: yes, If i dont catch this I get db error:
```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "osquery_packquery_pack_id_slug_94e4132b_uniq"
DETAIL:  Key (pack_id, slug)=(8763, 0a8oqwevsh4r) already exists.
```

zentral/contrib/osquery/serializers.py
```diff
@@ -220,3 +221,45 @@ def update(self, instance, validated_data):
...
+        pq_qs = PackQuery.objects.all()
+        if self.instance:
+            pq_qs = pq_qs.exclude(id=self.instance.id)
+        if pq_qs.filter(slug=slug).exists():
+            raise serializers.ValidationError(["PackQuery with this Query already exists."])
```
np5, line 240: The problem is not that a pack query with the same query exists. The problem is that 2 different queries with names resulting in the same slug cannot be added to the same pack. This is a corner case that we need to solve properly. I have tested this, and the bug exists in the UI as well. A better solution would be to add the PK of the query to the slug when there is a conflict:
- If no conflict, slug
- If conflict, slug + PK of the query (maybe with a separator, like "-")

np5 (changes requested):

zentral/contrib/osquery/serializers.py
```diff
@@ -220,3 +221,45 @@ def update(self, instance, validated_data):
         if compliance_check_deleted:
             query.refresh_from_db()
         return query
+
+
+class PackQuerySerializer(serializers.ModelSerializer):
+    pack = serializers.PrimaryKeyRelatedField(queryset=Pack.objects.all())
```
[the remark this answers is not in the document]
xarbit: yes, to make it editable.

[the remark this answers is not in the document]
np5: Why the query too ?

zentral/contrib/osquery/serializers.py
xarbit: ok, so basically to allow this by appending the pk to the slug name separated by "-" ..

np5: Yes, allow this, only append the PK if there is a conflict.

zentral/contrib/osquery/serializers.py
```diff
@@ -225,41 +225,27 @@ def update(self, instance, validated_data):
...
+        slug = slugify(query.name)
         pq_qs = PackQuery.objects.all()
         if self.instance:
             pq_qs = pq_qs.exclude(id=self.instance.id)
         if pq_qs.filter(slug=slug).exists():
-            raise serializers.ValidationError(["PackQuery with this Query already exists."])
-        return value
-
-    def validate(self, data):
+            slug = slugify(f"{slug}-{query.id}")
```
np5, line 240: no need to slugify the slug

np5 (changes requested):

docs/apps/osquery.md
```diff
@@ -1142,3 +1142,209 @@ $ curl -X DELETE \
...
+### /api/osquery/packqueries/
+
+#### List all Pack Queries.
+
+* method: GET
+* Content-Type: application/json
+* Required permission: `osquery.view_packquery`
+* Optional filter parameter:
+	* `pack_id`: primary key of the pack.
+    * `slug`: slug of the packquery.
```
np5, line 1155: I think you removed the filter by slug in the commit before?

np5 (changes requested):

docs/apps/osquery.md
np5: Also, nitpick, you could fix the identation

tests/osquery/test_api_views.py
```diff
@@ -2401,3 +2418,433 @@ def test_delete_query_permission_denied(self):
...
+        response = self.get(reverse("osquery_api:pack_queries"), include_token=False)
+        self.assertEqual(response.status_code, 401)
+
+    def test_get_packqueries_permission_denied(self):
+        response = self.get(reverse("osquery_api:pack_queries"))
+        self.assertEqual(response.status_code, 403)
+
+    def test_get_packqueries_filter_by_pack_id_not_found(self):
+        self.set_permissions("osquery.view_packquery")
+        response = self.get(reverse("osquery_api:pack_queries"), {"pack_id": 9999})
```
np5, line 2434: Why are we not getting a `Select a valid choice. That choice is not one of the available choices.` like in the other parts of the API?

tests/osquery/test_api_views.py
```diff
@@ -2401,3 +2418,433 @@ def test_delete_query_permission_denied(self):
...
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json(), [])
+
+    def test_get_packqueries_filter_by_slug_not_found(self):
```
np5, line 2438: There is no filter by slug in the view. Here you are only testing that there are not pack queries in the DB.

tests/osquery/test_api_views.py
```diff
@@ -2401,3 +2418,433 @@ def test_delete_query_permission_denied(self):
...
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.json(), {
+            "snapshot_mode": ["A compliance check query can only be scheduled in 'snapshot' mode."]
+        })
+
+    def test_update_packquery_fields_invalid(self):
+        self.set_permissions("osquery.change_packquery")
+        pack_query = self.force_packquery()
+        data = {
+            "pack": 5,
```
np5, line 2619: not really bullet proof to pick a random low value for the pack id that doesn't exist. Who knows what can happen.

tests/osquery/test_api_views.py
```diff
@@ -2401,3 +2418,433 @@ def test_delete_query_permission_denied(self):
...
+            "query": 10,
```
np5, line 2620: not really bullet proof to pick a random low value for the query id that doesn't exist. Who knows what can happen.

tests/osquery/test_api_views.py
```diff
@@ -2401,3 +2418,433 @@ def test_delete_query_permission_denied(self):
...
+        response = self.post_json_data(reverse("osquery_api:pack_queries"), data)
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.json(), {
+            "snapshot_mode": ["A compliance check query can only be scheduled in 'snapshot' mode."]
+        })
+
+    def test_create_packquery_fields_invalid(self):
+        self.set_permissions("osquery.add_packquery")
+        data = {
+            "pack": 5,
```
np5, line 2778: See above

tests/osquery/test_api_views.py
```diff
@@ -2401,3 +2418,433 @@ def test_delete_query_permission_denied(self):
...
+            "query": 10,
```
np5, line 2779: See above

np5 (changes requested):

zentral/contrib/osquery/api_views.py
```diff
@@ -416,6 +416,27 @@ def post(self, request, *args, **kwargs):
                         status=status.HTTP_201_CREATED)
 
 
+# Pack Queries
+
+class PackQueryFilter(filters.FilterSet):
+    pack_id = filters.ModelChoiceFilter(field_name="pack_id", queryset=Pack.objects.all())
+    #pack_id = filters.NumberFilter(field_name="pack__id")
```
np5, line 423: Do not leave a comment here. Also your linter should catch this since a comment must start with `# ` and not `#`.

np5 (changes requested):

zentral/contrib/osquery/api_views.py
xarbit: oh, that shouldn't be there

np5 (approved):

PR #622, opened 2023-02-08
Fix: slug conflict in PackQuery UI

xarbit:


tests/osquery/test_setup_packs.py
```diff
@@ -41,19 +42,31 @@ def _login(self, *permissions):
...
+        slug = slugify(query.name)
         if force_pack:
             pack = self._force_pack()
-            PackQuery.objects.create(pack=pack, query=query, interval=12983,
+            PackQuery.objects.create(pack=pack, query=query, interval=12983, slug=slug,
                                      log_removed_actions=False, snapshot_mode=force_compliance_check)
         sync_query_compliance_check(query, force_compliance_check)
         return query
 
+    def _force_pack_query(self, query_name=None, force_compliance_check=False):
```
np5, line 63: I do not think this method is necessary, it is used only once.

np5 (changes requested):

np5 (approved):

PR #698, opened 2023-05-24
feat: Adding MS Intune Inventory App to store configurations for Inventory support

mazzi:
Created a new app called `intune` to handle MS Intune Inventory configuration support.

- CRUD operations
- Tests for views

Heavily inspired on WSOne app.


ee/zentral/contrib/intune/forms.py
```diff
@@ -0,0 +1,48 @@
...
+from zentral.contrib.inventory.models import BusinessUnit
+from .models import Tenant
+
+
+class TenantForm(forms.ModelForm):
+    client_secret = forms.CharField(widget=forms.PasswordInput(render_value=True))
+
+    class Meta:
+        model = Tenant
+        fields = (
```
np5, line 11: Use `__all__` if all the models non-editable fields are present in the form

ee/zentral/contrib/intune/models.py
```diff
@@ -0,0 +1,53 @@
...
+
+from zentral.core.secret_engines import decrypt_str, encrypt_str, rewrap
+
+
+logger = logging.getLogger("zentral.contrib.intune.models")
+
+
+class Tenant(models.Model):
+    business_unit = models.ForeignKey("inventory.BusinessUnit", on_delete=models.PROTECT)
+    name = models.TextField(unique=True)
```
np5, line 14: CharField probably better to indicate that it should be a bit smaller.

ee/zentral/contrib/intune/models.py
```diff
@@ -0,0 +1,53 @@
...
+    description = models.TextField(blank=True, null=True)
+    # Authentication
+    tenant_id = models.TextField(help_text="The microsoft Azure Tenant ID")
```
np5, line 17: See if we could use a UUID field here, to restrict a bit the format. Same about the other fields. If not then CharFields probably, and you could remove the override for the form widgets.

tests/intune/test_index_view.py
```diff
@@ -0,0 +1,185 @@
...
+import operator
+from django.contrib.auth.models import Permission
+from django.db.models import Q
+from django.utils.crypto import get_random_string
+from accounts.models import APIToken, User
+from django.contrib.auth.models import Group, Permission
+from zentral.contrib.inventory.models import MetaBusinessUnit
+from zentral.contrib.intune.models import Tenant
+
+
```
np5, line 14: Use https://github.com/zentralopensource/zentral/blob/7245d57d31f0d329a711c8e7d7cee43f02cf5a15/tests/wsone/test_setup_views.py#L14

It will probably solve the test errors

conf/start/zentral/base.json
```diff
@@ -89,6 +89,8 @@
     "zentral.contrib.osquery": {},
     "zentral.contrib.santa": {
       "metrics": true
-    }
+    },
```
np5, line 92: Not sure about activating this app by default.

tests/intune/test_index_view.py
```diff
@@ -0,0 +1,185 @@
...
+        self._login()
+        response = self.client.get(reverse("intune:tenants"))
+        self.assertEqual(response.status_code, 403)
+
+    def test_tenants(self):
+        tenant = self._force_tenant()
+        self._login("intune.view_tenant")
+        response = self.client.get(reverse("intune:tenants"))
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, tenant.name)
```
np5, line 73: Add a assertTemplateUsed test please

np5 (changes requested):

conf/start/zentral/base.json
np5: You should not be using the default config, but the devmdm one, or another custom one.

ee/zentral/contrib/intune/models.py
```diff
@@ -0,0 +1,53 @@
...
+
+logger = logging.getLogger("zentral.contrib.intune.models")
+
+
+class Tenant(models.Model):
+    business_unit = models.ForeignKey("inventory.BusinessUnit", on_delete=models.PROTECT)
+    name = models.CharField(max_length=256, unique=True)
+    description = models.TextField(blank=True, null=True)
+    # Authentication
+    tenant_id = models.CharField(max_length=256, help_text="The microsoft Azure Tenant ID")
```
np5, line 17: I think this field should have a unique=True constraint. Sorry that I missed it before.

ee/zentral/contrib/intune/models.py
```diff
@@ -0,0 +1,53 @@
...
+from zentral.core.secret_engines import decrypt_str, encrypt_str, rewrap
+
+
+logger = logging.getLogger("zentral.contrib.intune.models")
+
+
+class Tenant(models.Model):
+    business_unit = models.ForeignKey("inventory.BusinessUnit", on_delete=models.PROTECT)
+    name = models.CharField(max_length=256, unique=True)
+    description = models.TextField(blank=True, null=True)
```
np5, line 15: TextFields and CharFields are better without the null=True, to avoid having a double empty value, null or "". See [the Django docs](https://docs.djangoproject.com/en/4.2/ref/models/fields/#null) for more context.

ee/zentral/contrib/intune/models.py
```diff
@@ -0,0 +1,53 @@
...
+    # Authentication
+    tenant_id = models.CharField(max_length=256, help_text="The microsoft Azure Tenant ID")
+    client_id = models.UUIDField(help_text="The client ID of your app registration")
```
np5, line 18: Client ID should also have a unique=True constraint. I do not see a case for the same client being used with 2 different tenants, and this could avoid simple config errors.

np5 (changes requested):
Sorry, a few things I have missed about the model during the first review. Please try to consolidate the initial DB migration in one file after the changes.

ee/zentral/contrib/intune/models.py
mazzi: Thought about it. I'll add it.
I'm not happy with having name as `unique=True` but was on the specs.

ee/zentral/contrib/intune/models.py
mazzi: I'll add. Pretty impossible to have duplicated entries as this is the application ID that the [Azure app registration portal](https://go.microsoft.com/fwlink/?linkid=2083908) assigned to the app.

mazzi:
Thanks for the review 🙌

np5 (approved):

PR #701, opened 2023-05-30
feat: Intune Inventory Import

mazzi:
Early PR on Intune Inventory Import.

- New `django manage` commands: `--list-tenants`, `--tenant` to list and to import machines for a certain tenant
- New dependencies

Future work: check `TODO`s
- [ ] Enrich snapshot tree with more information, like apps, that will need another remote call.
- [x] Limit the amount of results when fetching remote information using `last_sync_day` or similar. [API Query options](https://learn.microsoft.com/en-us/graph/query-parameters).
- [ ] Automated tests 🙃 

Based on `WSOne` and [MS Graph Rest API](https://learn.microsoft.com/en-us/graph/api/resources/intune-devices-manageddevice?view=graph-rest-1.0)

PR #702, opened 2023-05-31
feat: Intune API

mazzi:
Three new endpoints for Intune
- List
- Retrieve (`tenant_id`)
- Start sync (`tenant_id`)

Tests included based on WSOne ones.

Depends on #701


ee/zentral/contrib/intune/api_views.py
```diff
@@ -0,0 +1,44 @@
...
+from rest_framework.views import APIView
+from rest_framework.response import Response
+from zentral.core.events.base import EventRequest
+from zentral.utils.drf import DefaultDjangoModelPermissions, DjangoPermissionRequired
+from .models import Tenant
+from .serializers import TenantSerializer
+from .tasks import sync_inventory
+
+
+class TenantList(generics.ListAPIView):
```
np5, line 13: Implement the other operation so ListCreate, not only read only. Also, use the [ListCreateAPIViewWithAudit](https://github.com/zentralopensource/zentral/blob/6e63568a1ebf9a77b01f6cc04a96e68509a7cac4/zentral/contrib/santa/api_views.py#L25) base view.

ee/zentral/contrib/intune/api_views.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    List all Tenants
+    """
+    queryset = Tenant.objects.all()
+    permission_classes = [DefaultDjangoModelPermissions]
+    serializer_class = TenantSerializer
+
+
+class TenantDetail(generics.RetrieveAPIView):
```
np5, line 22: Same with [RetrieveUpdateDestroyAPIViewWithAudit](https://github.com/zentralopensource/zentral/blob/6e63568a1ebf9a77b01f6cc04a96e68509a7cac4/zentral/contrib/santa/api_views.py#LL34C27-L34C64)

tests/intune/test_api_view.py
```diff
@@ -0,0 +1,176 @@
...
+        response = self.get(reverse("intune_api:tenants"))
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(
+            {'id': tenant.pk,
+             'business_unit': tenant.business_unit.pk,
+             'name': tenant.name,
+             'description': tenant.description,
+             'tenant_id': tenant.tenant_id,
+             'client_id': str(tenant.client_id),
+             'client_secret': tenant.client_secret,
```
np5, line 104: Probably not the best for usability. We should decrypt the secret I think.

np5 (changes requested):

ee/zentral/contrib/intune/models.py
```diff
@@ -35,6 +35,18 @@ def save(self, *args, **kwargs):
...
         super().save(*args, **kwargs)
 
+    def serialize_for_event(self):
+        return {
+            'business_unit': self.business_unit.pk,
+            'name': self.name,
+            'description': self.description,
+            'tenant_id': self.tenant_id,
+            'client_id': str(self.client_id),
+            'client_secret': self.client_secret,
```
np5, line 45: It is good to have the encrypted client secret in the event, because otherwise we could not see the different when the client secret is updated. **BUT** the default secret encryption in Zentral is … not an encryption, it is just a base64 encoding. So, I think it would be better to replace the `client_secret` member with a `client_secret_hash` and use hashlib.sha256 on the decrypted client secret. This way 1) it is easy for the users to compare with the value they know, 2) we never leak the secret, even if the secret engine is not configured in Zentral. Also, the secret is generated by Microsoft and is good enough to not have to salt the hash I think…

np5 (changes requested):

PR #714, opened 2023-06-03
fix: Remove unnecessary method from intune views

mazzi:
Removed `get_context_data` from intune views, as is not needed (same as default one).


np5 (changes requested):
This cannot work, because we need to test the perms, and we need the context attributes for the templates. We need to add a test for the "intune:index" view, because this should have triggered an error!

mazzi:
So probably I didn't understand your comment when you wrote:
```
no need to redefine get_context_data() or form_valid() or delete() if you are not changing what the default views are doing
you can keep the views very clean
the specific stuff can be managed usually in the forms
```
All good 👍

PR #842, opened 2023-11-06
fix: Path alignment on OSQuery ATC details

mazzi:


mazzi:
Closed in favor of https://github.com/zentralopensource/zentral/pull/845

PR #854, opened 2023-11-14
fix: UI Bugfixing IV

mazzi:


server/static_src/scss/styles.scss
```diff
@@ -60,6 +60,14 @@ pre.scrolleable {
     padding-bottom: 2em;
 }
 
+td > p:last-child,
+td > ul:last-child,
```
np5, line 64: Seems a lot redundant with 

https://github.com/zentralopensource/zentral/blob/5da3ad771a107dfdbc8f34857a9a1f2c29f44c19/server/static_src/scss/styles.scss#L41-L69C2

So maybe only  `<p/>` ?

np5 (changes requested):

PR #856, opened 2023-11-15
fix: Zentral logo dark/light mode

mazzi:
Using CSS.

### DIFF
--- a/server/static_src/scss/styles.scss
+++ b/server/static_src/scss/styles.scss
@@ -186,12 +186,12 @@ td > p:last-child {
     display: block;
 }
 
-.zentral-logo.light {
-    background-image: url(../images/logo-light.svg);
+[data-bs-theme=light] .zentral-logo {
+        background-image: url(../images/logo-light.svg);
 }
 
-.zentral-logo.dark {
-    background-image: url(../images/logo-dark.svg);
+[data-bs-theme=dark] .zentral-logo {
+        background-image: url(../images/logo-dark.svg);
 }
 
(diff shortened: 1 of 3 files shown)


server/static_src/scss/styles.scss
```diff
@@ -182,12 +182,16 @@ pre.scrolleable {
     display: block;
 }
 
-.zentral-logo.light {
-    background-image: url(../images/logo-light.svg);
+[data-bs-theme=light] {
+    .zentral-logo {
```
np5, line 186: Please do not use nested CSS, especially for one nested rule.

server/static_src/scss/styles.scss
```diff
@@ -182,12 +182,16 @@ pre.scrolleable {
...
+        background-image: url(../images/logo-light.svg);
+    }
 }
 
-.zentral-logo.dark {
-    background-image: url(../images/logo-dark.svg);
+[data-bs-theme=dark] {
+    .zentral-logo {
```
np5, line 192: Same

np5 (changes requested):

server/static_src/scss/styles.scss
mazzi: This will grow, and this is `scss`, but as you wish.

np5: Yes, but nesting is still a young feature and we do not really need it since we already have a SCSS pipeline.

np5 (approved):

PR #859, opened 2023-11-16
fix: No results placeholder after search

mazzi:


zentral/contrib/mdm/templates/mdm/depdevice_list.html
```diff
@@ -80,7 +80,7 @@ <h2 class="m-0">Device{{ page_obj.paginator.count|pluralize }} ({{ page_obj.pagi
 
     {% pagination next_url previous_url %}
 
-{% elif form.is_bound %}
+{% elif form.has_changed %}
     {% url 'mdm:dep_device' as empty_results_url %}
```
np5, line 84: This link doesn't work

zentral/contrib/mdm/templates/mdm/enrolleddevice_list.html
```diff
@@ -74,7 +74,7 @@ <h2 class="m-0">Device{{ page_obj.paginator.count|pluralize }} ({{ page_obj.pagi
 
     {% pagination next_url previous_url %}
 
-{% elif form.is_bound %}
+{% elif form.has_changed %}
     {% url 'inventory:enrolled_device' as empty_results_url %}
```
np5, line 78: This link doesn't work

zentral/contrib/monolith/templates/monolith/manifest_list.html
```diff
@@ -53,7 +53,7 @@ <h2>Manifest{{ paginator.count|pluralize }} ({{ paginator.count }})</h2>
     </table>
 
     {% pagination next_url previous_url %}
-{% elif form.is_bound %}
+{% elif form.has_chaged %}
```
np5, line 56: this fails because of a typo, should be has_changed.

zentral/core/probes/templates/probes/index.html
```diff
@@ -76,7 +76,7 @@ <h2 class="mb-3">Probe{{ paginator.count|pluralize }} ({{ paginator.count }})</h
     </div>
 
     {% pagination next_url previous_url %}
-{% else %}
+{% elif form.has_changed %}
```
np5, line 79: Doesn't work, because active is already set in the form. Also, the link doesn't really links to all the probes.

There is a `is_initial` method on the form. Or maybe because it is not that big of an issue, we just remove the active by default, and have simply the same mechanism as with the other ones.

np5 (changes requested):
You should write automatic tests for these cases, to avoid these small issues.

mazzi:
> You should write automatic tests for these cases, to avoid these small issues.

Sure

zentral/core/probes/templates/probes/index.html
mazzi: Thanks for your review.
There's a decision to be taken. Do we remove the initial state? or we keep it?

np5: Remove the default to active. Make it work like the other search forms please.

PR #960, opened 2024-04-17
Use header authentication for the Santa agent

np5:
- The enrollment secret must be present as `Bearer` token in the `Zentral-Authorization` header.
 - The existing URLs are deprecated and will be removed in the near future.


np5:
The impact should be minimal. The deprecated endpoints are still present to maintain compatibility for the existing deployments.

np5:
No issues noticed after deployment.

PR #1061, opened 2024-09-18
Fix the Santa TEAMID target search

np5:
Use the collected signing IDs, to find team IDs event when the app comes from the App Store.


headmin (approved):
LGTM

PR #1066, opened 2024-10-11
Stop event queues on APNS worker errors

np5:
This allows the process to exit gracefully. Otherwise, an event publishing thread might still be running, and the process would not be restarted by systemd for example.


headmin (approved):
LGTM

PR #1067, opened 2024-10-11
Fix incident status in machine list

np5:
Closed machine incidents belonging to open incidents must be ignored.


headmin (approved):
LGTM!

PR #1069, opened 2024-10-23
Rephrasing copy re: password resets

arubdesu:
Tweaking expressions to be more in line with how other systems phrase things, feel free to cherry pick, glad to provide examples from other systems if that would be better proof-wise


np5 (approved):
LGTM

PR #1083, opened 2024-11-20
Ensure Santa ballot unicity across realms

np5:
Make sure that not only a ballot from the same realm user is a match, but also a ballot from a different realm user with the same username. This avoids double votes in the case of 2 SAML apps with different scopes (All the employees for the portal, only the admins for the Zentral console) but from the same IdP, so with the same usernames.


headmin (approved):
LGTM

PR #1114, opened 2025-01-30
Clean sync when no rule counts in Santa preflight

np5:
This should fix the sync issue when a device is erased and enrols on the same configuration, with the same enrollment.


headmin (approved):
LGTM

PR #1126, opened 2025-02-22
Fix Realm Group Mappings interference with SCIM

np5:
We do not want the Realm Group Mappings to interfere with the memberships managed by SCIM. When we apply the Realm Group Mappings, we preserve the memberships to the Realm Groups with a SCIM identifier.


headmin (approved):
LGTM

PR #1130, opened 2025-02-23
Switch back to MDM command for legacy profiles

np5:
At the moment, the MDM agent and daemon do no retry the failed downloads of configuration profiles. This is problematic, especially when some payloads disrupt the network (Firewall config). The devices could be left in a state that requires a reboot to recover from. This is not good enough, especially during an MDM enrollment.


headmin (approved):
LGTM

PR #1132, opened 2025-02-26
Use notifier for probes

np5:
Remove dependency on PostgreSQL listen/notify. That should reduce the number of connections to the database.


headmin (approved):
LGTM

PR #1141, opened 2025-03-06
Better Monolith repositories cache

np5:
- Only update the cache if DB request successsful
 - Reload the cache if missing repo


headmin (approved):
LGTM

PR #1143, opened 2025-03-07
Fix MDM scoping for required artifacts

np5:
A required artefact also included directly in a blueprint should always be considered for installation.

Before this fix, if an artefact was included in the serialized artifacts with a `depth` of `1` because of a requirement, and later with a `depth` of `0`, the `depth` would stay at `1`. Because of that, the artefact would not be included in the dependency tree if the artefact requiring it was not also included (not in scope).


headmin (approved):
LGTM

PR #1167, opened 2025-05-10
Fix DB error during Santa bundle rule migration

np5:
Different legacy bundle rules can cover the same binary. That could cause integrity error during the migration.


headmin (approved):
LGTM

PR #1169, opened 2025-05-21
Make MDM declaration ServerToken mandatory in API

np5:
This should make it easier when adding this to the Terraform provider. If the ServerToken can be automatically generated by the server, this could introduce drift.

We keep the option in the GUI.


headmin (approved):
LGTM

PR #1182, opened 2025-07-07
Basic Auth and TLS overrides for Elasticsearch

np5:
Allow basic auth configuration and cert verification overrides in Elasticsearch event store configurations.


soysauceconspiracy (approved):
LGTM

PR #1184, opened 2025-07-09
Make Santa target search filters mandatory

np5:
Do not run a full Santa target search without filters in the Targets view.


soysauceconspiracy (approved):
LGTM

PR #1192, opened 2025-07-20
Stores updates

np5:
This is used in the SaaS to limit the number of event stores a customer can create. It might be useful somewhere else too!


soysauceconspiracy (approved):
LGTM

PR #1196, opened 2025-07-22
Fix store queue deletion because of missing PK

np5:
Some queues backends might use the PK to identify the queue associated with a store. After a deletion, this PK must be preserved to be able to mark the queue as deleted.


soysauceconspiracy (approved):
LGTM

PR #1230, opened 2025-09-22
Post the DEP profile updates in background task

np5:
When a DEP profile is assigned to a lot of devices, the profile update with ABM can take a long time. This commit moves the update to a background task.
We also only post an update when the DEP profile changes, and avoid posting an update when only DEP enrollment attributes that do not affect the profile are changed.


soysauceconspiracy (approved):
LGTM

PR #1263, opened 2025-11-03
Publish audit events on creating or deleting a api token for a user.

daniel-zentral:
Publish a audit event when creating or deleting a api token for a user.


np5 (approved):
LGTM

PR #1272, opened 2025-11-07
Fix MDM apps & books preprocessor

np5:
The device notification events were sent directly using the default queues instead of being generated by the preprocessor. On queues with threads, like AWS SQS/SNS, a separate thread would be started to send the event. This thread because it is not managed by the preprocessor would block the graceful shutdown.


daniel-zentral (approved):
LGTM

PR #1279, opened 2025-11-11
Remove unused import and TestCase renaming

Sefux:
Better not import scss in javascript files


Sefux:
Turns out to be sass-loader standard. Closing this pr.

PR #1295, opened 2025-11-24
Added sync of machine tags based on google workspace group tag mappings.

daniel-zentral:


zentral/contrib/google_workspace/tasks.py
```diff
@@ -0,0 +1,13 @@
+import uuid
+import logging
+from zentral.contrib.google_workspace.models import Connection as ApiConnection
```
np5: I don't really like the renaming if it is not necessary. And if it is necessary, I prefer `APIConnection`.

zentral/contrib/google_workspace/tasks.py
```diff
@@ -0,0 +1,13 @@
...
+from zentral.contrib.google_workspace.utils import sync_group_tag_mappings as sync
+
+
+def sync_group_tag_mappings(api_connection_pk: uuid) -> None:
+    api_connection_qs = ApiConnection.objects.filter(pk=api_connection_pk)
```
np5: It is a bit too convoluted. Simply get and except Connection.DoesNotExist

zentral/contrib/google_workspace/utils.py
```diff
@@ -1,18 +1,17 @@
 import logging
-import uuid
 import psycopg2.extras
 from collections import defaultdict
+from collections.abc import Iterator
 from django.db import connection, transaction
 from zentral.contrib.google_workspace.models import Connection as ApiConnection
```
np5: Same, just Connection or APIConnection or GWConnection…

tests/google_workspace/test_utils.py
```diff
@@ -0,0 +1,469 @@
...
+
+
+class UtilsTestCase(TestCase):
+
+    @classmethod
+    def setUpTestData(cls):
+        cls.principal_user_source = PrincipalUserSource.objects.create(type=PrincipalUserSource.INVENTORY)
+        cls.source = Source.objects.create(module="source module", name="source name")
+
+    def _given_user_email(self):
```
np5, line 23: To simulate an inventory machine, it is much easier to just "commit" a reduced inventory tree and used the existing functions: https://github.com/zentralopensource/zentral/blob/6eaf1f1e67feedae9b0363868a2c39ac8496a119/tests/inventory/test_update_machine_tags.py#L57-L71

tests/google_workspace/test_utils.py
```diff
@@ -0,0 +1,469 @@
...
+            name=f"business_unit_{get_random_string(5)}"
+        )
+        return Taxonomy.objects.create(
+            meta_business_unit=unit,
+            name=f"taxonomy_{get_random_string(5)}"
+        )
+
+    def _given_group_tag_mapping(self, connection, tag, group_email=None):
+        if not group_email:
+            group_email = f"{connection.name}@zentral.com"f"{connection.name}@zentral.com"
```
np5: Detail: see [model](https://github.com/zentralopensource/zentral/blob/6eaf1f1e67feedae9b0363868a2c39ac8496a119/zentral/contrib/google_workspace/models.py#L28-L29). You can use `{connection}`.

tests/google_workspace/test_utils.py
```diff
@@ -0,0 +1,469 @@
...
+            self.assertIsInstance(event, MachineTagEvent)
+            self.assertEqual(
+                event.payload,
+                self._expected_machine_tag_event_payload(tag)
+            )
+
+            metadata = event.metadata.serialize()
+            self.assertTrue(metadata["machine_serial_number"] == serial_number
+                            or metadata["machine_serial_number"] == other_serial_number)
+            self.assertEqual(sorted(metadata["tags"]), ["machine"])
```
np5: no need to sort if only one

tests/google_workspace/test_utils.py
```diff
@@ -0,0 +1,469 @@
...
+
+        # device has tag
+        self._given_machine_tag(snapshot.serial_number, tag)
+
+        # When
+        with self.captureOnCommitCallbacks(execute=True) as callbacks:
+            sync_group_tag_mappings(connection.pk)
+
+        # Then
+        actual = MachineTag.objects.filter(serial_number=snapshot.serial_number, tag_id=tag.pk)
```
np5: A bit unclear here. You are only verifying that the tag is not touched, and that the event is not on the tag. Maybe for clarity also test that the other tag is there and that the event is for the other tag.

zentral/contrib/google_workspace/utils.py
```diff
@@ -0,0 +1,124 @@
...
+            delete from inventory_machinetag im
+            using serial_numbers sn
+            where not sn.add_operation
+                and im.serial_number = sn.serial_number
+                and im.tag_id = sn.tag_id
+            returning im.serial_number, im.tag_id, 'removed'::text as action)
+        select
+            it.serial_number, 'added' action, t.id pk, t.name, tx.id taxonomy_pk, tx.name taxonomy_name
+        from
+            inserted_tags it join inventory_tag t on (it.tag_id = t.id)
```
np5: same formatting as the delete query below to make it easy to see the similarity.

zentral/contrib/google_workspace/utils.py
```diff
@@ -0,0 +1,124 @@
...
+import psycopg2.extras
+from collections import defaultdict
+from django.db import connection, transaction
+from zentral.contrib.google_workspace.models import Connection as ApiConnection
+from zentral.contrib.google_workspace.api_client import APIClient
+from zentral.contrib.inventory.events import MachineTagEvent
+from zentral.core.events.base import EventMetadata
+
+
+logger = logging.getLogger('zentral.contrib.google_workspace.views')
```
np5: utils, not views.

zentral/contrib/google_workspace/views.py
```diff
@@ -51,7 +51,7 @@ def get(self, request, *args, **kwargs):
             api_client = APIClient.from_oauth2_state(state)
             api_client.complete_authorization(code)
         except Exception as e:
-            logger.info("Unable to authorize connection.")
+            logger.info("Unable to authorize connection.", e)
```
np5: In that case, e will not be used. Better to simply use logger.exception: no need for `as e`. And we will get a traceback.

zentral/contrib/google_workspace/utils.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def send_machine_tag_added_events():
+        send_machine_tag_events(results)
+
+    transaction.on_commit(send_machine_tag_added_events)
+    return len(results)
+
+
+def sync_group_tag_mappings(api_connection: ApiConnection) -> None:
+    if (api_connection):
```
np5: Is there a case where there is no api_connection ?

np5 (changes requested):

zentral/contrib/google_workspace/utils.py
daniel-zentral: We can remove the type alias here. 
I originally included it as I was querying the Connection inside this module and, for me, this alias made easier to distinguish between `zentral.contrib.google_workspace.models.Connection `and `django.db.connection`. As query moved to the task module removing the alias is more than fine for me as I do not like aliasing for imports either.

zentral/contrib/google_workspace/tasks.py
daniel-zentral: I just added the aliasing here to be consistent with the [utils module](https://github.com/zentralopensource/zentral/pull/1295#discussion_r2558966059). 
Will remove the alias here aswell.

zentral/contrib/google_workspace/tasks.py
daniel-zentral: In the end I do not care. My two cents here are I do prefer code wich does not rely on error/exception handling for a valid case a bit more. But one can argue, it's just personal preference. Will change the code to execption handling.

tests/google_workspace/test_utils.py
daniel-zentral: Ah. This is really nice and handy.
... would it be possible to provide such things as a kind of a test fixture?

tests/google_workspace/test_utils.py
daniel-zentral: ok

tests/google_workspace/test_utils.py
daniel-zentral: Good point. I have to be more aware at c&p.

tests/google_workspace/test_utils.py
daniel-zentral: It clearly does not hurt to add this assertion if it helps to increase readability. 
The reason I omitted the assertion is that the other tag is just a vehicle to enforce the sync.

zentral/contrib/google_workspace/utils.py
daniel-zentral: Thanks for spotting. :)

zentral/contrib/google_workspace/utils.py
daniel-zentral: ups ... 🙈

zentral/contrib/google_workspace/views.py
daniel-zentral: Discussed offline

zentral/contrib/google_workspace/utils.py
daniel-zentral: Removed as discussed offline

np5 (approved):
LGTM

PR #1297, opened 2025-11-27
Added api views to list connections and group tag mappings

daniel-zentral:
Added api views to list connections and list, create, update, delete group tag mappings of connections

### DIFF
--- a/tests/google_workspace/test_api_client.py
+++ b/tests/google_workspace/test_api_client.py
@@ -1,4 +1,5 @@
 import uuid
+import json
 from django.test import TestCase
 from unittest.mock import patch, Mock
 from django.utils.crypto import get_random_string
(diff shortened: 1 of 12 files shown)


tests/google_workspace/test_api_client.py
```diff
@@ -13,8 +13,14 @@ class ApiClientTestCase(TestCase):
     def _given_connection(self):
         name = get_random_string(12)
         client_config = """{"web":{}}"""
+        user_info = f"""{{
```
np5: use json.dumps() instead ?

tests/google_workspace/test_utils.py
```diff
@@ -44,8 +44,14 @@ def _force_machine(self, user_email, serial_number=None):
     def _given_connection(self):
         name = get_random_string(12)
         client_config = """{"web":{}}"""
+        user_info = f"""{{
```
np5: Same.

np5 (approved):
LGTM

PR #1305, opened 2025-12-04
Fix `managed_updates` sub manifest filtering

np5:
The `managed_updates` were always included, regardless of the tags and shards.


daniel-zentral (approved):
LGTM

PR #1315, opened 2025-12-09
Auditevent for monolith submanifestpkginfo

sebastian-zentral:


zentral/contrib/monolith/models.py
```diff
@@ -306,8 +306,9 @@ def get_absolute_url(self):
     def serialize_for_event(self):
         return {"pk": self.pk, "name": self.name, "created_at": self.created_at}
 
-    def linked_objects_keys_for_event(self):
-        return {"munki_pkginfo_name": ((self.name,),)}
+    # TODO: this one can go?
```
np5: We have to maintain backward compatibility, at least for the standard log retention, especially [because of the current search](https://github.com/zentralopensource/zentral/blob/5c22ace14557f789b1f9ae7ef4f1b877cb0738eb/zentral/contrib/monolith/views.py#L430). We can add `monolith_pkg_info_name` with the `pk`. After the standard retention period is passed, we can remove the original link, and change the search.

zentral/contrib/monolith/models.py
```diff
@@ -556,8 +557,9 @@ def serialize_for_event(self, keys_only=False):
         return d
 
     def linked_objects_keys_for_event(self):
-        return {"munki_pkginfo_name": ((self.name.name,),),
-                "munki_pkginfo": ((self.name.name, self.version),)}
+        return {"monolith_pkg_info": [(self.pk,)],
```
np5: Sorry, same here, keep the old ones for backward compatibility.

zentral/contrib/monolith/models.py
```diff
@@ -797,6 +799,30 @@ def default_shard(self):
...
+            "options": self.options,
+            "created_at": self.created_at,
+            "updated_at": self.updated_at
+        })
+        return d
+
+    def linked_objects_keys_for_event(self):
+        return {"monolith_sub_manifest_pkg_info": [(self.pk,)],
+                "monolith_sub_manifest": [(self.sub_manifest.pk,)],
+                "monolith_pkg_info_name": [(self.pkg_info_name.pk,)]}
```
np5, line 826: Add the old ones for backward compatibility.

zentral/contrib/monolith/views.py
```diff
@@ -674,13 +672,14 @@ def get_context_data(self, **kwargs):
         return context
 
     def form_valid(self, form):
-        smpi = form.save()
-        for _, manifest in smpi.sub_manifest.manifests_with_tags():
+        sub_manifest = self.object.sub_manifest
+        super().form_valid(form)
```
np5: save the result of this in a response variable, and at the end of the method, simply `return response`?

tests/monolith/utils.py
```diff
@@ -99,8 +99,15 @@ def force_category(repository=None, name=None):
     return PkgInfoCategory.objects.create(repository=repository, name=name or get_random_string(12))
 
 
-def force_name(name=None):
-    return PkgInfoName.objects.get_or_create(name=name or get_random_string(12))[0]
+def force_name(name=None, setup_pkginfo=False):
+    pkginfo_name = PkgInfoName.objects.get_or_create(name=name or get_random_string(12))[0]
+    if setup_pkginfo:
+        PkgInfo.objects.create(
```
np5: Reuse `_force_pkg_info` ?

zentral/contrib/monolith/models.py
sebastian-zentral: The old key/values are back in.

zentral/contrib/monolith/models.py
sebastian-zentral: Reverted

zentral/contrib/monolith/models.py
sebastian-zentral: Same here

zentral/contrib/monolith/views.py
sebastian-zentral: top

PR #1320, opened 2025-12-11
Add command tests for 'import_mac_assignments' & 'debug_inventory_cli…

sebastian-zentral:
…ents'


zentral/contrib/inventory/management/commands/import_mac_assignments.py
```diff
@@ -43,7 +43,7 @@ def handle(self, *args, **kwargs):
             if self.verbosity:
                 row_iterator = tqdm(csv.reader(lines), total=len(lines))
             else:
-                row_iterator = lines
+                row_iterator = csv.reader(lines)
```
sebastian-zentral, line 48: Please check. Where is the command used?

np5: Thanks!

PR #1362, opened 2026-01-20
Edit api tokens

sebastian-zentral:


server/accounts/models.py
```diff
@@ -266,13 +273,16 @@ def get_active_with_key(self, key):
...
-    user = models.ForeignKey(User, on_delete=models.CASCADE, related_name="api_token")
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
     created_at = models.DateTimeField(auto_now_add=True)
     expiry = models.DateTimeField(blank=True, null=True)
     name = models.TextField(blank=True, default="")
 
     objects = APITokenManager()
 
+    def get_absolute_url(self):
+        return reverse("accounts:user", args=(self.user.pk,))
```
np5: Add anchor "#{self.pk}" and add `id="{{token.pk}}` on the `<tr>`.

tests/server_accounts/test_users_models.py
```diff
@@ -1,7 +1,8 @@
+from datetime import date, timedelta
+
+from accounts.models import APIToken, User
```
np5, line 3: Funny, weird ordering

server/accounts/models.py
np5: Oder "apitoken-{{ pk }}"

tests/server_accounts/test_users_models.py
sebastian-zentral: hm, yes. if we write server.accounts.models it will be recognize as our module. The module path is than wrong but could be a way to fix this.

tests/server_accounts/views/test_api_token_views.py
```diff
@@ -74,12 +68,12 @@ def test_create_api_token_not_self(self, post_event):
 
     @patch("zentral.core.queues.backends.kombu.EventQueues.post_event")
     def test_create_api_token_no_perms(self, post_event):
+        """User can not create token for a service account without the permissions."""
+        self.login()
         service_account = User.objects.create_user(get_random_string(19),
```
np5: Use self.service_account?

tests/server_accounts/views/test_api_token_views.py
```diff
@@ -88,22 +82,30 @@ def test_create_api_token_no_perms(self, post_event):
...
+        """Users are allowed to create token for themselves."""
+        self.login()
+        response = self.client.post(
+                reverse("accounts:create_user_api_token", args=(self.ui_user.id,)), follow=True)
+        self.assertEqual(response.status_code, 200)
+
     def test_create_api_token_form(self):
+        """User can create token for a service account with the required permissions."""
+        self.login("accounts.view_user", "accounts.add_apitoken")
         service_account = User.objects.create_user(get_random_string(19),
```
np5: self.service_account?

np5 (dismissed):
LGTM

PR #1463, opened 2026-04-14
20260414 monolith catalog api

sebastian-zentral:
- Fix filter for catalog list api view (add repository)
- Add pagination for list api view


np5 (changes requested):
Good PR! Don't forget: add a CHANGELOG with breaking change on the API endpoint because it is paginated now.

np5 (approved):
LGTM

PR #1464, opened 2026-04-15
Fix MDM tagging issue

np5:
Only the newest enrollment session with a realm user was considered when tagging. If the newest enrollment session on a device was unauthenticated, but an older authenticated enrollment session existed, the tags were applied.


sebastian-zentral (approved):
lgtm

PR #1465, opened 2026-04-22
google_workspace: remove orphaned machine tags when user leaves all mapped groups

jarrah-anthropic:
## Summary

`sync_group_tag_mappings` does not remove a machine's tag when the principal user leaves *all* mapped groups. Cross-group reconciliation works (a user moving from group A to group B has tag A removed), but a user who leaves every mapped group keeps their managed tags indefinitely.

## Root cause

`_iter_group_members_tags` iterates only `email_tags.items()`, which contains *current* group members. An ex-member's email is not in the dict, so no `(email, tag_pk, False)` removal tuple is yielded and the DELETE in `_sync_machine_tags` never fires for them.

## Repro

1. Map group G → tag T. User U is a member; their machine M has principal_user U.
2. Sync → M tagged T. ✓
3. Remove U from G.
4. Sync → log shows `Found 1 managed tags for N-1 emails` (count dropped) but `removed 0`. Tag T remains on M.

We hit this validating the feature for production use — leaving the access group did not revoke the tag-scoped Santa rule.

## Fix

Add a follow-up DELETE in `_sync_machine_tags` for orphaned tags: `inventory_machinetag` rows where `tag_id` is in the connection's managed set and the machine's current principal user is not in the current member set. Results are unioned into the existing event/count reporting. The original CTE query is unchanged.

Handles the empty-members edge case (all tagged machines cleaned) and leaves tags not managed by any mapping untouched.

## Tests

Two new cases in `tests/google_workspace/test_utils.py`:
- user leaves all groups → tag removed, `removed: 1` reported, `MachineTagEvent` emitted
- orphan removal does not touch unmanaged tags on the same machine

### DIFF
--- a/tests/google_workspace/test_utils.py
+++ b/tests/google_workspace/test_utils.py
@@ -28,11 +28,11 @@ class UtilsTestCase(TestCase):
     def _given_serial_number(self):
         return f"sn_{get_random_string(8)}"
 
-    def _force_machine(self, user_email, serial_number=None):
+    def _force_machine(self, user_email, serial_number=None, source_name='zentral'):
         if not serial_number:
             serial_number = self._given_serial_number()
         MachineSnapshotCommit.objects.commit_machine_snapshot_tree({
-            "source": {'module': 'io.zentral.tests', 'name': 'zentral'},
+            "source": {'module': 'io.zentral.tests', 'name': source_name},
             "serial_number": serial_number,
             "principal_user": {
                 "source": {"type": "LOGGED_IN_USER", "properties": {"method": "System Configuration"}},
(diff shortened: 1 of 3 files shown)


jarrah-anthropic:
Thanks for the single-query rewrite — much cleaner than my two-pass version.

One thing I noticed re-reading it: `execute_values(..., page_size=1000)` runs the full query once per 1000-tuple chunk, and `deleted_tags` removes any managed-tag row not in *that chunk's* `machine_managed_tags`. So with >1000 `(email, tag)` tuples:

- chunk 1 inserts machines 1–1000's tags, then deletes tags for machines 1001+ (their emails aren't in chunk-1's `given_values`, so they're absent from `machine_managed_tags`)
- chunk 2 re-inserts 1001–2000's tags, deletes 1–1000's
- …only the final chunk's machines end up tagged

The previous query was chunk-safe because both INSERT and DELETE were scoped to the chunk's specific tuples; the new "delete what's not in the set" semantics need the whole set at once.

432b39e9 swaps `(values %s)` for `unnest(%(emails)s::text[], %(tag_ids)s::int[])` and a single `cursor.execute`, so the CTE always sees the complete member list. Same query structure otherwise. Happy to adjust if you'd rather handle it differently.

PR #1468, opened 2026-04-30
Increase contrast for text color for background color

andreas-zentral:
The contrast for the text color on light backgrounds is too low. E.g. the yellow `canary` tag comes with a white font color. This PR increases the contrast for such cases.

|Before|After|
|---|---|
|Screenshot 2026-04-30 at 10 48 43|Screenshot 2026-04-30 at 10 48 27|

The Incident Severity tags are using darker text color and a test caught that change. 

|Current|
|---|
|Screenshot 2026-04-30 at 11 22 35|

### Resources

- https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html

### DIFF
--- a/tests/core_incidents/test_incidents_extras.py
+++ b/tests/core_incidents/test_incidents_extras.py
@@ -9,7 +9,7 @@ class IncidentsExtrasTestCase(SimpleTestCase):
                                          (100, "Minor", "DDDD00"),):
             result = (
                 '<span class="label rounded p-1" '
-                f'style="white-space:nowrap;background-color:#{color};color:#FFF">'
+                f'style="white-space:nowrap;background-color:#{color};color:#000">'
                 f'{display}&nbsp;<i class="bi bi-exclamation-triangle-fill"></i>'
                 '</span>'
             )
--- a/zentral/utils/color.py
+++ b/zentral/utils/color.py
@@ -1,5 +1,9 @@
 import re
-import colorsys
+
+
+def _channel_luminance(c):
+    c = c / 255.0
+    return c / 12.92 if c <= 0.03928 else ((c + 0.055) / 1.055) ** 2.4
 
 
 def text_color_for_background_color(color):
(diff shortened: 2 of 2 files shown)


andreas-zentral:
> Prompt: Is the change compliant with https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html                                                                           
   
  The luminance formula is the WCAG one. The sRGB linearization in _channel_luminance (threshold 0.03928, ((c + 0.055) / 1.055) ** 2.4) and the channel weights (0.2126 / 0.7152 /      
  0.0722) match WCAG 2.x's relative-luminance definition.                                                                                                                               
                                                                                                                                                                                        
  AA (4.5:1, normal text): always met. The 0.179 threshold is the crossover where black and white give equal contrast against the background. At that exact point both yield ≈ 4.58:1,  
  and away from it the chosen color does strictly better. Worked out from the WCAG ratio formula (L1+0.05)/(L2+0.05):                                                                   
  - black gives ≥ 4.5:1 when L_bg ≥ 0.175                                                                                                                                               
  - white gives ≥ 4.5:1 when L_bg ≤ 0.1833                                                                                                                                              
  - the intervals overlap, so the better of {black, white} always clears 4.5:1.                                                                                                         
                                                                                                                                                                                        
  AA (3:1, large text): always met by the same argument (intervals [0, 0.30] and [0.10, 1] overlap).
                                                                                                                                                                                        
  AAA (7:1, normal text): NOT guaranteed. Black needs L_bg ≥ 0.30, white needs L_bg ≤ 0.10. Backgrounds with L_bg between 0.10 and 0.30 (a wide band of mid-tones — saturated red       
  #FF0000 is L ≈ 0.21, falling right in this gap) cannot reach 7:1 with either pure black or pure white text. Hitting AAA there would require constraining the tag's background color   
  itself, not just the text color.                                                                                                                                                      
                             
  So: WCAG 2.0 AA-compliant for both normal and large text, given any user-chosen tag background. AAA is out of reach for some background colors no matter what text color we pick.

andreas-zentral:
> Prompt: colorsys import was removed. Can the dependency be removed entirely?                                                                                                                  
                                                                                                                                 
Yes — no other file imports colorsys. It's a Python stdlib module (not a pip package), so there's nothing to uninstall or remove from requirements.txt. The removal of the import line
   in zentral/utils/color.py is the only cleanup needed, and that's already done.

andreas-zentral:
> Prompt: can the formula for the contrast calculation be found in colorsys?                                                                                                                    
                                                                                                                                                                                        
No. colorsys only does color-space conversions: rgb_to_hls, rgb_to_hsv, rgb_to_yiq (and their inverses). None of these is WCAG relative luminance.                                    
                                                     
  The closest is rgb_to_yiq — its Y channel is a different luminance approximation (NTSC weights 0.299 R + 0.587 G + 0.114 B, applied to gamma-encoded sRGB without linearization). It's
   a reasonable heuristic and would likely give the right answer for yellow, but it does not match the WCAG 2.x definition, which requires:                                             
                                                                                                                                                                                        
  1. sRGB → linear-light per-channel (the ((c+0.055)/1.055)^2.4 step), and                                                                                                              
  2. the Rec. 709 weights 0.2126 / 0.7152 / 0.0722.                                                                                                                                     
                                                                                                                                                                                        
  Both pieces are absent from colorsys, so the manual implementation is necessary if we want true WCAG compliance.

np5 (approved):

andreas-zentral:
haha, GitHub. Using the "Rebase branch" button in webUI results in unsigned commits and a force push. 😑

Screenshot 2026-04-30 at 15 20 02

np5 (approved):
LGTM

PR #1469, opened 2026-04-30
Enrollment endpoint

sebastian-zentral:
GET-Endpoint for enrollment information


np5 (changes requested):
All Good, just move the authentication classes to the public_views.py module of each module (inventory & munki).

np5 (approved):
LGTM

PR #1477, opened 2026-05-12
Fix machine snapshot last commit performance

np5:
Sort by `created_at` instead of `id` which helps the PostgreSQL query planner!


sebastian-zentral (approved):
crazy bug

PR #1478, opened 2026-05-12
Fix munki report sorting with naive vs aware datetimes

sebastian-zentral:
add a new datetime util function `parse_naive_datetime`


np5 (approved):
LGTM

PR #1480, opened 2026-05-17
Fix AWS SNS/SQS worker shutdown hang on out_queue back-pressure

np5:
## Summary

- The AWS SNS/SQS `PreprocessWorker` / `EnrichWorker` could hang on shutdown when the run loop raised — e.g. a `django.db.utils.OperationalError` in a preprocessor — leaving the worker process alive and "stuck on AWS SQS".
- Three helper threads put receipt handles into a downstream bounded queue with **no timeout** after a successful AWS call, and did not re-check `stop_event` while blocked:
  - [`SQSSendThread.send_entries`](zentral/core/queues/backends/aws_sns_sqs/sqs.py) → `published_message_queue` (PreprocessWorker)
  - [`SNSPublishThread.run`](zentral/core/queues/backends/aws_sns_sqs/sns.py) → `published_message_queue` (EnrichWorker, 10 threads)
  - [`ConsumerProducerFinalThread.run`](zentral/core/queues/backends/aws_sns_sqs/consumer.py) → `delete_message_queue`
- If the downstream queue was full at the moment the run loop raised, `BaseConsumer.run` would set `stop_event` and then hang forever on `thread.join()`, because the blocked put never observed it.
- Loop each put with `timeout=1` and bail out when `stop_event` is set, mirroring the pattern already used in [`SQSReceiveThread`](zentral/core/queues/backends/aws_sns_sqs/sqs.py).
- On the shutdown branch the receipt handle is dropped — SQS visibility timeout will redeliver the message, which matches the existing tolerance for in-flight handles being leaked from the LRU in `increment_receipt_handle_unpublished_event_count`.

## Test plan

- [ ] `docker compose -f docker-compose.yml -f docker-compose.tests.yml run web tests_with_coverage` (existing AWS SNS/SQS tests still pass)
- [ ] In staging, simulate a SQL outage during preprocessing (e.g. drop the DB connection while a raw event is being processed) and verify the preprocess worker exits within a few seconds instead of hanging
- [ ] Verify dropped-receipt log line appears only on the forced-shutdown path, not in steady state

### DIFF
--- a/tests/core_queues/test_aws_sns_sqs.py
+++ b/tests/core_queues/test_aws_sns_sqs.py
@@ -2,6 +2,7 @@ from datetime import datetime
 import json
 import os.path
 import queue
+import threading
 from unittest.mock import Mock, patch
 import boto3
 from botocore.stub import Stubber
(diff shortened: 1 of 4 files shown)


soysauceconspiracy (dismissed):
LGTM

soysauceconspiracy (approved):
LGTM

PR #1484, opened 2026-05-21
Fix threading issue in Jamf preprocessor

np5:
Do not start an extra thread to send the tagging events.


sebastian-zentral (approved):
lgtm

PR #1490, opened 2026-05-26
Fix multiprocessing / runworkers

np5:
Runworkers is only recommended in docker-compose dev environments, to run all workers in one containers.


sebastian-zentral (approved):
looks good

PR #1493, opened 2026-05-29
2026-05-29 Prevent role escalation in user/service-account forms

np5:
## Summary

- Block non-superusers from adding groups to other users (or themselves) that they aren't already members of — the K8s-RBAC / AWS-permission-boundary escalation-prevention pattern. Removals stay unrestricted by design.
- Shared `RoleMembershipGrantMixin` on `UpdateUserForm` and `ServiceAccountForm` covers both update and create paths; rejection happens in `clean()` and the multi-select renders ungrantable groups as disabled options.
- Side polish: the `groups` field is now labeled "Roles" with help text aligned to PBAC (no more "user will get all permissions granted to each group"), the multi-select is taller (10 rows), and the disabled-options widget moved to `zentral/utils/forms.py` for reuse.

The widget filter is UX only — `disabled` is client-side; the server validation is what enforces the rule. Documented inline.

## Test plan

- [x] `docker compose -f docker-compose.yml -f docker-compose.tests.yml run --rm -e ZENTRAL_CONF_DIR=/zentral/tests/conf -e ZENTRAL_FORCE_ES_OS_INDEX_REFRESH=1 -e ZENTRAL_POLICIES_SYNC=0 -e ZENTRAL_PROBES_SYNC=0 -e ZENTRAL_QUIET=1 -e ZENTRAL_STORES_SYNC=0 web python server/manage.py test --keepdb tests.server_accounts.views.test_users_views` — 104 tests pass (11 new for this PR).
- [x] As a non-superuser with `accounts.change_user`, edit another user: confirm the form labels the field "Roles", that groups you don't belong to render as disabled in the multi-select, and that POST-ing a smuggled value gets rejected with `You cannot grant role membership you don't have: <name>`.
- [x] As a non-superuser, confirm you *can* remove the target from a group you aren't in (asymmetry by design).
- [x] As a superuser, confirm no options are disabled and any group can be added.
- [x] Same checks against `Create service account` (the create path uses the same form).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ Tags cannot be applied to all the machines in a business unit anymore.
 
 The API endpoint for the monolith catalog list is paginated now. Remember to upgrade the Terraform Provider.
 
+#### 🧨 Role grants restricted to the actor's own roles
+
+Non-superusers can no longer grant a user or service account a role they don't belong to themselves — the rule covers every UI path that assigns roles: editing a user, creating a service account, and editing a service account. Removing existing memberships is unaffected. Operators who delegated user management without granting the underlying roles will need to either add those roles to the delegate or perform the assignment as a superuser.
+
 ### Bug fixes
 
 Replaced slow Santa `zentral_santa_targets_*`  metrics with `zentral_santa_target_states` metrics.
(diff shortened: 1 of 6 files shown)


sebastian-zentral (dismissed):
lgtm

andreas-zentral (approved):

PR #1501, opened 2026-06-04
Do not reindent multi-line SQL queries for display

sebastian-zentral:
Do not reindent multi-line SQL queries for display, remove sqlparse


np5 (approved):
LGTM

PR #1503, opened 2026-06-05
Drop format_suffix_patterns from the API URLConfs

sebastian-zentral:
The optional .json/.api format suffix routes are no longer wanted: JSON is already the only configured renderer.


np5 (approved):
LGTM

PR #1506, opened 2026-06-09
Add viewMachineTag to the Viewer action group

sebastian-zentral:
viewMachineTag is custom-registered in inventory/pbac.py, so unlike the model-default view* actions the engine auto-assigns to the Viewer group (_register_model_default_legacy_perm_actions), it only carried Admin and User. A viewer-only role therefore couldn't see machine tags even though the action is read-only. Add ActionGroupBasename.VIEWER so it lands in both the namespace-scoped and global Viewer groups, matching every other view action.


np5 (approved):
LGTM

PR #1513, opened 2026-06-11
Clear inactive enforcement fields in SoftwareUpdateEnforcementForm

sebastian-zentral:
When the Type radio switches between LATEST and ONE_TIME, the inactive side's fields could still carry stale or invalid values from the bound POST data — for example a leftover max_os_version="ABC" would surface a "Not a valid OS version" error even though ONE_TIME ignores it. clean() now wipes cleaned_data, drops any field errors, mutates self.data so a re-render blanks the inputs, and zeroes the instance attributes for the inactive side.

### DIFF
--- a/zentral/contrib/mdm/forms.py
+++ b/zentral/contrib/mdm/forms.py
@@ -1487,19 +1487,21 @@ class SoftwareUpdateEnforcementForm(forms.ModelForm):
 
     def clean(self):
         super().clean()
-        enforcement_type = self.cleaned_data.get("enforcement_type")
-        if enforcement_type == "ONE_TIME":
+        if self.cleaned_data.get("enforcement_type") == "ONE_TIME":
             required_fields = (f for f in self.one_time_fields if f != "build_version")
             other_fields = self.latest_fields
         else:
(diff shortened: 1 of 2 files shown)


zentral/contrib/mdm/forms.py
```diff
@@ -1494,12 +1494,17 @@ def clean(self):
         else:
             required_fields = self.latest_fields
             other_fields = self.one_time_fields
+        self.data = self.data.copy()
+        for field in other_fields:
+            empty = None if field in ("delay_days", "local_time", "local_datetime") else ""
+            self.cleaned_data[field] = empty
+            self._errors.pop(field, None)
+            self.data[self.add_prefix(field)] = "" if empty is None else empty
```
np5: Why?

zentral/contrib/mdm/forms.py
```diff
@@ -1494,12 +1494,17 @@ def clean(self):
...
+            setattr(self.instance, field, empty)
```
np5: Why?

np5 (approved):
LGTM

PR #1518, opened 2026-06-18
Add OpenAPI 3 schema generation (drf-spectacular)

sebastian-zentral:
Generate an OpenAPI 3 schema from the DRF views with drf-spectacular, behind a dedicated OPENAPI config flag (off by default — not for production). When enabled it installs drf-spectacular and exposes the schema, Swagger UI under /api/schema/.

settings: OPENAPI flag, gated SPECTACULAR_SETTINGS. ENUM_NAME_OVERRIDES split up for different modules.

urls: schema/swagger-ui routes, gated on OPENAPI.

templates: project overrides for Swagger UI  adding the CSP nonce and dropping external CDN/font references so the UIs load under Zentral's strict CSP; assets served via npm package.

conf/start: OPENAPI flag, false by default (and DEBUG set to true).

docs: development/openapi.md and the django.OPENAPI configuration entry.

The schema is not yet clean: plain APIViews and dynamically-built serializers still produce validation warnings, to be sharpened per module in follow-up changes.


docs/development/openapi.md
```diff
@@ -0,0 +1,44 @@
+# OpenAPI schema
+
+Zentral uses [drf-spectacular](https://drf-spectacular.readthedocs.io/) to generate an [OpenAPI 3](https://www.openapis.org/) schema from its Django REST Framework views.
+
+## Enabling the schema
+
+The settings and URL wiring is enabled only when the [`django.OPENAPI`](../configuration/django.md#djangoopenapi) flag is `true`. It is `false` by default — do not turn it on in production deployments.
+
+The `devmdm` docker compose stack runs with `OPENAPI: true`; the default stack does not.
```
np5: devmdm is internal.

server/server/settings.py
```diff
@@ -30,6 +33,8 @@
 
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = django_zentral_settings.get('DEBUG', False)
+# SECURITY WARNING: don't run with openapi turned on in production!
+OPENAPI = django_zentral_settings.get('OPENAPI', False)
```
np5: Add & DEBUG = True as extra security?

server/server/settings.py
```diff
@@ -92,6 +97,63 @@
     ),
 }
 
+# OpenAPI schema generation is a development-only concern
+if OPENAPI:
+    INSTALLED_APPS += [
+        'drf_spectacular',
+        'drf_spectacular_sidecar',
```
np5: What does sidecar do? Necessary?

server/server/settings.py
```diff
@@ -92,6 +97,63 @@
...
+    REST_FRAMEWORK['DEFAULT_SCHEMA_CLASS'] = 'drf_spectacular.openapi.AutoSchema'
+
+    SPECTACULAR_SETTINGS = {
+        "TITLE": "Zentral API",
+        "DESCRIPTION": "Management and public APIs for the Zentral MDM platform.",
+        "VERSION": "1.0.0",
+        "COMPONENT_SPLIT_REQUEST": True,
+        "GENERIC_ADDITIONAL_PROPERTIES": 'dict',
+        "SERVE_INCLUDE_SCHEMA": False,
+        "SCHEMA_PATH_PREFIX": r"/(api|public)/",
```
np5: Do not document public, this is only for the agent and should never be implemented or consumed manually.

server/server/settings.py
```diff
@@ -92,6 +97,63 @@
...
+            "deepLinking": True,
+            "docExpansion": "list",
+            "tagsSorter": "alpha",
+        },
+        # Pin component names for choice fields whose auto-derived names
+        # collide. drf-spectacular can't name these on its own because the same
+        # field name ("backend", "platforms", …) carries different choice sets
+        # across modules, or the same choice set is reached via several field
+        # names. Each entry maps the desired component name to its choice set.
+        "ENUM_NAME_OVERRIDES": {
```
np5: Feels weird to have the overrides in a central place. Is it possible to decentralize the overrides?

server/templates/drf_spectacular/swagger_ui.html
```diff
@@ -0,0 +1,54 @@
...
+        width: auto;
+        display: block;
+      }
+    </style>
+    {% endblock head %}
+  </head>
+  <body>
+    {% block body %}
+    <header class="zentral-swagger-header">
+      <img src="{% static 'images/swagger-logo.svg' %}" alt="Zentral">
```
np5: Can we re-use the logo we already have for the UI?

requirements_dev.txt
```diff
@@ -1,3 +1,4 @@
 -c constraints.txt
 
 coverage
+drf-spectacular-sidecar         # locally-served Swagger UI / ReDoc bundles (CSP-friendly)
```
np5: necessary?

server/server/settings.py
np5: Also can break if we install it and we do not have the dev requirements.

np5 (changes requested):
I am wondering if we should include the schema in the commit, and not simply generate it as part of the CI/CD workflow.

server/server/settings.py
sebastian-zentral: Sidecar is removed now

server/server/settings.py
sebastian-zentral: Overrides are now split into the specific modules

requirements_dev.txt
sebastian-zentral: no, gone now

server/server/settings.py
sebastian-zentral: All public endpoints will be removed now

sebastian-zentral:
The openapi schema is removed from the repository. It will be generated by the CI.

PR #1522, opened 2026-06-30
Answer the santactl doctor preflight probe gracefully

np5:
## Summary

`santactl doctor` validates the sync connection by sending a `POST` to `preflight/santactl-doctor-test` — a fixed sentinel machine id that is **not** a real machine UUID — and judges the result purely by HTTP status (200 or 400 → success). See northpolesec/santa: [`SNTSyncManager.mm`](https://github.com/northpolesec/santa/blob/main/Source/santasyncservice/SNTSyncManager.mm) (builds the request) and [`SNTCommandDoctor.mm`](https://github.com/northpolesec/santa/blob/main/Source/santactl/Commands/SNTCommandDoctor.mm) (`validateSync`, status handling).

The probe is sent through the configured sync session, so it carries our `Zentral-Authorization: Bearer <secret>` header and routes to `PreflightView`. But `BaseSyncView.post` parses the machine id as a UUID before anything else, so the sentinel raised `PermissionDenied` → **HTTP 403**. Santa treats 403 as a failure, so `santactl doctor` reported a broken sync connection even against a healthy, correctly-authenticated server.

## Change

- Intercept the `santactl-doctor-test` sentinel in the preflight view **before** the UUID parse (gated to preflight only via `allow_doctor_test`).
- `doctor_test_response()` still validates the enrollment secret — and the client certificate when mTLS is required — then returns **HTTP 200 `{}`**. No enrollment, no machine snapshot, no event is produced (the sentinel is never a real machine).
- Result: a correctly configured client passes (`[+] Preflight request succeeded`); a misconfigured one still gets a meaningful failure (401/403) rather than a false OK.
- The other sync endpoints (ruledownload/eventupload/postflight) keep rejecting the sentinel.

## Tests

Added 5 tests to `tests/santa/test_santa_api_views.py`:
- probe with valid secret → 200 `{}`, no machine created, no event posted
- probe with no auth → 401
- probe with bad secret → 403
- probe with mTLS required but no client cert → 403
- sentinel rejected on a non-preflight endpoint (ruledownload → 403)

Full module passes (`tests.santa.test_santa_api_views`, 52 tests OK).

### DIFF
--- a/tests/santa/test_santa_api_views.py
+++ b/tests/santa/test_santa_api_views.py
@@ -21,6 +21,7 @@ from zentral.contrib.santa.models import (
     Target,
     TargetCounter,
 )
+from zentral.contrib.santa.public_views import BaseSyncView
 from zentral.core.incidents.models import Severity
 from zentral.utils.time import naive_utcnow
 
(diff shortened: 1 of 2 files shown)


soysauceconspiracy (approved):
LGTM

PR #1532, opened 2026-07-10
Fix EventRequestUser.serialize mutating shared session token expiry

np5:
A single request builds one EventRequest and shares it across every event it emits (e.g. bulk machine tag changes through the API). serialize() turned the API token expiry into an ISO string by writing back into the live session dict, so the first event mutated expiry to a str and the next event raised AttributeError ("'str' object has no attribute 'isoformat'") -> HTTP 500. The tag change was already committed and the first event posted, so the client got a 500 for a request that had partly succeeded and the remaining tag events were lost.

Copy the session and token dicts instead of mutating in place, and only call isoformat() on a datetime, so serialize() is idempotent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 2026.5
+
+
+### Bug fixes
+
+Fixed a 500 error on bulk machine tag changes made through the API with an expiring token.
+
+
 ## 2026.4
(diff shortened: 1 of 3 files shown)


soysauceconspiracy (approved):
LGTM

PR #1533, opened 2026-07-12
Fix three unhandled 500s: cert extension parse, DDM status-report dedup, snapshot-commit race

np5:
## What

Three independent fixes for unhandled server errors (HTTP 500s) surfaced while reviewing application error logs. Each is its own commit.

### 1. Guard `is_ca()` against certificates with non-canonical extensions — `zentral/utils/certificates.py`
`is_ca()` accessed `certificate.extensions` outside any handler. `cryptography` raises a plain `ValueError` when an extension isn't DER-canonical (e.g. a `BasicConstraints` `ca` boolean encoded as an explicit default), which escaped and 500'd every machine-snapshot ingestion that parses such a cert (munki `post_job` and the other snapshot endpoints, via `prepare_ms_tree_certificates`). A machine reporting such a cert 500'd its check-in on every run, so its inventory stopped updating. A cert whose extensions can't be parsed is now treated as not-a-CA.

### 2. Dedup DDM status-report artifacts by version — `zentral/contrib/mdm/declarations/status_report.py`
A device can report the same artifact version under two server-tokens (the token carries retry / OS-version / reinstall-interval suffixes) during a re-push. Both collapse to one `artifact_version_pk`, so the target-artifact bulk upsert's `ON CONFLICT (target, artifact_version_id)` rejected the two rows with a `CardinalityViolation`, 500-ing the declarative-management status check-in (`/public/mdm/checkin/`). The status-report parser now keeps one entry per `artifact_version_pk` — the most present of the reported statuses (ranking exposed as `TargetArtifact.Status.presence_rank`) — and logs a warning when it collapses a duplicate.

### 3. Skip the machine-snapshot commit on a concurrent version race — `zentral/contrib/inventory/models.py`
Two near-simultaneous check-ins from the same machine + source can both pick the same next `MachineSnapshotCommit` version; one wins, the other hits the `(serial_number, source, version)` unique constraint. The manager already returned quietly when the winner had stored an identical snapshot, but raised `MTOError` (→ 500) when it had stored a different one (e.g. santa preflight). A differing snapshot just means a tree field changed inside the race window — the current snapshot is already up to date and the next check-in reconciles — so it now logs and skips instead of raising.

## Scope / not included
- Fix 3 addresses the commit race specifically. A separate, larger source of preflight 500s observed in production is nginx→gunicorn connection refusal under worker starvation — an availability/capacity matter, not an application bug.
- A monolith repository-sync get-then-create race (`_import_pkg_info` / `_import_catalogs`) is real but rare; it'll get its own PR (advisory-lock fix), since making it right spans several sites and a per-repo sync lock is the cleaner approach.

## Tests
Each fix ships focused tests (`tests/utils/test_certificates.py`, `tests/mdm/test_declarations_utils.py`, `tests/inventory/test_machine_snapshot.py`), exercising the new branches. Run with the project's docker-compose test setup.

### DIFF
--- a/tests/inventory/test_machine_snapshot.py
+++ b/tests/inventory/test_machine_snapshot.py
@@ -1,8 +1,10 @@
 import copy
 from datetime import datetime, timedelta
+from unittest.mock import Mock, patch
 
 from dateutil import parser
 from django.core.cache import cache
+from django.db import IntegrityError
 from django.test import TestCase, override_settings
 from django.utils.timezone import is_aware, make_naive
 
@@ -15,6 +17,7 @@ from zentral.contrib.inventory.conf import (
     VM,
     os_version_display,
     os_version_version_display,
+    update_ms_tree_platform,
     update_ms_tree_type,
 )
 from zentral.contrib.inventory.models import (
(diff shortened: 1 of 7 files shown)


soysauceconspiracy (approved):
LGTM

PR #1534, opened 2026-07-12
Fix uncommittable machine snapshot trees: subtree dedup, poison raw-event drop, dead serialize

np5:
Machines occasionally report the same `osx_app_instance` twice. The resulting `MTOError` made the whole snapshot uncommittable, and on the async path (munki, osquery) the raw event was redelivered forever, crashing the preprocess worker on each attempt. Three commits:

### 1. Dedup identical subtrees in `prepare_commit_tree` instead of failing

The subtree lists map to many to many relationships, i.e. sets: a duplicated subtree carries no information and could never round-trip its hash. Duplicated subtrees are now skipped with a warning, and the list is replaced by the original filtered through the skipped indexes. Only subtrees are deduplicated, never the scalar items. The happy path costs nothing: the skipped indexes are only collected, and the list only rebuilt, when a duplicate is found. Scalar items — the only index-sensitive ones — hash their index via the position in the hash list, i.e. their index in the filtered list, keeping the tree hash consistent with the stored value.

### 2. Drop machine snapshot trees on deterministic preprocessing errors

In the machine snapshot preprocessor, only recoverable DB errors (`InterfaceError`, `OperationalError`) are retried, now via `RetryLater` instead of a worker crash. Anything else logs a single error line — source name, serial number, error message, no traceback — and drops the tree; machine snapshots are periodic full reports, so a dropped tree costs one check-in. `MTOError` now calls `super().__init__` so that `str(error)` is not empty, and `commit_machine_snapshot_and_yield_events` leaves error handling to its callers.

### 3. Remove dead full `serialize()` before the `add_machine` event

The discarded `serialize()` walked every m2m relation (certificates, osx_app_instances, …) for each new machine snapshot, only for the result to be thrown away; the event payload is built by the following `serialize(exclude=...)` call.

All changed lines are covered by tests (`tests.inventory`: 515 OK).

### DIFF
--- a/tests/inventory/test_machine_snapshot.py
+++ b/tests/inventory/test_machine_snapshot.py
@@ -32,7 +32,7 @@ from zentral.contrib.inventory.models import (
     Tag,
 )
 from zentral.contrib.inventory.utils.db import inventory_events_from_machine_snapshot_commit
-from zentral.utils.mt_models import MTOError
+from zentral.utils.mt_models import MTOError, prepare_commit_tree
 from zentral.utils.time import naive_utcnow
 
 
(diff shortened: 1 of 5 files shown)


soysauceconspiracy (approved):
LGTM

PR #1547, opened 2026-07-24
Add code signing fields to the inventory OSXAppInstance

sebastian-zentral:
Add team_id, cd_hash, entitlements, signing_time and secure_signing_time to Inventory::OSXAppInstance so app instances can carry Apple code signing information. The fields participate in the Merkle tree hash like the existing ones, so they are part of the app instance identity; existing snapshots are unaffected, as empty values are excluded from the hash.

Surface the fields wherever OSXAppInstance data is exposed:
- machine macOS apps page: a "Code signing" column for team_id, cd_hash and the signing timestamps, plus an entitlements modal (jsonprettyprint)
- full JSON export: all five fields (entitlements as jsonb text, like  extra_facts)
- per-machine CSV export: the four scalar fields (entitlements omitted)

Producers supply the values in the machine snapshot tree; datetimes must be naive UTC to keep the MT hash consistent.

Extend the inventory tests (model commit/serialize/diff, exports, machine apps view) to cover the new fields.


soysauceconspiracy (approved):
LGTM

PR #1551, opened 2026-07-28
Serialize datetimes & UUIDs in event payloads: kombu fallback, Turbo pks, enrollments

np5:
Five commits, each independently green, for values that reach the event pipeline unserialized. Rebased on `main` (`0b65b6d4`).

## 1. `str()` the Turbo UUID pks (`zentral/contrib/turbo/models.py`)

`Configuration`, `Job`, `Script`, `MSCPCheck`, `RecurringJob` and `OneTimeJob` have UUID primary keys, so `serialize_for_event` was putting `uuid.UUID` objects in the event payloads — six sites, including the nested `job.pk` on both schedulers.

Not limited to the audit events: `Configuration.serialize_for_event(keys_only=True)` is what `_add_enrollment_keys` attaches to **every** agent request, enrollment and result event, so the configuration pk was enveloped on the machine-facing path too.

`AuditEvent.build` already stringifies the `object.pk` it sets itself, which left a single event carrying its configuration pk in two different shapes — visible in the test assertions before this change.

The Turbo datetimes were already isoformatted, and the agent-protocol views already stringify their UUIDs and run times deliberately; that commit is UUID-only.

## 2. Serialize the enrollment datetimes (`inventory` + `mdm`)

- **`BaseEnrollment` / `EnrollmentSecret`** (`zentral/contrib/inventory/models.py`) — `BaseEnrollment.serialize_for_event` returned its `created_at` raw, and `EnrollmentSecret.serialize_for_event` returned `created_at`, `revoked_at` and `expired_at` raw through its `getattr` loop. Every enrollment audit event in munki, osquery, santa, monolith and turbo carried them, and so did the mdm ones — `MDMEnrollment` is a separate hierarchy, but it composes the secret block in.
  The secret loop converts by type (`isinstance(val, datetime)`) rather than by attribute name: the timestamps sit in a literal tuple next to `is_revoked` / `is_expired`, and a name test would either duplicate that list or key off an `_at` suffix a future field could break.

- **The MDM enrollments** (`zentral/contrib/mdm/models.py`) — `MDMEnrollment`, `DEPEnrollment`, `UserEnrollment`, `EnrollmentCustomView` and `DEPEnrollmentCustomView` all returned `created_at` / `updated_at` raw. `OTAEnrollment` is covered by the base class it calls.

## 3. Serialize the MDM enrollments through their base class (`zentral/contrib/mdm/models.py`)

Fixing the timestamps surfaced why there were five copies to fix: `MDMEnrollment.serialize_for_event` was only ever used by `OTAEnrollment`. `DEPEnrollment` and `UserEnrollment` each reimplemented it and both had drifted, dropping `realm` — which `UserEnrollment` always has, its realm being required. `DEPEnrollment` also omitted the enrollment secret its two siblings include.

The base now owns the whole payload — pk, name, realm, timestamps, enrollment secret — plus the `keys_only` form `DEPEnrollmentCustomView` needs. `OTAEnrollment` and `UserEnrollment` drop their overrides; `DEPEnrollment` keeps one only to add the DEP profile `uuid`, in both forms as before.

**This one changes payload shape**, and it reaches the enrollment session event streams too, since all three `*EnrollmentSession.serialize_for_event` methods embed the full enrollment form:

| | before | after |
|---|---|---|
| `OTAEnrollment` | pk, name, realm, timestamps, secret | unchanged |
| `UserEnrollment` | pk, name, timestamps, secret | **+ realm** |
| `DEPEnrollment` | pk, uuid, name, timestamps | **+ realm, + secret** |

Additive only, so payload-keyed probes and filters keep working. The secret block has never contained the secret value itself — only pk, quota, counts, tags, mbu and timestamps — and OTA/user events have always carried it.

Worth knowing for review: the existing DEP tests could not have caught this, because they build the expected payload by calling `serialize_for_event` itself. The new tests assert the shared shape directly.

## 4. Register non-enveloping kombu JSON encoders (`zentral/utils/json.py`)

Every queue backend and every store backend except ES/OpenSearch serializes with kombu's JSON encoder, which never raises on a datetime or a UUID — it wraps the value in a `{"__type__": …, "__value__": …}` envelope. That round-trips inside Zentral, which is why this goes unnoticed, but:

- the envelope reaches the stores that dump the event themselves (ClickHouse, Splunk, Panther, Sumo Logic, Datadog, Kinesis, S3 Parquet, HTTP) as a nested object;
- the ES/OpenSearch client isoformats the same value, so one payload field ends up with **two shapes depending on the backend**;
- on ClickHouse each miss costs two dynamic JSON subpaths (`payload.x.__type__` / `.__value__`) that no `payload.x.:String` query matches.

`register_kombu_json_encoders()` re-registers `datetime`, `date`, `time` and `UUID` with a `None` marker — kombu's documented "pure transformation, no envelope" mode — so a miss lands in the stores as the plain string they expect. It is called from `zentral/core/events/__init__.py`, the one package every event class and both ends of the pipeline import, so it always runs before an event is serialized.

A net, not a licence: payloads are still expected to serialize at the source, which is what the other commits do.

## Notes for the reviewer

- **Registration order is load-bearing.** `datetime` is a `date` subclass and the encoder takes the first `isinstance` match, so registering `date` first would silently truncate datetimes to their date. Re-registration updates kombu's dict in place and preserves its datetime-first order; a test pins the time component so a reordering fails loudly.
- **Decoders are left registered**, so a message already enveloped in a queue keeps deserializing across the deploy.
- **Celery is unaffected**: no `apply_async` / `delay` call in the repo passes a datetime argument.
- **Decimal is deliberately untouched** — no model declares a `DecimalField`.
- This retires the `# this one is a bug` / `# not ideal fallback` comments in `tests/core_stores/test_clickhouse.py`, which documented the envelope shape.
- `BaseTurboEvent.get_linked_objects_keys()` reads pks back out of the payload, so its key tuples are now strings. `linked_objects_keys_for_event()` still returns raw pks: `encode_args` calls `str()` on them at serialize time, and core's own `AuditEvent.build` passes raw pks the same way.

## Still outstanding (not in this PR)

A wider audit found roughly 80 more raw-datetime insertions in `serialize_for_event`, in the non-enrollment parts of mdm (artifacts, blueprints, commands, configs, `SoftwareUpdate`'s `DateRangeField` bounds) and in monolith, santa, munki, probes, realms, accounts and `zentral/utils/backend_model.py` — plus the `target_artifact_update` payload built from a raw cursor in `zentral/contrib/mdm/artifacts.py`. Commit 4 keeps all of them out of the stores; correcting them at the source is follow-up work.

## Tests

- New guard in `tests/turbo/test_models.py` asserting each Turbo payload is stdlib-`json` encodable — the stdlib encoder raises where kombu's does not, so it catches both classes of leak on any future model — plus explicit `str(pk)` assertions.
- New `EnrollmentSecret` / `BaseEnrollment` tests in `tests/inventory/test_models.py`, covering the revoked/expired timestamps and the same stdlib-`json` guard.
- New `tests/mdm/models/test_enrollment_models.py` covering all five MDM enrollment serializers, the shared base payload and the `keys_only` form.
- New `KombuJSONEncodersTestCase` in `tests/utils/test_json.py` covering the four types, the datetime/date ordering trap, and enveloped-message decoding.
- Updated assertions: 7 across 5 Turbo test files, plus `tests/core_stores/test_clickhouse.py`, `tests/santa/test_setup_views.py` and three `tests/mdm/` files.
- Full suite on the rebased branch: `7394 tests, OK`.


soysauceconspiracy (approved):
LGTM

PR #1552, opened 2026-07-28
Compress the raw events too big for the queue, and stop building oversized SQS batches

np5:
Two commits: raw events too big for the queue are now compressed, and the SQS send thread no longer builds batches that the queue rejects whole.

## Why

Machine snapshot trees carrying a full app inventory serialize to more than SQS accepts. The queue rejects the message, the tree never reaches the preprocessor, and the snapshot is lost.

The per-message limit is not the only one that bites. SQS applies the same 1 MiB cap to the *sum* of a `send_message_batch`, and `SQSSendThread` batched ten messages with no size accounting — so one big message took the nine it happened to travel with down with it. `send_entries` clears its entries whatever the outcome, and on the producer path there is no receipt handle to re-enqueue, so all ten raw events were simply lost, behind a single log line.

## 1. Compress the raw events too big for the queue with zstd

- New [`zentral/core/queues/compression.py`](https://github.com/zentralopensource/zentral/blob/compress-big-raw-events/zentral/core/queues/compression.py): a raw event whose serialized form reaches 16KB is posted as one base64'd zstd blob under a dedicated `zstd_raw_event` key.
- The three backends call `compress_raw_event_if_needed()` in `post_raw_event`; `iter_preprocessed_events()` inflates the blob — the single entry point every backend already runs its preprocessors through.
- Producers and preprocessors are untouched: `MachineSnapshotPreprocessor` still receives the `{"ms_tree": …}` dict it always got, and smaller raw events go on the wire unchanged, still readable in the queue.

Notes for reviewers:

- **The threshold is 16KB, not the message limit.** A full batch of messages at the threshold stays far inside the cap, so the fix doesn't depend on getting that number right.
- **The blob is built with kombu's encoder**, the one every backend serializes raw events with, so what inflates is exactly what an uncompressed raw event deserializes to. `tests/core_queues/test_compression.py` pins that for datetimes and UUIDs.
- **All three backends compress, kombu included**, so dev on RabbitMQ exercises the same envelope path as prod on SQS.
- **An undecompressable blob is logged and dropped**, not re-enqueued — a retry would fail the same way, consistent with how the inventory preprocessor already treats deterministic errors.
- **base64, not base85.** Nothing outside the pipeline reads these blobs, so a denser encoding was on the table — 5/4 instead of 4/3, and the b85 alphabet needs no JSON escaping either. Measured, it isn't worth it: `b85encode`/`b85decode` are pure Python where the base64 pair has a C fast path, ~4.6ms to encode and ~6.3ms to decode a 90KB blob against 0.06ms either way. That would make the encoding the most expensive step in the path (zstd itself is 1.2ms) and put the 6.3ms on the preprocess worker, for 6% off the wire. A test pins the property that makes base64 free here: the wire size is the payload plus the wrapper and nothing more, because its alphabet contains no character JSON escapes.
- **No new dependency**: `compression.zstd` is stdlib as of Python 3.14, and the image is `python:3.14-trixie` (libzstd 1.5.7).

## 2. Split the SQS batches by size, not only by count

Compression cut the batch exposure by ~9x but did not remove it: a tree past ~4,270 apps still comes out over a tenth of the cap once compressed and base64'd, and ten of those overflow a batch again. `SQSSendThread` now accounts for the bytes as entries accumulate and sends what fits before adding one that would cross the cap, instead of always counting to ten.

- A message over the cap on its own can't be sent at all, here or anywhere else. It now travels alone, so it only takes itself down, and it's logged with its size and routing key.
- Message attributes count against the size too, so the routing key is part of the accounting.
- The budget keeps a margin under 1 MiB — the SQS-side accounting isn't byte for byte the one in `_entry_byte_size`, and being a few bytes optimistic would cost a whole batch.

This part is a pre-existing bug that predates compression; it's here because compression is what made the remaining window small enough to be worth closing properly.

## Measurements

Effective ratio is `0.75 × zstd ratio` — base64 is a flat 4/3, plus 22 bytes of JSON wrapper. On modelled munki trees, with the batch budget at 1,038,336 bytes:

| tree | raw | on the wire | effective | per batch | batch bytes |
|---|---|---|---|---|---|
| 500 apps | 107,558 | 10,626 | 10.1x | 10 | 106,260 |
| 2,000 apps | 430,058 | 46,110 | 9.3x | 10 | 461,100 |
| 5,000 apps | 1,075,058 | 121,422 | 8.9x | 8 | 971,376 |
| 20,000 apps | 4,300,058 | 500,050 | 8.6x | 2 | 1,000,100 |

The largest tree that still lets ten of its kind share one batch is ~4,271 apps. Past that the send thread splits the batch instead of having it rejected whole.

Verified end-to-end through the dev RabbitMQ, publishing a real 5,000-app tree via `post_raw_event` and inflating it in `iter_preprocessed_events`: 881,801 bytes → 22,634 on the wire (that tree is repetitive, hence the better-than-table ratio), preprocessor receives the original tree.

## Testing

New tests cover the round trip, the threshold boundary in both directions, encoder transparency, inflation ahead of the preprocessor, the no-JSON-escaping property, the corrupt-blob drop, one case per backend asserting a big raw event goes out compressed, and for the send thread: ten small messages still batching together, a byte-limit split, attribute bytes in the count, and the oversized-message-alone path.

```
docker compose -f docker-compose.yml -f docker-compose.tests.yml run --rm \
    -e ZENTRAL_CONF_DIR=/zentral/tests/conf -e ZENTRAL_FORCE_ES_OS_INDEX_REFRESH=1 \
    -e ZENTRAL_POLICIES_SYNC=0 -e ZENTRAL_PROBES_SYNC=0 -e ZENTRAL_QUIET=1 -e ZENTRAL_STORES_SYNC=0 \
    web python server/manage.py test --keepdb \
    tests.core_queues tests.inventory tests.munki tests.turbo tests.puppet tests.wsone
```

1,373 tests, all passing.

## Deploying

A preprocess worker still on the previous release drops a compressed raw event as an unparsable tree, so restart the preprocess workers before the web workers.

### DIFF
--- a/tests/core_queues/test_aws_sns_sqs.py
+++ b/tests/core_queues/test_aws_sns_sqs.py
@@ -26,9 +26,12 @@ from zentral.core.queues.backends.aws_sns_sqs import (
     ProcessWorker,
     SimpleStoreWorker,
 )
+from zentral.core.queues.compression import COMPRESSED_RAW_EVENT_KEY, decompress_raw_event
 from zentral.core.stores.backends.http import HTTPStoreSerializer
 from zentral.core.stores.backends.s3_parquet import S3ParquetStoreSerializer
 from zentral.core.stores.models import Store
+
+from .test_compression import build_big_raw_event
 
 
 class AWSSNSSQSQueuesTestCase(TestCase):
@@ -746,6 +749,19 @@ class AWSSNSSQSQueuesTestCase(TestCase):
         self.assertIs(args[2], eq._raw_events_queue)
         SQSSendThread.return_value.start.assert_called_once()
         self.assertEqual(eq._raw_events_queue.qsize(), 2)
+
+    @patch("zentral.core.queues.backends.aws_sns_sqs.EventQueues._setup_graceful_stop")
+    @patch("zentral.core.queues.backends.aws_sns_sqs.EventQueues.setup_queue")
+    @patch("zentral.core.queues.backends.aws_sns_sqs.SQSSendThread")
+    def test_post_raw_event_compresses_big_raw_event(self, SQSSendThread, setup_queue, setup_graceful_stop):
+        setup_queue.return_value = "https://example.com/rq"
+        eq = self.get_queues()
+        raw_event = build_big_raw_event()
+        eq.post_raw_event("routing-key", raw_event)
+        _, routing_key, posted_raw_event, _ = eq._raw_events_queue.get_nowait()
+        self.assertEqual(routing_key, "routing-key")
+        self.assertEqual(list(posted_raw_event), [COMPRESSED_RAW_EVENT_KEY])
+        self.assertEqual(decompress_raw_event(posted_raw_event), raw_event)
 
     @patch("zentral.core.queues.backends.aws_sns_sqs.setup_signal_handler")
     def test_setup_graceful_stop_wires_signals_once(self, setup_signal_handler):
--- a/tests/core_queues/test_aws_sns_sqs.py
+++ b/tests/core_queues/test_aws_sns_sqs.py
@@ -3,6 +3,7 @@ import os.path
 import queue
 import signal
 import threading
+import time
 from datetime import datetime
 from unittest.mock import Mock, patch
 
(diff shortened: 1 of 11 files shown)


soysauceconspiracy (approved):
LGTM

PR #1553, opened 2026-07-28
Cache and prefetch the MT objects when committing an inventory tree

np5:
Committing an inventory tree walks every subtree and looks each one up by `mt_hash`, **once per occurrence**. An unchanged tree is already cheap — the root hash hits and nothing below it is walked — but any change means a new root hash and a full walk, and a machine snapshot inlines the same signing certificate chain in every one of its app instances.

Measured with a synthetic snapshot tree, counting queries per `commit()`:

| scenario | `main` | + cache/prefetch | + full_clean skips |
|---|---:|---:|---:|
| first commit, 200 new app instances | 2629 | 2033 | **1229** |
| re-commit with 1 new app instance | 234 | 35 | **29** |
| unchanged re-commit | 1 | 1 | **1** |

Four commits, each independently green and at 100% on the module. Batching the inserts is the follow-up in #1554, stacked on this branch — it is a much bigger change and wants its own review.

## 1. Cover the untested MT object paths and drop a dead branch

`zentral/utils/mt_models.py` sat at 94%, so this brings it to 100% before anything changes underneath: the `Hasher` value errors, the tree-shape errors, the concurrent-creation recovery in both directions, the post-save hash mismatch, the `get_mt_field` guards, and two `serialize()`/`diff()` branches.

It also drops the `bytes` branch of `Hasher.hexdigest()`, which was unreachable rather than untested — `add_field` only ever stores a `str` (ints and datetimes are converted) or a list of hex digests, and raises on anything else. **No digest changes**, which matters since `mt_hash` is persisted.

## 2. Cache and prefetch the MT objects when committing a tree

`MTObjectManager.commit()` threads an `MTCommitCache` through the recursion, seeded by one chunked `mt_hash__in` lookup per model. `prepare_commit_tree()` already resolves the related model of every subtree, so it can collect them while it computes their hashes — no second walk, no change to the tree format.

- **The prefetch only runs once the root hash has missed.** Doing it eagerly would have turned the unchanged-tree path from 1 query into ~10 plus thousands of fetched rows. That path is still exactly one query, and a test pins it.
- **The cache is keyed by `(model, mt_hash)`, not by the hash alone.** `mt_hash` is only unique per table, and it does collide across tables here: `BusinessUnit` and `MachineGroup` share their hashable field set (`key` excluded on both, `meta_business_unit` on `BusinessUnit`), and a machine snapshot tree carries both — `business_unit` and `groups`. A hash-only cache hands a `BusinessUnit` to the `groups` m2m. Regression test included.

## 3. Skip the full_clean queries that cannot fail

`full_clean()` runs *after* `obj.save()`, which makes both of its query-generating checks dead weight — 2 queries per created object:

- `validate_unique` re-checks `mt_hash` — a unique column, excluding the row's own pk, right after our insert succeeded. Since the insert comes first, the DB has already enforced every unique constraint on the object, so the check can never find anything.
- The foreign keys committed from the tree were fetched or created a few frames up, so `ForeignKey.validate`'s existence query asks about rows we are holding.

Only the foreign keys set from the tree are excluded, which is exactly the set where the existence query is the only remaining check: a non-empty value passes the null and blank checks, and these fields declare no choices or validators. Choices, `max_length` and blank validation still run on the agent-supplied values, with a test asserting that a bad `platform` is still rejected. A required foreign key missing from the tree is still rejected by the not-null constraint, exactly as before — `save()` precedes `full_clean()`, so that case never reached the validation to begin with, and a test pins the ordering.

## 4. Roll back the row when the committed object hash does not match

The post-save verification sat in the `else:` of the try, which runs after `transaction.atomic()` has already committed. A mismatch raised `MTOError` but left the row behind, in a table where the `mt_hash` **is** the identity — every later commit of that subtree would resolve to it. Moving the check inside the block takes the row down with the exception.

## Memory

The cache holds model instances, so it is bounded three ways: capped at `MAX_COMMIT_CACHE_SIZE` (10000) with a graceful fall back to the uncached path past it, lookups chunked at 1000 so there is no giant `IN` list, and its lifetime is one top-level `commit()` call. The collector is only 40-char strings — strictly smaller than the tree already resident in memory. Nothing is cached across requests, which would go stale against `cleanup_inventory`.

The cap applies to *new* keys only: a prefetched absence must always be upgradable to the committed object, otherwise every later occurrence retries the insert and burns a savepoint rollback.

## Notes for the reviewer

- Behaviour-preserving. The new `_cache` kwarg is private and keyword-only in practice; no caller passes a positional second argument, and `BusinessUnit.objects.commit(tree, meta_business_unit=...)` still reaches `**extra_obj_save_kwargs`.
- The `issubclass(model, AbstractMTObject)` guard in the collector is deliberate: an excluded FK can point at a model without an `mt_hash` (`BusinessUnit.meta_business_unit`), and collecting it would turn a clear `MTOError` into a `FieldError` from the prefetch.
- Independent of any change to the ingestion payload format — a deduplicated wire format would benefit from this, but does not need it.
- Commits 1 and 4 are behaviour fixes and could be split off; 2 and 3 build on each other.

## Verification

- Full suite: **7389 tests, OK**.
- `zentral/utils/mt_models.py`: **100%**, 0 of 311 statements missed, measured over the full suite.
- `ruff` clean.

## What is left

Of the 1229 queries left in the first-commit case, only ~400 are the inserts themselves. The rest is ~800 `SAVEPOINT`/`RELEASE` statements — one `transaction.atomic()` per committed object. Collapsing those needs `bulk_create`, which the collector added here makes possible: that is #1554, which takes the same tree to 34 queries.

### DIFF
--- a/tests/inventory/test_machine_snapshot.py
+++ b/tests/inventory/test_machine_snapshot.py
@@ -29,10 +29,11 @@ from zentral.contrib.inventory.models import (
     MachineTag,
     MetaMachine,
     Source,
+    SourceManager,
     Tag,
 )
 from zentral.contrib.inventory.utils.db import inventory_events_from_machine_snapshot_commit
-from zentral.utils.mt_models import MTOError, prepare_commit_tree
+from zentral.utils.mt_models import Hasher, MTOError, prepare_commit_tree
 from zentral.utils.time import naive_utcnow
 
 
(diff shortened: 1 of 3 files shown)


soysauceconspiracy (approved):
LGTM

PR #1557, opened 2026-07-30
Audit the MDM command creation paths without serializing secrets

np5:
## Why

Creating an MDM command left no trace. Deleting a queued one was already audited, so the trail recorded the destruction of commands but not their creation — an operator could erase a device, lock it, or push an arbitrary custom command, and nothing recorded who did it.

Two entry points create commands on an operator's behalf:

| | View | Covers |
|---|---|---|
| UI | `CreateEnrolledDeviceCommandView` (`views/management.py`) | all 11 registered manual commands |
| API | `CreateEnrolledDeviceCommandView` (`api_views/enrolled_devices.py`) | `erase/`, `lock/`, `send_custom_command/` |

## What changed

**Both paths now post a `zentral_audit` event with the `CREATED` action**, following the `CreateViewWithAudit` / `ListCreateAPIViewWithAudit` convention — built with `AuditEvent.build_from_request_and_instance` and posted in `transaction.on_commit`, so a rolled-back creation emits nothing. The audit sits in the shared base class of each entry point, so the three API subclasses and any future one are covered by construction.

**`AuditEvent.build` takes an optional `machine_serial_number`.** It previously only passed tags/uuid/index/request, so an audit event for a `DeviceCommand` never reached the machine's event timeline. Additive with a default; other contrib modules can use it.

**Command parameters are now serialized through an allowlist.** They're where the audit value is and also where the secrets are, so `Command.serialize_for_event()` reports them via two new declarations on the command classes:

```python
audit_public_kwargs = ()   # serialized verbatim
audit_secret_kwargs = ()   # "<redacted>"
                           # anything undeclared -> "<omitted>"
```

A class that declares nothing is silent rather than leaky, so a new secret kwarg can't leak by being forgotten. `<redacted>` tells an auditor a value was supplied without disclosing it. Both sentinels are plain strings, so a field never changes type between events and the event stores' dynamic mappings stay stable.

Declared across 15 classes from the kwargs each actually writes. Secrets: `PIN` (DeviceLock, EraseDevice), `new_password` (SetFirmwarePassword, SetRecoveryLock, SetAutoAdminPassword), `admin_password` (AccountConfiguration), `encryption_key` (RotateFileVaultKey — an encrypted RSA private key).

**`CustomCommand` overrides the builder.** Its payload is a property list typed by an operator and can hold anything, so only the shape is audited — `RequestType` plus the names of the other top level keys, never a value. An unparsable payload degrades to `<omitted>`.

```json
{"action": "created",
 "object": {"model": "mdm.devicecommand",
            "new_value": {"name": "DeviceLock",
                          "kwargs": {"Message": "Lost laptop", "PIN": "<redacted>"}}}}
```

## Notes for review

**Why redaction and not a digest.** Device lock PINs are six digits (`pin_regex = r"[0-9]{6}"`), so an unkeyed digest sitting in an event store is reversible by exhaustive search — equivalent to logging the PIN. Hashing the ciphertext instead would correlate nothing, since the secret engine's output isn't deterministic. Keyed HMAC fingerprints can be added later behind the same declarations if cross-fleet correlation ("same firmware password on 500 devices?") becomes worth the key management.

**Engine-created commands are deliberately out of scope.** The ~22 creation sites in `commands/scheduling.py` and in command response processing run while a device checks in, with no operator to attribute them to. Auditing the per-check-in inventory refreshes would also dwarf the operator events on a large fleet. `request.view` distinguishes the two audited paths (`mdm:create_enrolled_device_command` vs `mdm_api:lock_enrolled_device`), so no separate provenance field was needed.

**Existing payload assertions are unchanged.** Parameterless commands get no `kwargs` entry, following the surrounding convention of omitting falsy values.

**The registry import in `models.py` is lazy** because `commands` imports `models`. Importing the submodule runs the package `__init__`, which is what registers every command; an unknown command name fails closed to all-omitted.

## Tests

2980 tests pass across `tests.mdm`, `tests.core_events`, `tests.core_stores` and `tests.core_probes`. flake8 clean.

New `tests/mdm/test_command_audit.py` (13 tests). The structural ones:

- `test_no_encrypted_kwarg_is_missing_from_this_list` scans the command modules for the `field="..."` names in their `get_secret_engine_kwargs` helpers and fails if a new encrypted field appears unclassified — the guard that should prevent the next leak.
- `test_undeclared_and_secret_kwargs_are_never_serialized` feeds every registered class a kwargs dict of sentinels and asserts the sentinel only survives under an explicitly public key.
- Plus no-double-declaration and no-encrypted-kwarg-is-public across the whole registry.

Four view-level tests assert the real events: the PIN and a custom-command `UnlockToken` never appear anywhere in the payload, request metadata carries IP / method / path / view / user / token-auth, and an invalid form posts no event at all.

## Not verified

`mkdocs build --strict` could not be run — mkdocs isn't in the container image and `pip install` fails there on venv permissions. The new docs section was checked by hand instead: no links or images (so strict mode's broken-anchor and missing-file classes don't apply), balanced fences, consistent heading levels, and the JSON example parses. Worth one real run before merge.

### DIFF
--- a/docs/apps/mdm.md
+++ b/docs/apps/mdm.md
@@ -504,6 +504,45 @@ A Recovery Password Configuration can only be deleted if it is not linked to any
 3. Use the *Delete* button in the list view or on the configuration details page.
 4. Confirm the deletion when prompted.
 
+## Command audit trail
+
+Every command an operator creates is recorded as a `zentral_audit` event with the `created` action. This covers both entry points:
+
+- the *New command* forms in the web interface, under a device's detail page
+- the `erase/`, `lock/` and `send_custom_command/` HTTP API endpoints
+
+Each event carries the full request context — the authenticated user (including whether an API token was used, and which one), the source IP, the user agent, the HTTP method, the path and the resolved view name — alongside the serialized command. Events are tagged with the device serial number, so they also appear in the device's event timeline, and are linked to the command UUID.
+
+Deleting a queued command is audited the same way, with the `deleted` action.
+
+### Command parameters and secrets
+
+The command parameters are included in the audit event, but a parameter is only serialized verbatim if the command explicitly declares it as safe to log. Every other parameter is reported by name with a placeholder value, so the audit trail records that a parameter was supplied without disclosing it:
+
+- `<redacted>` — a secret, such as the PIN of a *Device lock* or *Erase device* command, or the new password of a *Set firmware password*, *Set recovery lock* or *Set auto admin password* command
+- `<omitted>` — a parameter that is not declared as safe to log
+
+Commands that carry no parameters have no parameters entry at all.
+
+*Custom* commands are a special case, because their property list payload is written by an operator and can hold anything. Only the shape of the payload is audited: its `RequestType` and the names of the other top level keys, never their values.
+
+```json
+{
+  "action": "created",
+  "object": {
+    "model": "mdm.devicecommand",
+    "pk": "42",
+    "new_value": {
+      "name": "DeviceLock",
+      "uuid": "8f2c1b9e-...",
+      "kwargs": {"Message": "Lost laptop", "PIN": "<redacted>"}
+    }
+  }
+}
+```
+
+Note that commands Zentral creates on its own — the inventory refreshes, artifact installations and other work scheduled while a device checks in — are not part of this audit trail. It records operator actions.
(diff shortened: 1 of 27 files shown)


soysauceconspiracy (approved):
LGTM

PR #1558, opened 2026-08-01
Audit blocking and unblocking an enrolled device

np5:
## Why

Blocking a device cuts it off from MDM, and neither entry point left any trace of who did it or when:

| | View | Audited before |
|---|---|---|
| UI | `BlockEnrolledDeviceView` / `UnblockEnrolledDeviceView` | no |
| API | `block/` / `unblock/` | no |

Command creation and deletion are audited (#1557), so a device's timeline recorded the commands sent to it but not the moment it stopped being managed at all.

## The part that isn't obvious

`EnrolledDevice.serialize_for_event()` did not report `blocked_at` — it returned `pk`, `udid`, `serial_number`, `platform` and `name`. A naive `AuditEvent` with `Action.UPDATED` would therefore have carried a `prev_value` and `new_value` that were **byte-identical**: an event recording that something changed, with no way to tell what, or in which direction.

So the model now reports it. This is the same call made for command `kwargs` in #1557 — extend the serialization contract so the standard audit machinery carries the information, rather than inventing a parallel event type.

Two details:

- **Isoformatted at the source**, per #1556 (`ee9032fb`). A raw datetime does not raise in production — kombu wraps it in a `{"__type__", "__value__"}` envelope that then lands in the stores. The new field is covered by `assert_serialize_for_event_is_json_native`.
- **Blast radius is three lines.** The only non-`keys_only` callers are the `MDMDownloadEvent` payloads in `events/downloads.py`, where the extra field is harmless. The existing model test only exercises `keys_only=True`.

## What changed

Both interfaces post an `updated` audit event from a **shared base class** — `UpdateEnrolledDeviceBlockView` in each of the UI and API modules — so block and unblock cannot drift apart, and a future third variant inherits the audit. Subclasses supply only `verify_block_state()` and `update_block_state()`; response bodies, status codes, querysets, templates and the block-only push notification are all unchanged.

Events are tagged with `machine_serial_number`, the argument added in #1557, so the transition lands on the device timeline next to the commands.

```json
{
  "action": "updated",
  "object": {
    "model": "mdm.enrolleddevice",
    "prev_value": {"serial_number": "C02...", "blocked_at": null},
    "new_value": {"serial_number": "C02...", "blocked_at": "2026-08-01T09:15:22"}
  }
}
```

## Notes for review

**No-ops cannot emit an event.** All four views already reject a request that would not change anything — 404 in the UI (the querysets are `.allowed()` / `.blocked()`), 400 in the API — and there is a test on each path asserting nothing is posted.

**`purge_state(full=True)` also clears `blocked_at`**, from the device check-in path in `public_views/mdm.py`. A device can therefore become unblocked without an operator. That is the engine path with nobody to attribute it to — the same call made for engine-created commands in #1557 — so it stays out. The documentation now says so explicitly, because an auditor reading only these events could otherwise conclude a device is still blocked when it is not.

**One existing assertion changed**: `test_block_enrolled_device_post` asserted exactly one `on_commit` callback; the block path now registers two (the push notification and the audit event). The test was extended rather than loosened.

**Docs section renamed** from *Command audit trail* to *Audit trail*, with the shared request-context and engine-actions paragraphs lifted to the top and `### Commands` / `### Blocking a device` underneath, since it now covers more than commands.

## Tests

2585 tests pass across `tests.mdm` and `tests.core_events`. flake8 clean.

Six new tests: audit event content on each of the four paths (transition direction, serial, request context, token-auth on the API), and no-event assertions on the rejected block/unblock cases. Plus a model test covering `blocked_at` in both states and asserting the payload stays JSON-native.

## Not verified

`mkdocs build --strict` could not be run — mkdocs is not in the container image and `pip install` fails there on venv permissions. The section was checked by hand: balanced fences, consistent heading nesting, both JSON examples parse, no links or images, and no stale references to the old section name.

### DIFF
--- a/docs/apps/mdm.md
+++ b/docs/apps/mdm.md
@@ -504,18 +504,22 @@ A Recovery Password Configuration can only be deleted if it is not linked to any
 3. Use the *Delete* button in the list view or on the configuration details page.
 4. Confirm the deletion when prompted.
 
-## Command audit trail
+## Audit trail
 
-Every command an operator creates is recorded as a `zentral_audit` event with the `created` action. This covers both entry points:
+Operator actions on a device are recorded as `zentral_audit` events. Each one carries the full request context — the authenticated user (including whether an API token was used, and which one), the source IP, the user agent, the HTTP method, the path and the resolved view name — and is tagged with the device serial number, so it also appears in the device's event timeline.
+
+Actions Zentral takes on its own — the inventory refreshes, artifact installations and other work scheduled while a device checks in — are not part of this audit trail. It records operator actions.
+
+### Commands
+
+Every command an operator creates is recorded with the `created` action. This covers both entry points:
 
 - the *New command* forms in the web interface, under a device's detail page
 - the `erase/`, `lock/` and `send_custom_command/` HTTP API endpoints
 
-Each event carries the full request context — the authenticated user (including whether an API token was used, and which one), the source IP, the user agent, the HTTP method, the path and the resolved view name — alongside the serialized command. Events are tagged with the device serial number, so they also appear in the device's event timeline, and are linked to the command UUID.
+Command events are linked to the command UUID, and carry the serialized command. Deleting a queued command is audited the same way, with the `deleted` action.
 
-Deleting a queued command is audited the same way, with the `deleted` action.
-
-### Command parameters and secrets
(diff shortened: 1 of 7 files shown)


soysauceconspiracy (approved):
LGTM

PR #1560, opened 2026-08-01
Audit the artifact and blueprint artifact changes made from the web interface

np5:
## Why

Artifacts decide which profiles, apps and declarations get installed, and linking one to a blueprint deploys it to every device using that blueprint. None of that was audited from the web interface.

What makes this drift rather than a decision is the coverage pattern:

- the **HTTP API** audits all of it, via `ListCreateAPIViewWithAudit` / `RetrieveUpdateDestroyAPIViewWithAudit` in `api_views/artifacts.py` and `blueprints.py`
- in the **same web interface module**, 12 views already audit — blueprint, FileVault config, recovery password config and software update enforcement CRUD
- the artifact views, sitting between them, did not

## What changed

**Seven views only needed their base swapped** for the audit variant — `UpdateArtifactView`, `DeleteArtifactView`, `CreateBlueprintArtifactView`, `UpdateBlueprintArtifactView`, `DeleteBlueprintArtifactView`, `UpdateArtifactVersionView`, `DeleteArtifactVersionView`.

**Three build the event themselves**, being `FormView`/`TemplateView` based rather than model views. Two are base classes, so every concrete form is covered by construction:

| View | Covers |
|---|---|
| `BaseCreateArtifactView` | declaration, cert asset, data asset, enterprise app, profile, provisioning profile creation |
| `BaseUpgradeArtifactVersionView` | all seven upgrade forms |
| `CreateAssetArtifactView` | an app created from the Apps and Books catalog |

## Notes for review

**The blueprint refresh still runs after the event is built.** Four of these views reserialize the blueprint after saving, and the delete views do it *after* the row is gone — `DeleteViewWithAudit` builds the event from `self.object` before deleting, then the override reserializes from the in-memory instance. Two tests assert the refresh still happens, so the audit can't silently break deployment.

**`AssociateLocationAssetView` is deliberately excluded.** It launches a Celery task and mutates nothing during the request, so an `updated` event would carry an identical `prev_value` and `new_value` — the same empty-event shape that `blocked_at` had to be added to `serialize_for_event()` to avoid in #1558.

**Two views were invisible to a first pass.** `UpdateArtifactView`, `DeleteArtifactView` and `UpdateBlueprintArtifactView` define no methods at all — they inherit everything from the generic base — so a scan looking for mutating methods misses them. Worth knowing if anyone audits the remaining gaps the same way.

**Still unaudited, deliberately out of scope**: the enrollment views (OTA, user, DEP), the push certificate and DEP token views in `views/setup.py`, `ChangeEnrolledDeviceBlueprintView`, `AssignDEPDeviceProfileView`, and the realm group tag mapping views — whose model has no `serialize_for_event()` yet, so it needs that first.

## Tests

2595 tests pass across `tests.mdm` and `tests.core_events`. flake8 clean.

New `tests/mdm/test_management_artifact_audit.py`, 10 tests — one per audited path, asserting a single event with the right action, model and pk, plus the prev/new transition where it is meaningful. The existing per-view test cases already cover the behaviour of these views in detail, so these only check the event; extending ten long existing tests would have destabilised them for no extra signal.

## Not verified

`mkdocs build --strict` could not be run — mkdocs is not in the container image and `pip install` fails there on venv permissions. Checked by hand: balanced fences, consistent heading nesting, both JSON examples parse, no links or images.

### DIFF
--- a/docs/apps/mdm.md
+++ b/docs/apps/mdm.md
@@ -506,7 +506,7 @@ A Recovery Password Configuration can only be deleted if it is not linked to any
 
 ## Audit trail
 
-Operator actions on a device are recorded as `zentral_audit` events. Each one carries the full request context — the authenticated user (including whether an API token was used, and which one), the source IP, the user agent, the HTTP method, the path and the resolved view name — and is tagged with the device serial number, so it also appears in the device's event timeline.
+Operator actions are recorded as `zentral_audit` events. Each one carries the full request context — the authenticated user (including whether an API token was used, and which one), the source IP, the user agent, the HTTP method, the path and the resolved view name. Actions that target a single device are also tagged with its serial number, so they appear in that device's event timeline.
 
 Actions Zentral takes on its own — the inventory refreshes, artifact installations and other work scheduled while a device checks in — are not part of this audit trail. It records operator actions.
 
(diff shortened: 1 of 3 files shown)


soysauceconspiracy (approved):
LGTM

PR #1561, opened 2026-08-01
Audit the enrollment changes made from the web interface

np5:
## Why

An enrollment is the path onto the MDM: creating one opens a way in, revoking one closes it. The HTTP API audits DEP and OTA enrollment writes through the DRF mixins, but none of the web interface views did — and `EnrollmentSecret.revoke()` still carried a literal `# TODO events` comment.

Eight views, now audited:

| Type | Views |
|---|---|
| Automated Device Enrollment | create, update |
| OTA | create, update, revoke |
| User | create, update, revoke |

**User enrollments were not in the request** ("DEP / OTA"), but the three types are siblings with near-identical views — auditing two and leaving the third would recreate exactly the drift this closes. Say the word if you'd rather split it out.

## A helper, on the eighth copy

These views all drive two forms (an enrollment form plus an enrollment secret form), so they implement `post()` themselves and can't use the mixins. That made this the eighth hand-rolled copy of the same `transaction.on_commit` block across four PRs, so it's now `post_audit_event()` in `zentral/utils/views.py`, next to the mixins it complements. Existing hand-rolled sites are left alone; they can adopt it incrementally.

## Three things the helper cannot paper over

All three are covered by tests, and each would have produced a quietly wrong audit trail:

**A `ModelForm` writes the posted data onto its instance while validating, not in `save()`.** So `prev_value` has to be serialized at the very top of `post()`, before the forms are constructed. Serialize it later and the event reports the *new* name as the old one. `UpdateViewWithAudit` sidesteps this by re-reading from the DB; the hand-written posts can't.

**`revoke()` is a no-op when already revoked**, so the revoke views only post when the enrollment wasn't revoked yet — otherwise `prev_value == new_value` and the event records nothing. Same shape as the block/unblock case in #1558.

**`define_dep_profile()` is what persists a DEP enrollment, and it talks to Apple.** The `created` event is posted from the `else` branch of the `DEPClientError` handler, so a failed call to Apple can't leave a `created` event behind for an enrollment that doesn't exist. There's a test for exactly that.

## The enrollment secret never reaches the events

Worth stating explicitly since these objects carry a live credential: `EnrollmentSecret.serialize_for_event()` reports the quota, request count, expiry, revocation state, business unit, tags and serial/UDID restrictions — but **not** `secret`. A test asserts the secret string appears nowhere in the payload, so a regression there can't silently put a working enrollment credential into every event store.

It does report `is_revoked` and `revoked_at`, which is why revocation produces a meaningful transition with no model change needed:

```json
{"action": "updated",
 "object": {"model": "mdm.otaenrollment",
            "prev_value": {"enrollment_secret": {"is_revoked": false}},
            "new_value": {"enrollment_secret": {"is_revoked": true, "revoked_at": "…"}}}}
```

## Tests

3053 tests pass across `tests.mdm`, `tests.core_events` and `tests.server_accounts`. flake8 clean.

New `tests/mdm/test_enrollment_audit.py`, 10 tests: the event on each of the eight paths, plus no-event assertions on a repeat revoke and on a failed DEP call.

**Four existing assertions changed** in `test_setup_dep_enrollment.py`. They counted `on_commit` callbacks to check whether the DEP profile push was queued; the audit event adds one. Notably `test_update_dep_enrollment_post_update_admin_update_pwd` asserted *zero* callbacks ("the DEP profile has not changed") and is now 1 — correct, since the enrollment was still updated even though the DEP profile wasn't. Extended rather than loosened.

## Deliberately out of scope

Still unaudited, and each wanting its own decision: the DEP **virtual server** and **token** views in `views/setup.py` (`ConnectDEPVirtualServerView`, `RenewDEPTokenView`, `UpdateDEPVirtualServerView`) — these handle an uploaded encrypted token, so they need a careful look at what `serialize_for_event` would expose, and `DEPToken` has no serializer yet. Also the DEP **device** views (`AssignDEPDeviceProfileView`, `RefreshDEPDeviceView`, `DisownDEPDevice`), which are machine-scoped and want `machine_serial_number` plus a hook on the mixins. Push certificates and locations are neither DEP nor OTA.

## Not verified

`mkdocs build --strict` could not be run — mkdocs is not in the container image and `pip install` fails there on venv permissions. Checked by hand: balanced fences, consistent heading nesting, all three JSON examples parse, no links or images.

### DIFF
--- a/docs/apps/mdm.md
+++ b/docs/apps/mdm.md
@@ -575,6 +575,26 @@ Artifacts decide which profiles, apps and declarations get installed, and linkin
 
 Blueprints themselves, along with the FileVault, recovery password and software update enforcement configurations, are audited the same way.
 
+### Enrollments
+
+An enrollment is the path onto the MDM, so creating one, changing it and revoking it are all audited, for the Automated Device Enrollment, OTA and user enrollment types alike, from the web interface as well as the HTTP API. A revocation is reported with the `updated` action, and shows up as the `is_revoked` and `revoked_at` fields of the enrollment secret changing:
+
+```json
+{
+  "action": "updated",
+  "object": {
+    "model": "mdm.otaenrollment",
+    "prev_value": {"name": "…", "enrollment_secret": {"is_revoked": false}},
+    "new_value": {"name": "…", "enrollment_secret": {"is_revoked": true,
+                                                     "revoked_at": "2026-08-01T09:15:22"}}
+  }
+}
+```
+
+The enrollment secret itself is never included — only its metadata, such as its quota, request count, expiry, revocation and the business unit, tags and serial numbers it is restricted to. Revoking an already revoked enrollment changes nothing and records nothing.
+
+For Automated Device Enrollment, the enrollment only exists once Apple has accepted the profile, so a failed call to Apple records nothing either.
+
 ## HTTP API
 
 ### `/api/mdm/dep/devices/`
--- a/tests/mdm/test_enrollment_audit.py
+++ b/tests/mdm/test_enrollment_audit.py
@@ -1,0 +1,255 @@
+import json
+import uuid
+from unittest.mock import Mock, patch
+
+from django.contrib.auth.models import Group
+from django.test import TestCase
+from django.urls import reverse
+from django.utils.crypto import get_random_string
+
+from accounts.models import User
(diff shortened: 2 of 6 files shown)


soysauceconspiracy (approved):
LGTM

PR #1562, opened 2026-08-02
Let the audit views tag their events with a machine serial number

np5:
## Why

`AuditEvent.build_from_request_and_instance()` takes a `machine_serial_number`, so an event about one device also lands on that device's timeline. But the audit view mixins called it with fixed arguments, so only the hand-rolled sites could pass it — and the views that audit a machine-scoped object could not. That was the gap behind "the argument is only usable from hand-rolled sites" noted when it was added in #1557.

## The hook

All five audit views — `CreateViewWithAudit`, `UpdateViewWithAudit`, `DeleteViewWithAudit` and the two DRF ones — now call `get_audit_machine_serial_number()`, which defaults to `None`. Every existing view keeps its payload *and* its metadata byte-identical; that's what the 4150 passing tests confirm.

The default lives in a new `AuditViewMixin` shared by the three Django views. The DRF pair keep their own copy: their `on_commit_callback_extra()` takes the instance, and the two hierarchies have nothing else in common, so folding them into one mixin would have meant reconciling that signature for no benefit.

## The hook ships with a caller

`ChangeEnrolledDeviceBlueprintView` is the first view to override it, so this isn't an unused extension point. **This is slightly more than was asked for** — happy to split it out if you'd rather, though a hook with no caller is untestable in any meaningful way.

It needed one thing first: **`blueprint` added to `EnrolledDevice.serialize_for_event()`**. That view changes nothing else, so without it `prev_value` and `new_value` would have matched and the event would have recorded that *something* changed without saying what — the same trap `blocked_at` had to be added to avoid in #1558. Worth noting the pattern: every machine-scoped view needs a field added to the serializer first, and `AssignDEPDeviceProfileView` will need `DEPDevice.serialize_for_event()` written from scratch.

## Realm group tag mappings

They decide the tags a device inherits from its user's groups, and their three views were plain model views blocked on a model with no serializer. `RealmGroupTagMapping.serialize_for_event()` added — both FKs already had `keys_only` serializers and there are no secrets — and the views swapped to the audit variants.

`CreateView` and `DeleteView` are now entirely unused in `management.py`, so they came out of the import.

## Notes for review

**Their tests look up the audit event rather than taking `call_args_list[0]`.** `update_realm_tags()` posts its own `machine_tag` events from the same request, so neither the callback count nor the audit event's position is fixed — the update path emits three callbacks. An index-based assertion would have passed for the wrong reason, so there's a `get_audit_event()` helper that filters by type and asserts exactly one.

**Two existing assertions changed.** `test_change_enrolled_device_blueprint_post` counted one on-commit callback (the push notification) and is now 2; the `EnrolledDevice` model test asserted the exact serialized dict and gains `blueprint: None`. Both extended rather than loosened.

## Tests

4150 tests pass across `tests.mdm`, `tests.utils`, `tests.core_events`, `tests.core_probes`, `tests.core_stores`, `tests.server_accounts` and `tests.inventory` — deliberately wide, since the mixins are used by every module. flake8 clean.

New `tests/utils/test_audit_views.py` asserts the hook exists and defaults to `None` on all five views, that the Django three share the mixin, and that an override is picked up. Plus audit-event coverage on the three mapping views, the blueprint change (including `machine_serial_number`), and `blueprint` in the model serializer with the JSON-native assertion from #1556.

## Not verified

`mkdocs build --strict` could not be run — mkdocs is not in the container image and `pip install` fails there on venv permissions. Checked by hand: balanced fences, consistent heading nesting, all three JSON examples parse, no links or images.

### DIFF
--- a/docs/apps/mdm.md
+++ b/docs/apps/mdm.md
@@ -545,6 +545,10 @@ Commands that carry no parameters have no parameters entry at all.
 }
 ```
 
+### Changing a device's blueprint
+
+Assigning a device to another blueprint, or removing it from one, is recorded with the `updated` action, with the previous and the new blueprint in the event.
+
 ### Blocking a device
 
 Blocking and unblocking a device are recorded with the `updated` action, from the *Block*/*Unblock* buttons on the device detail page as well as from the `block/` and `unblock/` HTTP API endpoints. The transition is visible in the `blocked_at` field, which is `null` when the device is not blocked:
@@ -573,7 +577,7 @@ Artifacts decide which profiles, apps and declarations get installed, and linkin
 - creating a new **artifact version**, through the upgrade forms, and updating or deleting an existing one
 - linking an artifact to a **blueprint**, changing how it is scoped there, and unlinking it
 
-Blueprints themselves, along with the FileVault, recovery password and software update enforcement configurations, are audited the same way.
+Blueprints themselves, along with the FileVault, recovery password and software update enforcement configurations, are audited the same way, as are the realm group tag mappings, which decide the tags a device inherits from its user's groups.
 
 ### Enrollments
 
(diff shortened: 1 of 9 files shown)


soysauceconspiracy (approved):
LGTM

PR #1565, opened 2026-08-02
Attribute the APNs notifications an operator asks for

np5:
## Why

Poking a device or a user changes no state, so it has no audit event — the `mdm_device_notification` event it produces is the only record that it happened. And that event was built with `EventMetadata(machine_serial_number=serial_number)` alone: it said a device was poked, and never who poked it.

## What changed

`build_mdm_device_notification_event()` takes an optional `request` and builds an `EventRequest` from it. The parameter is threaded down through `send_enrolled_device_notification()` / `send_enrolled_user_notification()` / `_send_target_notification()`.

**The request is passed from every operator entry point, not only the two poke views** — blocking a device from the web interface or the API, and changing a device's blueprint, all notify the device too. Those already have an audit event carrying the attribution, so the notification event is not the sole record there, but an event that names the request which caused the push is strictly more useful than one that doesn't, and the alternative is an inconsistency with no principle behind it.

**The engine callers deliberately pass nothing** — artifact reconciliation ([artifacts.py:1048](zentral/contrib/mdm/artifacts.py:1048)), the Apps and Books preprocessor, and the `send_device_notification` management command. So the presence of a request is what distinguishes a poke an operator asked for from one Zentral decided on by itself.

## Notes for review

**The two poke views had no tests at all** — `grep poke_enrolled_device tests/` came back empty. They now have the usual redirect and permission checks plus an attribution test each.

Those attribution tests drive the **real APNs code path** with only `httpx.Client.post` mocked, following `test_apns.py`, so the event is genuinely built and posted rather than asserted against a mocked helper. They need a full-size push certificate key: `force_push_certificate(with_material=True)` defaults to `reduced_key_size=True`, and a 512-bit key makes `create_client_ssl_context` raise `EE_KEY_TOO_SMALL`. That's built once in `setUpTestData` with a comment, since it's slow.

**Three existing assertions changed.** `send_enrolled_device_notification.assert_called_once_with(enrolled_device)` no longer matches now that the call carries `request=`. Rather than just adding the argument — the request object isn't conveniently available in those tests, since `response.wsgi_request` after `follow=True` is the redirect target, not the POST — they now assert the positional argument and that the passed request's method is `POST`, which is the property that actually matters.

The five assertions in `test_artifacts.py` are untouched: those cover the engine path, which still calls without a request.

## Tests

2611 tests pass across `tests.mdm` and `tests.core_events`. flake8 clean.

## Not verified

`mkdocs build --strict` could not be run — mkdocs is not in the container image and `pip install` fails there on venv permissions. Checked by hand: balanced fences, consistent heading nesting, JSON examples parse, no links or images.

### DIFF
--- a/docs/apps/mdm.md
+++ b/docs/apps/mdm.md
@@ -549,6 +549,12 @@ Commands that carry no parameters have no parameters entry at all.
 
 Assigning a device to another blueprint, or removing it from one, is recorded with the `updated` action, with the previous and the new blueprint in the event.
 
+### Poking a device
+
+Poking a device or a user sends it an APNs notification so that it checks in. It changes nothing, so there is no audit event for it — but the `mdm_device_notification` event it produces carries the request context of the operator who asked, alongside the device serial number and whether Apple accepted the notification.
+
+The notifications Zentral sends on its own, while reconciling artifacts or handling an Apps and Books notification, carry no request, which is how an operator poke can be told apart from one Zentral decided on.
+
 ### Blocking a device
 
 Blocking and unblocking a device are recorded with the `updated` action, from the *Block*/*Unblock* buttons on the device detail page as well as from the `block/` and `unblock/` HTTP API endpoints. The transition is visible in the `blocked_at` field, which is `null` when the device is not blocked:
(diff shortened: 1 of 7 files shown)


soysauceconspiracy (approved):
LGTM

PR #1572, opened 2026-08-04
mdm: back off exponentially when polling for app installs

np5:
## Why

`ManagedApplicationList` reschedules itself while an app install is pending, but used a **fixed 15s delay with a cap of 10 retries** — so it stopped observing the device after ~2.5 minutes. Any install taking longer than that was never seen, and the `TargetArtifact` stayed at `AwaitingConfirmation` indefinitely even though the app had installed.

Large App Store apps routinely need far more than 2.5 minutes, and a device can stay unreachable for days, so the window was too short by construction. Two things make the consequence worse than a cosmetic delay:

- **It never self-corrects.** Once the command stops rescheduling, nothing re-examines that artifact, so the wrong status is permanent until something else triggers a new install.
- **The status reads as pending, not failed**, so the artifact looks like it is still progressing rather than surfacing as something to investigate.

The result is that a successful install of a large app is indistinguishable from one that never happened.

## What changed — commit 1, store apps (`ManagedApplicationList`)

```python
FIRST_RETRY_DELAY_SECONDS = 20
RETRY_DELAY_GROWTH_FACTOR = 2
MAX_RETRIES = 15


def get_retry_delay_seconds(retries):
    return FIRST_RETRY_DELAY_SECONDS * RETRY_DELAY_GROWTH_FACTOR ** retries
```

Doubling with **no ceiling**, so the schedule spans about a week: **16 checks over ~7.6 days**, versus 11 checks over 2.5 minutes.

| # | delay | at | | # | delay | at |
|---|---|---|---|---|---|---|
| 1 | — | 15s | | 9 | +42m40s | 1h25m |
| 2 | +20s | 35s | | 10 | +1h25m | 2h50m |
| 3 | +40s | 1m15s | | 11 | +2h50m | 5h41m |
| 4 | +1m20s | 2m35s | | 12 | +5h41m | 11h22m |
| 5 | +2m40s | 5m15s | | 13 | +11h22m | 22h45m |
| 6 | +5m20s | 10m35s | | 14 | +22h45m | 1.90d |
| 7 | +10m40s | 21m15s | | 15 | +1.90d | 3.79d |
| 8 | +21m20s | 42m35s | | 16 | +3.79d | **7.59d** |

Shape of the curve, and why these constants:

- **8 checks inside the first hour**, 4 inside the first 3 minutes, so a quick install is still confirmed about as fast as before.
- **No ceiling**, which keeps the schedule a single closed-form expression with no clamp. The consequence is that the last gap — half the window, with doubling — is what bounds how stale a status can get: 3.8d here.
- **The extra retries are close to free.** The chain stops as soon as the device reports a terminal status, so a normal install ends after a handful of checks; only one that never resolves walks the full schedule.
- Delays are exact integers (`20 · 2ⁿ`), so no rounding is involved.
- Both the delay and the retry cap replace inline `# TODO hardcoded` markers.

A gentler factor is available if the tail gap ever matters more than the command count: ×1.5 over 24 retries spans the same ~7.8d in 25 checks and holds the last gap to 2.6d (a third of the window) with 12 checks in the first hour. Doubling was chosen as the simpler schedule, since the extra sampling only ever applies to artifacts that never resolve.

## What changed — commit 2, enterprise apps (`InstalledApplicationList`)

Same defect: it grew the delay linearly and gave up after 10 retries, ~14 minutes.

```python
FIRST_RETRY_DELAY_SECONDS = 15
RETRY_DELAY_GROWTH_FACTOR = 1.35
MAX_RETRIES = 32
```

**Nothing is shared between the two commands, on purpose.** Enterprise apps are locally hosted packages rather than App Store downloads, so there is no reason for the two schedules to move together, and only the install-check path retries here at all — this command also collects the app inventory, which never reschedules. Beyond the retry block the two have little in common: one reads a single `Status` string per identifier, the other infers state from five download/install booleans across a key-matched list. A shared mixin would own about six lines and add more indirection than it removes.

The factor is 1.35 rather than 2 so the curve stays dense exactly where the linear one was. **Every check the old schedule made still happens at least as early:**

| check | old (linear, stops at 11) | new (×1.35) |
|---|---|---|
| 2–3 | 30s · 1m | 30s · 50s |
| 4–6 | 1m45s · 2m45s · 4m | 1m17s · 1m54s · 2m44s |
| 7–9 | 5m30s · 7m15s · 9m15s | 3m51s · 5m22s · 7m25s |
| 10–11 | 11m30s · **14m — gives up** | 10m10s · 13m53s |
| … | | continues to **7.35d** |

33 checks over 7.35 days, largest gap 1.9d (26% of the window). A test pins the early-cadence property directly, so a future tweak cannot quietly make the first quarter hour sparser than it used to be.

## Notes for the reviewer

- `first_delay_seconds = 15` in `install_application.py` is the delay before the first check and is intentionally untouched, which is why check #1 lands at 15s rather than 20s.
- Batching identifiers across artifacts is **not** a safe way to cut command volume: the command binds a single `artifact_version`, while `_update_device_artifact` loops `self.identifiers` and overwrites `ta_status` last-wins, so a batch would write one app's status onto another artifact and could also stop polling early for an app still installing. Making that safe needs an identifier→artifact_version map, which is out of scope.
- **Existing stuck artifacts are not migrated.** Re-polling them resolves the state, since where the app did install the first response reports `Managed`.

## Tests

Updated the one test pinning the old cap (`retries: 10` → `MAX_RETRIES`) and added five covering the schedule: the doubling sequence, the first delay, that it is genuinely uncapped (monotonic, last delay largest), a guard on the total window (7–8.5 days), the staleness bound (last gap under 4 days), and one asserting the rescheduled command's `not_before` reflects the backoff delay.

Both commands get the same coverage: growth sequence, first delay, genuinely uncapped, total window, and the rescheduled command's `not_before`. `InstalledApplicationList` adds the early-cadence guard described above.

```
tests.mdm.test_managed_application_list_command
tests.mdm.test_installed_application_list_command
tests.mdm.test_install_application_command
tests.mdm.test_install_enterprise_application_command .... 59 tests OK
```

### DIFF
--- a/tests/mdm/test_installed_application_list_command.py
+++ b/tests/mdm/test_installed_application_list_command.py
@@ -4,11 +4,17 @@ from datetime import datetime
 from unittest.mock import call, patch
 
 from django.test import TestCase
+from django.utils import timezone
 from django.utils.crypto import get_random_string
 
 from zentral.contrib.inventory.models import MetaBusinessUnit, MetaMachine
 from zentral.contrib.mdm.artifacts import Target
 from zentral.contrib.mdm.commands import InstalledApplicationList
+from zentral.contrib.mdm.commands.installed_application_list import (
+    FIRST_RETRY_DELAY_SECONDS,
+    MAX_RETRIES,
+    get_retry_delay_seconds,
+)
 from zentral.contrib.mdm.commands.scheduling import _update_extra_inventory, load_command
 from zentral.contrib.mdm.models import (
     Artifact,
@@ -309,7 +315,7 @@ class InstalledApplicationListCommandTestCase(TestCase):
             target,
             artifact_version,
             kwargs={"apps_to_check": [{"Identifier": "yolo.fomo", "ShortVersion": "1.0"}],
-                    "retries": 10},
+                    "retries": MAX_RETRIES},
             queue=False,
         )
         cmd.process_response(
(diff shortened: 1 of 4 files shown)


soysauceconspiracy (approved):
LGTM

PR #1575, opened 2026-08-05
mdm: rotate the recovery password and the FileVault PRK after a reveal

np5:
DEP enrollments can already schedule a rotation of the auto admin password after it has been read. This adds the same mechanism to the recovery password configurations and to the FileVault configurations, and makes queued auto rotations safe to deliver late.

### `mdm: verify the target again before delivering a queued command`

A queued command is only verified when it is created, but the channel and device requirements of some commands cover state which changes in between: `SetFirmwarePassword` cannot be sent while a firmware password change is pending, `RotateFileVaultKey` needs an escrowed PRK, `SetAutoAdminPassword` needs the auto admin account. A command queued with a delay has time to go stale.

New opt-in `Command.verify_target_before_delivery`, checked in `_get_next_queued_command`, enabled on those commands plus `SetRecoveryLock`. An unverified command **stays queued** instead of being dropped: the device becoming a valid target again is the common case, and the rotation deduplication then treats it as what it is, a rotation which is still pending.

### `mdm: add a command method to check for an already queued command`

The admin password auto rotation skips itself when a rotation is already queued for the device. That query is needed by every mechanism below, so it moves to `Command.is_queued_for_target` on the base class, where the target already knows which model and lookup its commands use — so it holds for the user channel too, unlike the hardcoded `DeviceCommand` query it replaces. Four call sites now share it.

### `mdm: rotate the recovery password after it has been revealed`

New `RecoveryPasswordConfig.reveal_rotation_delay`, in minutes: 60 by default, 0 to disable, at least 5 min so that a rotation cannot land before the operator had a chance to use the password, 1440 min (1 day) max.

Static passwords are never rotated, and, as for the interval based rotation, a firmware password is only rotated when `rotate_firmware_password` is set, since applying it requires a reboot. Multiple reveals do not queue up multiple rotations.

Note the relaxed rule: `rotate_firmware_password` used to require a rotation interval, it now accepts an interval **or** a reveal delay — otherwise reveal-only firmware rotation would be unconfigurable.

### `mdm: rotate the FileVault PRK after it has been revealed`

New `FileVaultConfig.prk_reveal_rotation_delay`, same semantics. It is deliberately left out of the configuration fingerprint (`FileVaultConfig.uuid`): it is only used server side, so changing it must not make every device receive the FileVault configuration again.

### `mdm: ask for the security info when a rotation does not return the key`

`RotateFileVaultKey` can return the new PRK encrypted with the reply certificate, but `RotateResult` and `EncryptedNewRecoveryKey` are [both optional](https://github.com/apple/device-management/blob/67045e2fa06f528b196c01edee6a8bf88b844beb/mdm/commands/rotate.file.vault.key.yaml) in Apple's schema, and the CMS payload can fail to decrypt. In those cases the new key was only collected at the next inventory update, leaving Zentral with a stale PRK — and a stale `filevault_prk_updated_at`, which is what the interval rotation keys off — for as long as the inventory interval.

The fallback is the escrowed key: [`FDE_PersonalRecoveryKeyCMS`](https://github.com/apple/device-management/blob/67045e2fa06f528b196c01edee6a8bf88b844beb/mdm/commands/information.security.yaml#L202) in the security info response carries it. That is what `SetupFileVault` already did for the first escrow, so both go through `SecurityInfo.create_for_prk_escrow` now, which shares the delay and does not queue a second command when one is already pending. A rotation whose response did give us a usable key schedules nothing.

### Upgrade notes

Both new fields default to 60 for new configurations, but the migrations set them to 0 on existing ones, so nothing starts rotating on upgrade.

### Tests

`tests.mdm` green, 2604 tests. `makemigrations --check` clean.


soysauceconspiracy (approved):
LGTM

PR #1576, opened 2026-08-05
mdm: show an error page when DEP web enroll lacks the device info header

sebastian-zentral:
The Setup Assistant web view occasionally issues the initial navigation to the DEP web enrollment URL without the `x-apple-aspen-deviceinfo` header.

Instead of letting the request fail with a `SuspiciousOperation`, the view renders `mdm/dep_missing_device_info.html`, a short template extending `base_error.html`, with a title naming the fault so a screenshot is enough for the support team to identify it. The response is still a `400`, and the event is still posted as a `failure`.

### DIFF
--- a/zentral/contrib/mdm/public_views/dep.py
+++ b/zentral/contrib/mdm/public_views/dep.py
@@ -1,21 +1,23 @@
 import base64
 import logging
 import plistlib
+
 from django.http import HttpResponse, HttpResponseRedirect, JsonResponse
-from django.shortcuts import get_object_or_404
+from django.shortcuts import get_object_or_404, render
 from django.template import Context, Engine
 from django.urls import reverse
 from django.views.generic import View
+
 from zentral.contrib.inventory.exceptions import EnrollmentSecretVerificationFailed
 from zentral.contrib.inventory.utils import verify_enrollment_secret
 from zentral.contrib.mdm.crypto import verify_iphone_ca_signed_payload
 from zentral.contrib.mdm.events import DEPEnrollmentRequestEvent
-from zentral.contrib.mdm.models import DEPEnrollmentSession, DEPEnrollment, EnrolledDevice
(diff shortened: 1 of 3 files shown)


zentral/contrib/mdm/public_views/dep.py
```diff
@@ -214,16 +216,22 @@ def dep_web_enroll_callback(request, realm_authentication_session, dep_enrollmen
...
-            return base64.b64decode(self.request.META["HTTP_X_APPLE_ASPEN_DEVICEINFO"])
-        except KeyError:
-            self.abort("Missing x-apple-aspen-deviceinfo header")
+        return base64.b64decode(self.request.META["HTTP_X_APPLE_ASPEN_DEVICEINFO"])
 
     def get(self, request, *args, **kwargs):
+        if "HTTP_X_APPLE_ASPEN_DEVICEINFO" not in request.META:
+            reason = "Missing x-apple-aspen-deviceinfo header"
+            logger.warning(reason)
+            self.post_event("warning", reason=reason)
```
np5: error

np5: not warning

zentral/contrib/mdm/templates/mdm/dep_web_enroll_error.html
```diff
@@ -0,0 +1,65 @@
...
+      body { color: #f5f5f7; background: #000000; }
+      .card { background: #1c1c1e; box-shadow: none; }
+      p { color: #c7c7cc; }
+      .code-label { color: #98989d; }
+      .code { color: #f5f5f7; background: #2c2c2e; border-color: #3a3a3c; }
+    }
+  </style>
+</head>
+<body>
+  <main class="card">
```
np5: Really no need for a code here, just a simple sentence. This way we can get rid of the variable in the views. Inherit from the error page as well maybe?

np5 (dismissed):

soysauceconspiracy (approved):
LGTM

PR #1577, opened 2026-08-06
turbo: fix what a schedule delivers at creation

np5:
## The bug

The one-time job done-gate in `public_views/config.py` is keyed on the `OneTimeJob`, not on the `Job` it points at:

```python
done = set(
    OneTimeJobMachine.objects
    .filter(serial_number=serial_number, one_time_job__in=open_one_time_jobs,
            last_result_at__isnull=False)
    .values_list("one_time_job_id", flat=True)
)
```

Note what is absent: any reference to which job produced the result, or to a version. It is a latch on the `(machine, schedule)` pair, and nothing reopens it.

`job` was editable on an existing schedule, from the API (`JobScopeSerializerMixin.validate` pinned only `configuration`) and from the UI (`job` is in the form fields, and `UpdateOneTimeJobView` uses that form).

So re-pointing a one-time job from script A to script B left every machine that had already run A latched, and **B never ran there**. It only reached the machines that had not finished A yet. No error, no event — the fleet silently split into "got it" and "never will". The `OneTimeJobMachine.last_result_version` rows also kept pointing at A's version numbering.

## The fix

`configuration` and `job` are now both pinned on update, in `JobScopeSerializerMixin.validate` and in `BaseJobScopeForm`, so the rule is stated once and applies to both schedule types.

The form does two things, and both are needed: the field is `disabled`, so it renders read-only and keeps its value; and `clean_job` rejects a posted job that isn't the current one. Disabled alone is a silent no-op — the browser drops the field, Django ignores whatever arrives, and a client that posted a different job would have seen a successful update that changed nothing.

To deliver something else, schedule it.

## Recurring jobs

Pinned too, deliberately, though the case is weaker. `RecurringJobMachine` gates nothing — recurring jobs are always served, the agent picks up the new job on the next check-in and the rows converge — so a swap there was a bookkeeping wart rather than a delivery bug. It still left one ledger mixing two jobs' version numbering, and two schedule types disagreeing about what is editable is its own cost.

One consequence: `RecurringJobForm.clean()` only checks the `(configuration, job)` unique constraint on creation now. A schedule whose job cannot change can only ever collide with itself, so the update branch was unreachable, and `test_update_recurring_job_duplicate_job_error` was testing a path that no longer exists — it is now `test_update_recurring_job_job_immutable`.

This also settles the Terraform provider, which marks `job_id` as `RequiresReplace` on both job resources ([#255](https://github.com/zentralopensource/terraform-provider-zentral/pull/255)) — now provably right rather than merely conservative.

## Tests

| test | without the fix |
|---|---|
| `test_update_one_time_job_job_immutable` (API) | 200 instead of 400 |
| `test_update_one_time_job_job_immutable` (setup) | silent 302, job swapped |
| `test_update_one_time_job_without_the_disabled_job` (setup) | — guards the real browser flow |
| `test_update_recurring_job_job_immutable` (API) | 200 instead of 400 |
| `test_update_recurring_job_job_immutable` (setup) | the old duplicate-job error |
| `test_update_recurring_job_without_the_disabled_job` (setup) | form invalid, job field required |

Each was run against the code without the change to confirm it fails for the right reason. Full `tests.turbo` suite: 404 passing.

### DIFF
--- a/tests/turbo/test_api_one_time_jobs.py
+++ b/tests/turbo/test_api_one_time_jobs.py
@@ -127,6 +127,16 @@ class TurboOneTimeJobAPITestCase(TurboAPITestCase):
         self.assertEqual(response.status_code, 400)
         self.assertEqual(response.json(), {"configuration": ["This field cannot be changed"]})
 
+    def test_update_one_time_job_job_immutable(self):
+        one_time_job = force_one_time_job()
+        other_script = force_script()
+        self.set_permissions("turbo.change_onetimejob")
+        response = self.put(reverse("turbo_api:one_time_job", args=(one_time_job.pk,)),
+                            {"configuration": str(one_time_job.configuration.pk),
+                             "job": str(other_script.job.pk)})
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.json(), {"job": ["This field cannot be changed"]})
+
     def test_delete_one_time_job(self):
         one_time_job = force_one_time_job()
         pk = one_time_job.pk
--- a/tests/turbo/test_api_recurring_jobs.py
+++ b/tests/turbo/test_api_recurring_jobs.py
@@ -145,6 +145,19 @@ class TurboRecurringJobAPITestCase(TurboAPITestCase):
         recurring_job.refresh_from_db()
         self.assertEqual(recurring_job.interval, 3600)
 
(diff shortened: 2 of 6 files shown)


sebastian-zentral (dismissed):

soysauceconspiracy (approved):
LGTM

PR #1582, opened 2026-08-07
Wait for the tasks and report their outcome in the bundle

np5:
Follow-up to 7fad830b, which moved the monolith repository sync to a celery task.

## Problem

Eleven pages launch a celery task and poll the task result, and each one carries its own copy of the same `waitForTask`/`launchTask` pair, in an inline script. None of them reports anything: the monolith repository sync posts to the API and reloads the page when the task is done, so a sync that fails, or that is skipped because the repository is already being synced, looks exactly like a sync that did nothing. The exports are worse, a failed export does nothing at all.

## Changes

The task waiting code moves to `server/static_src/js/main.js`, the bundle already loaded on every page, and is driven by the markup instead of being copied:

* a link with the `task` class posts to its `href` and polls `task_result_url`;
* `data-task-label` builds the messages: *`<label>` started.* while it runs, replaced by *done.*, *failed.*, or *skipped, already running.* when the task result is ready;
* a task result with a file sends the browser to the `download_url`, without one the page is reloaded to display what the task changed, a second after the success message;
* an export link keeps its format in `data-format`, and its enclosing form is serialized and posted with it, as the inventory export pages do today.

`base.html` wraps the django messages in a `#messages` container, so the same alerts can be added to the page from javascript.

The monolith repository page is the first one to use it: its inline script is gone, replaced by `data-task-label=\"Repository sync\"` on the link.

## Notes for the reviewer

* The other ten pages still wait for their tasks themselves. They bind the `task` class too, and both handlers would fire, launching the task twice, so the shared code only binds the links carrying a label. The `[data-task-label]` part of the selector can go away once every page is converted.
* The sync button is disabled with the bootstrap `disabled` class instead of the `disabled` property, which has no effect on a link. A second click used to start a second task that could only be skipped.
* `SKIPPED` is only returned by the monolith sync task today. Reading it in the shared code makes it the convention for the tasks that decline to run.
* `tests.monolith` passes (559 tests). The bundle was built and exercised in a browser against stubbed responses, for the started, done, skipped, failed, and download outcomes, and the export POST body was checked to be the one the inventory pages send today.

### DIFF
--- a/server/static_src/js/main.js
+++ b/server/static_src/js/main.js
@@ -78,3 +78,106 @@ $(document).ready(function () {
         }
     }
 });
+
+
+// tasks
+//
+// A link with the task class posts to its href to launch a celery task, then polls
+// the task result until it is ready. The messages are built with the task label. A
+// task result with a file sends the browser to the download URL, without one the
+// page is reloaded to display what the task changed. An export link carries the
+// format in data-format, and its enclosing form is posted with it.
+
+var TASK_POLLING_DELAY = 1000;
+var TASK_RELOAD_DELAY = 1000;
+
+function addMessage(message, tag) {
+    var alert = document.createElement("div");
+    alert.className = "alert alert-" + (tag === "error" ? "danger" : tag) + " alert-dismissible fade show";
+    alert.setAttribute("role", "alert");
+    alert.textContent = message;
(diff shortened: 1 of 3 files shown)


np5:
Closed by the branch rename, continued in #1583.

PR #1583, opened 2026-08-07
Wait for the celery tasks in the bundle, and report their outcome

np5:
Follow-up to 7fad830b, which moved the monolith repository sync to a celery task.

Supersedes #1582, closed by the branch rename.

## Problem

Eleven pages launch a celery task and poll the task result, and each one carries its own copy of the same `waitForTask`/`launchTask` pair, in an inline script. None of them reports anything. The monolith repository sync posts to the API and reloads the page when the task is done, so a sync that fails, or that is skipped because the repository is already being synced, looks exactly like a sync that did nothing. The exports are worse: a failed export leaves the dropdown as if it had never been clicked.

## Changes

**1. The task waiting code moves to `server/static_src/js/main.js`**, the bundle already loaded on every page, and is driven by the markup instead of being copied:

* a link with the `task` class posts to its `href` and polls `task_result_url`;
* `data-task-label` builds the messages: *`<label>` started.* while it runs, replaced by *done.*, *failed.*, or *skipped, already running.* when the task result is ready;
* a task result with a file sends the browser to the `download_url`, without one the page is reloaded to display what the task changed, a second after the success message;
* an export link keeps its format in `data-format`, and its enclosing form is serialized and posted with it.

It is plain javascript, no jquery: `fetch`, with the CSRF token read from the cookie since the header used to be set by the `$.ajaxSetup` call in `base.html`, a `FormData` instead of `serializeArray`, and a click listener on the document, which also picks up a task link added to a page after it loads.

`base.html` wraps the django messages in a `#messages` container, so the same alerts can be added to the page from javascript.

The monolith repository page is converted with it: its inline script is gone, replaced by `data-task-label="Repository sync"` on the link.

**2. The ten other pages follow**, one commit each, or grouped when they are the same page in five copies:

| commit | pages | flavour |
| --- | --- | --- |
| osquery | distributed query results | export, format in the query string |
| santa | targets | export, search and format in the query string |
| inventory | machine list | export, search and format in the query string |
| inventory | macOS, iOS, android apps, programs, debian packages | export, format and form in the body |
| mdm | DEP virtual server devices sync | sync, then reload |
| google_workspace | connection group tag mappings sync | sync, then reload |

Nothing changes on the wire: the pages that posted a body still post the same one, the pages that posted none still post none. Where an inline script held more than the task code — the machine list filters and charts, the DEP virtual server beta enrollment tokens — only the task functions and their click binding were removed.

**Around 500 lines of duplicated javascript deleted.** Every task page now reports what happened.

## Notes for the reviewer

* Every `.task` link is bound, so a link without a label is not silently inert: it says *undefined started.* The label is required.
* The monolith sync button is disabled with the bootstrap `disabled` class instead of the `disabled` property, which has no effect on a link. A second click used to start a second task that could only be skipped. The DEP virtual server and google workspace buttons had the same bug.
* `SKIPPED` is only returned by the monolith sync task today. Reading it in the shared code makes it the convention for the tasks that decline to run.
* `tests.monolith` (559), `tests.osquery` (563), `tests.santa`, `tests.inventory` (1225 together), `tests.mdm` and `tests.google_workspace` (2711 together) pass. The bundle was built and exercised in a browser against stubbed responses: launch failure, poll failure, done then reload, skipped, download, a click on the icon inside a link, a link added after load, and a link without a label. The POST bodies, headers and URLs are the ones the pages sent before.

### DIFF
--- a/server/static_src/js/main.js
+++ b/server/static_src/js/main.js
@@ -78,3 +78,107 @@ $(document).ready(function () {
         }
     }
 });
+
+
+// tasks
+//
+// A link with the task class posts to its href to launch a celery task, then polls
+// the task result until it is ready. The messages are built with the task label. A
+// task result with a file sends the browser to the download URL, without one the
+// page is reloaded to display what the task changed. An export link carries the
+// format in data-format, and its enclosing form is posted with it.
+
+const TASK_POLLING_DELAY = 1000
+const TASK_RELOAD_DELAY = 1000
+
+function csrfToken() {
+    const cookie = document.cookie.split("; ").find(c => c.startsWith("csrftoken="))
+    return cookie ? decodeURIComponent(cookie.slice("csrftoken=".length)) : ""
+}
+
+function getJSON(url, options) {
+    return fetch(url, options).then(response => {
+        if (!response.ok) {
+            throw new Error(`HTTP ${response.status}`)
+        }
+        return response.json()
+    })
+}
+
+function addMessage(message, tag) {
+    const alert = document.createElement("div")
+    alert.className = `alert alert-${tag === "error" ? "danger" : tag} alert-dismissible fade show`
+    alert.setAttribute("role", "alert")
+    alert.textContent = message
+    const closeButton = document.createElement("button")
+    closeButton.type = "button"
+    closeButton.className = "btn-close"
+    closeButton.setAttribute("data-bs-dismiss", "alert")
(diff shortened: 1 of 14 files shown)


soysauceconspiracy (dismissed):
LGTM

soysauceconspiracy (approved):
LGTM

PR #1584, opened 2026-08-07
Stop the TOTP tests failing on a time step boundary

np5:
## Problem

`test_add_totp_ok` and `test_login_totp_ok` fail intermittently in CI, for example in [this run](https://github.com/zentralopensource/zentral/actions/runs/31210486923/job/92972969181):

```
FAIL: test_add_totp_ok (tests.server_accounts.views.test_users_views.AccountUsersViewsTestCase.test_add_totp_ok)
AssertionError: False is not true : Template 'accounts/user_verification_devices.html' was not a template
used to render the response. Actual template(s) used: accounts/add_totp.html, ...
```

The form was re-rendered because the code was refused. Both tests generate a code with `TOTP(secret).now()` and post it, and both views verify with `pyotp`'s default `valid_window=0`: `AddTOTPForm.clean` in `server/accounts/forms.py`, and `UserTOTP.verify` in `server/accounts/models.py`. A code generated in the last moments of its 30 second time step has expired by the time the view verifies it.

## Change

A helper waits for the next time step when the current one is about to end, so the code is always verified inside the step it was generated in. It costs a 2 second wait in the few percent of runs that land near a boundary.

Only the tests change. Whether the views should accept the adjacent codes, as most TOTP implementations do, is a separate question.

### DIFF
--- a/tests/server_accounts/views/test_users_views.py
+++ b/tests/server_accounts/views/test_users_views.py
@@ -1,6 +1,7 @@
 import base64
 import json
 import re
+import time
 from unittest.mock import patch
 
 import pyotp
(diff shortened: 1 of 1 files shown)


soysauceconspiracy (approved):
LGTM

PR #1586, opened 2026-08-08
mdm: paginate the remaining API list endpoints

np5:
Eleven MDM list endpoints returned their full queryset in a single response. They now use `MaxLimitOffsetPagination` like the rest of the MDM API:

`acme_issuers/`, `scep_issuers/`, `filevault_configs/`, `recovery_password_configs/`, `software_update_enforcements/`, `ota_enrollments/`, `blueprints/`, `blueprint_artifacts/`, `locations/`, `location_assets/`, `push_certificates/`

`location_assets/` is the one that grows without bound; the others are configuration tables, but they were the last MDM list views without a `pagination_class`.

### Ordering

The ACME issuer, SCEP issuer, software update enforcement, blueprint artifact and location asset models have no `Meta.ordering`. Offset pagination over an unordered queryset can drop or repeat rows between pages, so those list querysets are ordered explicitly — by `name` where it is unique, by `pk` otherwise.

### Backward incompatibility

Those responses change from a bare list to the `{count, next, previous, results}` envelope. Added to the changelog, and the Terraform Provider will need a release built against an updated [goztl](https://github.com/zentralopensource/goztl).

### Tests

Full `tests.mdm` run: 2605 tests, OK. `mkdocs build --strict` passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ Repository syncs now run in the background. Repositories that need more than the
 
 ### Backward incompatibilities
 
+#### 🧨 MDM API endpoint pagination
+
+The API endpoints for the ACME issuer, SCEP issuer, FileVault configuration, recovery password configuration, software update enforcement, OTA enrollment, blueprint, blueprint artifact, location, location asset and push certificate lists are paginated now. Remember to upgrade the Terraform Provider.
+
+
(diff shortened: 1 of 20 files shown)


soysauceconspiracy (approved):
LGTM

PR #1588, opened 2026-08-08
munki: split the MunkiState contact timestamps

np5:
## Why

`MunkiState.last_seen` was `auto_now`, so **any** save moved it — including `force_full_sync()`, which does a bare `self.save()`. It therefore could not tell how recently a machine had actually reported, and the two Prometheus metrics built on it were wrong as a result:

- forcing a full sync on a long-dead machine moved it straight into the youngest bucket of `zentral_munki_active_machines_bucket`;
- and dragged every one of that machine's `ManagedInstall` rows into the youngest bucket of `zentral_munki_installed_pkginfos_bucket` with it.

A bulk force-sync visibly inflated "active machines". Neither metric could be fixed without a split, because there was nothing honest to point them at.

Separately, `MunkiState` rows were created *only* by the postflight, so a machine that polls `job_details` forever without ever completing a run left no server-side trace at all.

## What changed

| field | change | written by |
| --- | --- | --- |
| `last_seen` | renamed to `updated_at`, still `auto_now` | any save — honestly named now |
| `last_preflight_at` | new | `JobDetailsView` |
| `last_postflight_at` | new | `PostJobView` |
| `created_at` | new, for the codebase-standard pair | — |

- Both metric queries now read `last_postflight_at`. A machine that has only ever preflighted has a null age and is counted in `+Inf` alone, which is correct: it has never been active.
- `JobDetailsView` creates and updates the row, recording `last_preflight_at`.
- Migration `0015` renames, adds the three fields, and backfills `created_at` / `last_postflight_at` from the old `last_seen`.

## Notes for review

**The `job_details` response is unchanged.** The state is read with `.first()` *before* the response is built and the preflight write happens *after*, so the agent-facing payload is byte-identical. Gating the existing `managed_installs` / `last_seen_sha1sum` blocks on the newly-created row would have started sending `managed_installs: true` and `last_seen_sha1sum: null` on a machine's first contact, where those keys are currently absent. There is a test pinning this.

**Backfill accuracy.** The old `last_seen` is the best approximation of the last postflight available: only a row whose *last* write was a `force_full_sync()` is off, it errs on the recent side, and it self-corrects on that machine's next postflight. So no row is made worse than it is today. `end_time` was considered instead and rejected — it is only set when a report is included, and it is the agent's clock rather than server receipt time.

**Two small intentional behaviour changes.** `ForceFullSync.check_machine()` and the force-full-sync view now find a row for a machine that has only preflighted, so the action becomes available slightly earlier. Forcing a sync on such a machine is a no-op in effect, since its next preflight would request everything anyway.

## Testing

299 munki tests pass. Coverage on the changed code: `metrics_views.py` 100%, `public_views.py` 100%, migration 100%; `models.py` 99% with both misses pre-existing and outside this diff.

Seven tests added, including a regression test that `force_full_sync()` no longer rejuvenates a dead machine's metric bucket, and coverage of the null-`last_postflight_at` case landing in `+Inf` alone.

The migration's `RunPython` gets line coverage from the test-database build, which runs against an empty table, so its data logic is not behaviourally asserted. There is no migration-test harness in `tests/` and the existing `RunPython` migrations follow the same pattern; happy to add a `MigrationExecutor`-based test if wanted.

### DIFF
--- a/tests/munki/test_metrics_views.py
+++ b/tests/munki/test_metrics_views.py
@@ -23,9 +23,11 @@ class MunkiMetricsViewsTestCase(TestCase):
             failed_at=datetime.now() if failed else None,
             reinstall=reinstall,
         )
-        last_seen = naive_utcnow() - timedelta(days=age_days)
-        ms = MunkiState.objects.create(machine_serial_number=mi.machine_serial_number)
-        MunkiState.objects.filter(pk=ms.pk).update(last_seen=last_seen)
+        # age_days None ⇒ the machine has never posted a postflight
+        MunkiState.objects.create(
+            machine_serial_number=mi.machine_serial_number,
+            last_postflight_at=None if age_days is None else naive_utcnow() - timedelta(days=age_days),
+        )
         return mi
 
     def _make_authenticated_request(self):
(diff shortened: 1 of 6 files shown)


soysauceconspiracy (approved):
LGTM

PR #1589, opened 2026-08-08
Add a read-only compliance-check-id to serializers

np5:
Adds `compliance_check_id` to the osquery `QuerySerializer`, the inventory `JMESPathCheckSerializer` and the munki `ScriptCheckSerializer`, so API clients can correlate these objects with their compliance check statuses without a second lookup.

Read-only, and declared the same way as the turbo `ScriptSerializer` / `MSCPCheckSerializer` already do it.

## One fix this needed

The osquery query is the only one of the three whose compliance check FK is nullable — the other two are non-null `OneToOneField`s, so their id is always there.

Disabling a query's compliance check deletes it, and since the FK is `on_delete=SET_NULL`, the delete only NULLs the column. The in-memory instance the DRF response is serialized from kept pointing at the deleted check, so the PUT that disabled it would have echoed back a stale `compliance_check_id`. `sync_query_compliance_check` now clears the attribute, as the turbo one already does.

That also fixes an existing bug on the same path: the response to the request that disabled a check reported `compliance_check_enabled: true`, because that field reads the same stale attribute. `test_update_query_disable_compliance_check` covers both.

## Notes for the reviewer

- The four osquery query response examples in the docs gained the field. The request payload example is untouched — it is read-only.
- Munki script checks and inventory JMESPath checks have no REST API section in the docs, only Terraform resources, so there was nothing to update for them.
- `tests.osquery tests.inventory tests.munki` pass (1432 tests), `ruff check` is clean, and `mkdocs build --strict` exits 0.

### DIFF
--- a/docs/apps/osquery.md
+++ b/docs/apps/osquery.md
@@ -1339,6 +1339,7 @@ Response:
     {
         "id": 1,
         "compliance_check_enabled": false,
+        "compliance_check_id": null,
         "name": "GetApps",
         "sql": "SELECT * FROM apps;",
         "platforms": [],
@@ -1392,6 +1393,7 @@ Response:
 {
 	"id": 1,
 	"compliance_check_enabled": false,
+	"compliance_check_id": null,
 	"name": "GetApps",
 	"sql": "SELECT * FROM apps;",
 	"platforms": [],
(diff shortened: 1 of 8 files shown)


soysauceconspiracy (approved):
LGTM

PR #1590, opened 2026-08-09
santa: fix the false sync incidents

np5:
The Santa sync incidents fire for machines that are perfectly in sync. Most of it comes from the preflight comparing the rules the client reports with the rules Zentral synced, using assumptions that stopped being true when Santa moved its sync protocol to protobuf.

Each commit is independent and can be reviewed on its own.

### The reported rule counts

Santa serializes the preflight request with the protobuf JSON mapping, which omits the fields set to their default value, so a count of 0 is simply not sent. The enrolled machine kept its previous value forever, and a machine whose rules were removed stayed permanently out of sync with an incident that could not be closed. All the counts are now reset when they are missing from the request. The overflow guard is fixed at the same time, it let `2147483648` through and triggered a numeric value out of range error.

### The transitive rules

The client creates the transitive rules itself, and stores them as binary rules, so they are included in the reported binary rule count. They are never synced, and they were not subtracted from the reported count: every machine of a configuration with the transitive rules enabled was reported out of sync as soon as a compiler allowlisted a binary. Santa excludes them from its own rules hash for the same reason.

### The comparison and the clean sync

The comparison was made after the clean sync had deleted all the machine rules, but the client only clears its own rule database during the rule download stage, and still reports its rules during the preflight. Every clean sync therefore opened an incident, closed during the next sync. The comparison is now made first, and skipped for a machine that just enrolled, which has no synced rule to be compared with.

The clean sync trigger had the same root cause as the rule counts: "all the rule count keys are missing" now also matches a client that simply has no rule, and a clean sync was forced during each preflight of those machines. It is now based on the reported counts, and only rebuilds the machine rules if some rules were really synced with the machine.

### The incident updates

No update was sent the first time a machine was found in sync. The enrolled machines are replaced during a re-enrollment though, and the new one starts without a known sync state, so the incident of a machine that was out of sync before its re-enrollment could never be closed.

Finally, the incident updates of the replaced enrolled machines are carried by the enrollment event, and applied with the serial number of the machine that just enrolled. The machine ID reported by santa defaults to the hardware UUID, but it can also be a fixed value from the configuration profile, or read from a plist. When it is not unique, the machines keep replacing each other, and each enrollment was closing an incident still open for another machine.

### Not fixed here

Santa also adds the rules from the `StaticRules` configuration profile key to the counts it reports. Zentral has no way to know about them, so those machines still compare as out of sync.

A clean sync that downloads no rule does not clear the client rule database at all: the rule download stage returns before the cleanup is applied. Zentral has already dropped its machine rules by then. This one needs a fix in the client, and a follow up on the Zentral side to only commit the machine rules once the postflight confirms the sync.

### DIFF
--- a/tests/santa/test_configuration.py
+++ b/tests/santa/test_configuration.py
@@ -1,9 +1,28 @@
 from django.test import TestCase
 from django.utils.crypto import get_random_string
 from zentral.contrib.santa.models import Configuration
+from zentral.core.incidents.models import Severity
 
 
 class SantaConfigurationTestCase(TestCase):
+    # get_sync_incident_severity
+
+    def test_get_sync_incident_severity(self):
+        config = Configuration.objects.create(name=get_random_string(256),
+                                              sync_incident_severity=Severity.MAJOR.value)
+        self.assertEqual(config.get_sync_incident_severity(), Severity.MAJOR)
+
+    def test_get_unknown_sync_incident_severity_none(self):
+        config = Configuration.objects.create(name=get_random_string(256))
+        Configuration.objects.filter(pk=config.pk).update(sync_incident_severity=42)
+        config.refresh_from_db()
+        with self.assertLogs("zentral.contrib.santa.models", level="ERROR") as cm:
+            self.assertEqual(config.get_sync_incident_severity(), Severity.NONE)
+        self.assertEqual(
+            cm.output,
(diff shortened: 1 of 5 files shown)


np5:
#1592 is stacked on this branch.

soysauceconspiracy (approved):
LGTM

PR #1592, opened 2026-08-09
santa: commit the machine rules when the client confirms the sync

np5:
#1590 fixed the rule comparison itself and is merged; this branch is rebased on `main` and is seven commits. This one fixes what the comparison is made against: the machine rule ledger records rules the client may never have written to its own database.

### The enrolled machine cache is gone

The sync views cached the enrolled machine and its tag ids for ten minutes. The preflight wrote that cache, the rule download, the event upload and the postflight read it, and the postflight deleted it, so an entry only ever served one sync run. What it served could be older than the configuration it carried: turning the client certificate requirement on did not reach a machine with a cached entry.

Removing `MetaBusinessUnitTag` left the tags of a machine one indexed lookup on its serial number, so they no longer justify a cache of their own. They are now annotated on the query that reads the enrolled machine, and every stage works on the machine and the tags it was just given. A stage costs one query instead of the two a cache miss used to cost, and the preflight, the event upload and the postflight stop resolving the tags that only the rule download reads.

Nothing in the sync path can read a stale session anymore, and the session columns are written with targeted updates, so a stage never writes back the fields it read.

### The commit point

The client downloads the rules batch by batch, but it buffers them, and only writes them all to its rule database once the last batch is in. It then sends the postflight. A machine rule is currently considered synced as soon as the client asks for the next batch, which is several steps too early.

A rule download interrupted after the second batch therefore leaves the first batch recorded as synced, while the client wrote nothing. The rules of that batch are never sent again, the machine stays short of them for good, and the incident that comes with it can never be closed. The in flight batch was already discarded when the next session started, so this only affects the batches acknowledged during the interrupted session.

The rules of a sync session are now staged, and only recorded once the postflight confirms them. The postflight is only sent if the rule download and the database write both succeeded, so it is the only reliable signal we have.

### The lost sessions

Staged rules cannot simply be dropped during the next preflight: the sync may well have worked, with only the postflight lost on the way. The reported rule counts tell the two outcomes apart. They either match the ledger, and the session never reached the client, or they match the ledger updated with the session, and only the postflight was lost. Anything else is a machine out of sync, and the staged rules are discarded and sent again.

The removals are staged with their own flag, and keep the policy and the version of the rule the client has: once a rule is removed from a configuration, the ledger is the only place where it still exists, and dropping it would strand the copy the client has.

### The clean syncs

A clean sync was applied during the preflight, but the client only clears its own database during the rule download. Nothing was left to compare the client with in between, and an interrupted clean sync left the two sides disagreeing.

Worse, santa returns from the rule download stage before its own cleanup when it receives no rule at all. A clean sync without any rule to send - a machine re-enrolled in a configuration that has no rule for it - dropped the whole ledger while the client kept every rule it had. The machine was then permanently out of sync, with no rule left to send it, and no way to ever remove the ones it kept.

The clean sync is now applied during the postflight, and only if rules were really sent, so both sides keep their rules when the client skips its cleanup. The stale rules are removed with normal REMOVE rules during the next sync.

The rules committed before a clean session are replaced in place, so the ledger of a lost clean session cannot be restored. Another clean sync is requested instead, and the machine is not compared until it is over.

None of it has been reaching the client, because the sync type was sent lowercase. Santa lowercased what it received until 2024.5, so the value was picked up. 2024.6 moved the sync protocol to protobuf JSON, where the sync type is an enum whose values are `NORMAL`, `CLEAN` and `CLEAN_ALL`: the client parses the response with `ignore_unknown_fields` and without case insensitive enum parsing, and protobuf drops an unknown enumerator instead of failing, so `has_sync_type()` is false, the deprecated `clean_sync` boolean is not sent to clients from 2024.1 on, and the sync runs as a normal one.

Every clean sync asked of a client from 2024.6 on has therefore been a normal sync: nothing was dropped, and a machine that lost its rule database never rebuilt it. The ledger stayed consistent throughout, because the postflight commits the session with the sync type the client reports, and the client reports `NORMAL`. Upper case is right on both sides of that change - the older clients lowercase the value before comparing it, the newer ones match the enum value name.

### The acknowledgement round trip is gone

A cursor was sent with every batch that had rules, so the client always made one last request, getting nothing back, only to acknowledge the batch before it. That extra round trip was the only way to record the last batch: there was nowhere else to keep its cursor.

The postflight now commits the whole session, by session and not by cursor, so the last batch does not need to be acknowledged anymore. One extra rule is read from the database with each batch, to know whether another one is necessary without a second query, and the cursor is only sent when it is. The cursor is still cleared batch after batch, to keep the rules of the previous batches out of the cleanup of the next request.

A sync that has nothing to deliver is unchanged, it was already a single request. The one saved lands on the syncs that do send rules: enrollments, clean syncs and rule rollouts.

### The sync stage timestamps

`EnrolledMachine.updated_at` is `auto_now`, it moves on any save and cannot say how recently a machine reported. The inventory source grain does not answer it either: the preflight is what commits the machine snapshot, so a machine that polls forever without ever confirming a sync still looks perfectly alive there. That is exactly the failure this PR is about.

`last_preflight_at` and `last_postflight_at` are written by the stages that already update the enrolled machine, so no extra query is made, and a rule download that starts a session without a preflight does not stamp anything. Both are indexed: one row per machine and per enrollment is fleet sized. Same shape as the munki contact timestamps, and the columns a staleness watch needs to tell a machine that stopped syncing from one that never completes a run.

### Tests

Nothing was driving the sync stages as a sequence, only the views one by one, which is why all of this went unnoticed. `SantaSyncClient` simulates the client rule database and its stages, and can be stopped at any point. It models the parts of the client that matter here: the rules are only written once the whole download is over, the cleanup is skipped when no rule is received, and a clean sync keeps the transitive rules.

The four broken sequences are added as expected failures in the first commit, and turned into normal tests by the last one.

### Upgrade note

The sync session columns and the timestamps are folded into the same migration, `santa.0042`. A deployment that already applied an earlier version of that migration from this branch has it recorded as applied and will silently miss the new columns: `migrate santa 0041` then forward.


sebastian-zentral (dismissed):

soysauceconspiracy (approved):
LGTM

PR #1593, opened 2026-08-09
stores: fix the ClickHouse JSON columns deserialization

np5:
## Problem

A machine event page could display a payload value as a Python bytes literal:

```
{'pk': 312,
 'previous_status': {'status': 'OPEN',
                     'status_time': b'\x13\tX\x87I\xa9\xf1e\xc7\x18'},
 'status': 'CLOSED',
 'status_time': datetime.datetime(2026, 8, 3, 13, 56, 8, 367606),
 'type': 'munki_install_failed'}
```

The value is not corrupted. It is a ClickHouse binary encoded value: `0x13` is the
`DateTime64(P)` type tag, `0x09` the precision, and the last 8 bytes a little endian
tick count — here `2026-07-31 14:24:25.877383`.

## Cause

`metadata` and `payload` are `JSON` columns. That type only keeps `max_dynamic_paths`
paths (1024 by default) as sub columns; the extra ones go to the shared data structure,
where the values keep their binary type encoding. `clickhouse-connect` reads those as
opaque strings rather than decoding the embedded type tag, and hands back raw bytes for
the types it does not reconstruct. String, Int, Float and Bool survive — `Date`,
`DateTime64` and `Array` do not.

Since every event type shares the same `payload` column, the path budget is largely
exceeded, and which paths keep a sub column is decided by how often they are non null.
That is what makes the symptom look arbitrary: in the payload above, `status_time` is
set on every event of the type and decodes fine, while `previous_status.status_time` is
only set when a status actually changes, loses its sub column, and comes back as bytes —
right next to `previous_status.status`, which is a String and is unaffected.

## Fix

Let ClickHouse serialize both columns with `toJSONString()` and parse them back, instead
of having the client rebuild them from the sub columns. The datetimes ClickHouse infers
are then formatted by the server, hence the `iso` `date_time_output_format`.

## Tests

`tests.core_stores` passes (266 tests). The new
`test_fetch_machine_events_json_shared_data` stores an event whose payload exceeds the
path budget and checks the round trip; it fails on the unpatched backend.

### DIFF
--- a/tests/core_stores/test_clickhouse.py
+++ b/tests/core_stores/test_clickhouse.py
@@ -7,8 +7,9 @@ from zentral.core.probes.models import ProbeSource
 from zentral.core.probes.probe import Probe
 from zentral.core.stores.backends.all import StoreBackend
 from zentral.core.stores.backends.clickhouse import ClickHouseStore, ClickHouseStoreSerializer
+from zentral.utils.time import naive_utc_fromisoformat
 from .utils import build_login_event, force_store
-from . import BaseTestStore
+from . import BaseTestStore, get_from_dt, make_event
 
 
 class TestClickHouseStoreStorage(TestCase, BaseTestStore):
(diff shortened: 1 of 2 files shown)


soysauceconspiracy (approved):
LGTM

PR #1595, opened 2026-08-10
mdm: attribute a DEP device synchronization to the user who asked for it

sebastian-zentral:
Two small fixes in the same call, split out of #1581 so the bug fixes land separately from the new events.

## The task was not linked to a user

The task list had nothing to show in its user column for a DEP synchronization. Zentral creates the `UserTask` from the `before_task_publish` signal in `server/server/celery.py`, but only when the task kwargs carry `task_user` — and `DEPVirtualServerSyncDevicesView` never passed it. There was no row to render, so the column was structurally always empty.

It is passed now, the way `sync_repository_task` already does it for the monolith repository sync. A synchronization somebody triggered can be told apart from one Zentral ran on its own.

## `?full_sync` has never worked

Passing `task_user` exposed a bug in the same call:

```python
apply_async((server.pk,), force_full_sync=full_sync)
```

`Task.apply_async` is `(args=None, kwargs=None, task_id=None, producer=None, link=None, link_error=None, shadow=None, **options)`, so `force_full_sync` landed in `**options` — delivery settings such as `countdown` or `queue` — and never reached the task, which always ran with its default `False`.

**The `full_sync` query parameter has therefore never forced a full synchronization**, neither from the HTTP API nor from the *Synchronize* button. The existing test only checked that the response carried a `task_id`, so nothing caught it.

Both values are now passed in the kwargs dict `apply_async` expects, and the task takes `**kwargs` to absorb `task_user`, mirroring the monolith task.

## Tests

Three tests, each of which fails without the fix:

- the task kwargs a full synchronization is published with, asserted as a whole so an option cannot quietly replace a kwarg again
- the same for a delta synchronization triggered by a service account
- the `UserTask` really being created, and pointing at the right user

`tests.mdm`, `tests.server_accounts` and `tests.server_base` are green (3120 tests).

### DIFF
--- a/tests/mdm/test_api_dep_views.py
+++ b/tests/mdm/test_api_dep_views.py
@@ -1,3 +1,4 @@
+import uuid
 from unittest.mock import patch
 from urllib.parse import urlencode
 
@@ -6,7 +7,7 @@ from django.test import TestCase
 from django.urls import reverse
 from django.utils.crypto import get_random_string
 
-from accounts.models import APIToken, User
+from accounts.models import APIToken, User, UserTask
 from tests.zentral_test_utils.login_case import LoginCase
 from tests.zentral_test_utils.request_case import RequestCase
 from zentral.contrib.inventory.models import MetaBusinessUnit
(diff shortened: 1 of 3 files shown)


np5 (approved):
LGTM

PR #1596, opened 2026-08-10
tasks: attribute the remaining background tasks to the user who started them

sebastian-zentral:
Nine tasks a user or a service account can start were not linked to whoever started them. 

## Why it matters

`TaskViewMixin.get_queryset()` in `server/accounts/views/tasks.py` filters the task list for everybody who is not a superuser:

```python
if not self.request.user.is_superuser:
    queryset = queryset.filter(usertask__user=self.request.user)
```

`TaskView.get_object()` uses the same queryset, so a task with no `UserTask` row does not merely show an empty user column — it does not appear in the list at all, and its detail page 404s. The download button for a finished export lives on that page (`server/templates/accounts/tasks/_result.html`).

Zentral creates the `UserTask` from the `before_task_publish` signal in `server/server/celery.py`, but only when the published task kwargs carry `task_user`. None of these nine launches passed it.

## What changed

One commit per launch. Each passes `{"task_user": request.user.id}` as the task kwargs and takes `**kwargs` on the task so the extra kwarg does not raise, following what the inventory exports and the monolith repository sync already do. No task reads the value: it is consumed at publish time by the signal receiver.

### Exports that were unreachable

Both are permission-gated — `osquery.view_distributedqueryresult` and `santa.view_target` — so a user without the superuser flag could click *Export*, get a `201` with a task id, and never be able to reach the file.

| Task | Module |
| --- | --- |
| `export_distributed_query_results` | osquery |
| `export_targets` | santa |

### Tasks started from a button, invisible to the user who started them

| Task | Note |
| --- | --- |
| `sync_group_tag_mappings_task` | Google Workspace group tag mappings sync. It already passed the serialized event request, so it was attributable in the event store but not in the task list |
| `bulk_assign_location_asset_task` | MDM bulk location asset assignment. The view redirects with a *"task launched"* message, so the task list is the only place to follow it |

### API-token-only launches

Nothing is broken here — there is no button pointing a user at a task they cannot see. This is consistency, and it makes an API-driven cleanup or export attributable to the integration that triggered it. On these endpoints the task belongs to the **service account** that called them, which the tests assert explicitly.

| Task | Module | Note |
| --- | --- | --- |
| `cleanup_inventory` | inventory | already passed the event request |
| `export_full_inventory` | inventory | produces a file; the other inventory exports already pass `task_user` |
| `sync_software_updates_task` | mdm | |
| `sync_inventory` | intune (`ee/`) | |
| `sync_inventory` | wsone (`ee/`) | already passed the event request |

The two `ee/` changes are edited in place. No code moved between the licence trees.

## Tests

One test per launch, asserting the `UserTask` exists and points at the right user. Each was checked to fail without its fix (`accounts.models.UserTask.DoesNotExist`).

For the MDM bulk assignment the assertion was added to the existing `test_associate_location_asset_post`, which already sets up the POST and inspects the `TaskResult`, rather than duplicating the setup.

### DIFF
--- a/ee/zentral/contrib/intune/api_views.py
+++ b/ee/zentral/contrib/intune/api_views.py
@@ -38,7 +38,9 @@ class StartTenantSync(APIView):
 
     def post(self, request, *args, **kwargs):
         tenant = get_object_or_404(Tenant, tenant_id=self.kwargs["tenant_id"])
-        result = sync_inventory.apply_async(kwargs={'tenant_id': tenant.tenant_id})
+        result = sync_inventory.apply_async(
+            kwargs={'tenant_id': tenant.tenant_id, 'task_user': request.user.id}
+        )
         return Response({"task_id": result.id,
                          "task_result_url": reverse("base_api:task_result", args=(result.id,))},
                         status=status.HTTP_201_CREATED)
--- a/ee/zentral/contrib/intune/tasks.py
+++ b/ee/zentral/contrib/intune/tasks.py
@@ -49,7 +49,8 @@ def do_sync_inventory(client):
 
 
 @shared_task
-def sync_inventory(tenant_id):
+def sync_inventory(tenant_id, **kwargs):
+    # kwargs absorbs task_user, added by the API view for the UserTask created in the celery signal
     tenant = Tenant.objects.get(tenant_id=tenant_id)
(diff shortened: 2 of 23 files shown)


np5 (approved):
LGTM

PR #1599, opened 2026-08-10
mdm: fix the DEP synchronization and post events when Apple changes a device

np5:
The synchronization with Apple Business Manager was the one DEP workflow that left no trace in the event pipeline, and getting there turned up a handful of things wrong with the synchronization itself. Those come first, one commit each, so the ones most likely to be argued about are last and can be dropped without unpicking the rest.

## The synchronization

**It waited on the advisory lock.** A synchronization that lost the race parked a worker, and an open Postgres transaction, for up to ten minutes under `SET LOCAL lock_timeout = '600s'`, then died with a bare `OperationalError`. Waiting achieves nothing: the synchronization is cursor driven, so whatever the running one does not cover the next one picks up. `pg_try_advisory_xact_lock` replaces the blocking one, the way `sync_repository_task` already does it for the monolith repositories. The lock had to move to the callers — it was taken inside a generator, and a generator that returns early yields nothing, which is indistinguishable from a synchronization that found nothing to change. The task reports a `SKIPPED` status instead of a `SUCCESS` one, and the management command says so.

**`updated_at` did not move.** Five bulk updates change DEP devices through a queryset, which does not call `save()`, so the `auto_now` never fired. `DEPDeviceList` orders on `updated_at`, so a device the synchronization had just marked as deleted sorted as if nothing had happened to it, and a client paging by `updated_at` to pick up what changed missed all of it. Zentral also stops inventing `profile_assign_time` from its own clock; the field stays null until Apple reports the real one.

**Everything Apple reported was written.** A full fetch reports the whole fleet, and every device it reported was written whether or not anything had moved — rewriting the entire table on every run — while the task counted every device it had not created as updated, so a fleet that had not moved at all was reported as thousands of updates. The stored record is now compared with what would be written, and unchanged devices are left alone. The comparison is between serializations, not attributes: the datetimes parsed from the DEP API are aware and the stored ones are naive, so comparing attributes would find a difference in every timestamp on every run.

**A failed assignment discarded the whole synchronization.** The default enrollment was assigned at the end, inside the transaction that holds the lock, and the cursor is saved before it but in the same transaction — so one throttled `assign_profile` call discarded a completed delta walk and the cursor advance with it. It is a task of its own now, scheduled on commit, deriving its work list from the database rather than being handed the serial numbers: passing them would not fit in a broker message for a large virtual server, and deriving them makes the task idempotent, which is what lets it take the same lock and simply skip when a synchronization is running. It also makes it self healing — a device whose assignment failed is retried instead of waiting for Apple to report an operation on it — so the first run after this lands catches up on every device left unassigned by an earlier failure. It then reads the records back from Apple rather than assuming what the assignment made of them: the profile status is Apple's to decide, and the assignment time is Apple's to stamp.

## The events

Every device the synchronization really changes posts a `dep_device_change` event. The payload is an audit event's, so a DEP device has one history whichever end changed it and one shape to read it with, and the class is an `AuditEvent` subclass rather than a copy of its builder. The event type is not an audit one: these are machine generated, one per device of a virtual server, and they would bury the operator actions in the audit trail. The `zentral` tag goes with the type.

An `origin` says which end was talking to Apple, a synchronization or the default enrollment assignment. The events of one run share their metadata `id`, and a synchronization somebody asked for carries the request that asked for it.

Three kinds of change post nothing: the first synchronization of a virtual server, which imports the fleet rather than changing it; a device Apple reports unchanged; and a device whose record only moved because of the operation itself, `last_op_type` and `last_op_date`, which a delta synchronization reports for attributes Zentral does not store — though a deletion is reported even when it is all that moved. A device Apple stops reporting is an `updated`, not a `deleted`: the record stays, with `last_op_type` moved to `deleted`.

The event type gets its own Elasticsearch mapping entry mirroring the audit one, since the same free-form previous and new values have to stay out of the mapping.

## Requests sized from the account detail

The account detail carries a `urls` array of `{uri, http_method, limit}`, and Apple documents both objects — [Url](https://developer.apple.com/documentation/devicemanagement/url) and [Limit](https://developer.apple.com/documentation/devicemanagement/limit), "a ranged limit" of `{default, maximum}`. Against a real ABM account, `/server/devices` and `/devices/sync` advertise `{default: 1000, maximum: 1000}`, and no other endpoint advertises anything.

The client asked for 100 per page, hard coded, so **a full fetch of a large virtual server made ten times the requests Apple was willing to answer in one**. It now reads the maximum from the account detail, once per client, falling back to 100 when nothing is advertised or the account detail cannot be read.

Three call paths could also send an unbounded number of devices in one request — the default enrollment assignment, the profile payload of `define_dep_profile`, and whatever list `get_devices` and `disown_devices` were given. Apple does not reject those: a `/devices` request carrying 5000 serial numbers is answered in full. They are batched anyway, so that a batch is applied as it succeeds and one failure does not discard the rest.

Rate limiting was looked for and is not there. No DEP response carries a `Retry-After` or an `X-RateLimit` header — only `X-Apple-Request-UUID` and `X-Apple-Edge-Response-Time` — and the [documented error codes](https://developer.apple.com/documentation/devicemanagement/interpreting-automated-device-enrollment-error-codes) are 400, 401 and 403. Apple's answer to how much may be asked for is the account detail, not a response header. The assignment task still backs off exponentially on a 429 or a 5xx, which is now known to be defensive rather than load bearing.

## One smaller gap

Assigning an enrollment profile to a DEP device from the web interface posted an audit event; the same assignment through the HTTP API changed the device and recorded nothing. `RetrieveUpdateAPIViewWithAudit` is split out of the retrieve/update/destroy one, which now inherits from it and adds the destroy — a DEP device is never deleted.

## ⚠️ API change

The result of `/api/mdm/dep/virtual_servers/<pk>/sync_devices/` gains `unchanged` and `marked_deleted`, and a `status` of `SUCCESS` or `SKIPPED`, and `updated` drops to the devices that really changed. A client reading `operations.updated` as the number of devices Apple reported has to read `created + updated + unchanged` instead.

## Tests

Full suite green, 7734 tests. New coverage for the skipped-lock paths, the `updated_at` bumps, the batching and the profile split, the assignment task's work list and its retry behaviour, the created/unchanged/changed/marked-deleted paths, each event and each suppression rule, the account-detail limit resolution and its fallbacks, and the API audit event.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,11 @@ The API endpoints for the ACME issuer, SCEP issuer, FileVault configuration, rec
 The DEP token audit events do not carry the `has_expired` and `expires_soon` booleans anymore. They were computed when the event was serialized, and never refreshed afterwards, so a stored event kept reporting the state of the token at the time it was written. Use the `access_token_expiry` timestamp, which is still in the payload, and compare it with the event `created_at`. Probes keyed on the two removed fields need to be updated.
 
 
+#### 🧨 MDM DEP virtual server device sync task result
+
+In the result of `/api/mdm/dep/virtual_servers/<int:pk>/sync_devices/`, `operations` gains `unchanged` and `marked_deleted`, `updated` only counts the devices whose record changed, and a `status` of `SUCCESS` or `SKIPPED` is added. Read `created + updated + unchanged` for the number of devices Apple reported.
+
+
 #### 🧨 Monolith repository sync API endpoint
 
 `/api/monolith/repositories/<int:pk>/sync/` launches a background task instead of syncing the repository during the request. It responds with `201` and a `task_id`/`task_result_url` pair, in place of the `200 {"status": 0}` and `500 {"status": 1, "error": "…"}` responses it used to return. Poll `task_result_url` to know the outcome of the sync.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Recovery password configurations can schedule a password rotation after each rev
 
 FileVault configurations can schedule a PRK rotation after each reveal of the PRK.
 
+Synchronizing a DEP virtual server with Apple Business Manager now posts a `dep_device_change` event for every device it changed, with the previous and the new value of the record, like an audit event.
+
 
 #### Monolith
 
--- a/docs/apps/mdm.md
+++ b/docs/apps/mdm.md
@@ -579,11 +579,37 @@ Note that a device can also become unblocked without an operator: a full state p
 
 ### Automated Device Enrollment devices
 
-Assigning an enrollment profile to a device, and refreshing its record from Apple, are recorded with the `updated` action, tagged with the device serial number. The event reports every attribute Apple can change on the record, so a refresh shows what actually moved.
+Assigning an enrollment profile to a device, and refreshing its record from Apple, are recorded with the `updated` action, tagged with the device serial number. The event reports every attribute Apple can change on the record, so a refresh shows what actually moved. An assignment made over the HTTP API is recorded the same way as one made from the web interface.
 
 A refresh that Apple answers with an unknown serial number still marks the record as deleted, so it is recorded too. A profile assignment Apple refuses changes nothing and records nothing.
 
 Disowning a device has its own event, `dep_device_disowned`, rather than an audit event.
+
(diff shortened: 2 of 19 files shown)


sebastian-zentral (dismissed):
lgtm

sebastian-zentral (dismissed):

soysauceconspiracy (approved):
LGTM

PR #1600, opened 2026-08-10
mdm: report the enrollment in the enrollment session events

sebastian-zentral:
`EnrollmentSession.serialize_for_event()` takes an `extra_dict` from its four subclasses and drops it:

```python
def serialize_for_event(self, enrollment_session_type, extra_dict):
    d = {"pk": self.pk,
         "type": enrollment_session_type,
         "status": self.status}
    return {"enrollment_session": d}          # extra_dict never used
```

So `dep_enrollment_request`, `ota_enrollment_request`, `user_enrollment_request` and the re-enrollment events never said which enrollment was used, and all four subclasses were building a dict for nothing. The argument was removed from the body in an unrelated commit and the callers were never cleaned up.

## Two commits

**1. Nest the enrollment of a re-enrollment session.** Every other session wraps its enrollment under a key before handing it to the base serializer, which merges the dict into the session dict. The re-enrollment session passed it bare, so the enrollment's own `pk`, `name` and timestamps would land directly on the session and overwrite its `pk`.

Nothing changes in that commit, because the base serializer drops the dict anyway — the mistake is only invisible for that reason. It is fixed first so that the second commit does not turn it into a real bug.

**2. Apply the dict.** The enrollments are referenced **by their keys only**. These events are posted on every enrollment attempt, and the full serialization of an enrollment carries its realm, its timestamps and the metadata of its enrollment secret — the quota, the expiry, the restrictions — which is a lot to repeat for something the enrollment pk already identifies. The secret itself is never serialized either way.

The resulting payload, for a DEP enrollment session:

```json
{"enrollment_session": {"pk": 42, "type": "dep", "status": "STARTED",
                        "dep_enrollment": {"pk": 7, "name": "…", "uuid": "…"}}}
```

OTA, user and re-enrollment sessions get `{pk, name}`; the re-enrollment one nests under a neutral `enrollment` key, since its enrollment can be any of the three types.

## Tests

Three existing assertions were pinning the truncated payload and are updated. A new assertion covers the re-enrollment session and checks that the session `pk` survives, which is the bug the first commit prevents.

Full suite green (7607 tests).

### DIFF
--- a/tests/mdm/test_dep_enrollment.py
+++ b/tests/mdm/test_dep_enrollment.py
@@ -61,7 +61,10 @@ class TestDEPEnrollment(TestCase):
         self.assertEqual(session.status, "STARTED")
         self.assertEqual(
             session.serialize_for_event(),
-            {"enrollment_session": {"pk": session.pk, "type": "dep", "status": "STARTED"}}
+            {"enrollment_session": {"pk": session.pk, "type": "dep", "status": "STARTED",
+                                    "dep_enrollment": {"pk": enrollment.pk,
+                                                       "name": enrollment.name,
+                                                       "uuid": str(enrollment.uuid)}}}
         )
 
     def test_dep_enrollment_session_no_acme_payload(self):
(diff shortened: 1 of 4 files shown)


np5:
Superseded by #1601, which reports the enrollment at the top level of the payload, under a key derived from the enrollment model, instead of nested in the session.

PR #1601, opened 2026-08-10
mdm: report the enrollment in the enrollment session events

np5:
Supersedes #1600, which reported the enrollment nested in the session. This reports it at the top level, under a key derived from the enrollment model.

`EnrollmentSession.serialize_for_event()` took an `extra_dict` from its four subclasses and dropped it:

```python
def serialize_for_event(self, enrollment_session_type, extra_dict):
    d = {"pk": self.pk,
         "type": enrollment_session_type,
         "status": self.status}
    return {"enrollment_session": d}          # extra_dict never used
```

So the enrollment session events never said which enrollment was used, and all four subclasses were building a dict for nothing. The argument was removed from the body in an unrelated commit and the callers were never cleaned up.

## Report the enrollment

At the top level of the payload, next to the session, like the munki, osquery and inventory enrollment events already do — and under a key derived from the enrollment model: `dep_enrollment`, `ota_enrollment` or `user_enrollment`. For a DEP enrollment session:

```json
{"enrollment_session": {"pk": 42, "type": "dep", "status": "COMPLETED"},
 "dep_enrollment": {"pk": 7, "name": "…", "uuid": "…"}}
```

and for a re-enrollment session of that same enrollment, where only the session type differs:

```json
{"enrollment_session": {"pk": 51, "type": "re", "status": "COMPLETED"},
 "dep_enrollment": {"pk": 7, "name": "…", "uuid": "…"}}
```

Deriving the key gets three things at once. It is the same key as the object key of the enrollment's audit events, so both kinds of events answer to the same name. It is the same key whatever the session type, so one payload attribute per enrollment model instead of one per model *and* a bare `enrollment` key that only a re-enrollment session uses. And it removes the `extra_dict` argument, and with it the merge into the session dict, where a key could silently overwrite the session's own `pk`.

The session keeps its own dict: its `status` is the status of the session, not the outcome of the request that the top-level `status` reports.

The enrollments are referenced **by their keys only**. These events are posted on every enrollment attempt, and the full serialization carries the realm, the timestamps and the metadata of the enrollment secret — the quota, the expiry, the restrictions — which is a lot to repeat, and a couple of extra queries, for something the enrollment pk already identifies. The secret itself is never serialized either way.

The re-enrollment sessions were the only ones fetched without their enrollment in the MDM view, so they get the `select_related` the other three already had. Without it, reporting the enrollment would have cost one query per MDM request of every re-enrolled device, on the checkin and the command result paths.

## Two things that came with it

**The OTA enrollment events report their session.** The OTA view was the only one not doing so — it posted the phase alone, so an `ota_enrollment_request` event never said which enrollment the device was enrolling with. Both phases have the session in hand where the event is posted.

**The enrollment request events are linked to their enrollment.** They carried no linked object, so nothing pointed at the enrollment they now report and the events of an enrollment could not be fetched from a store. They are linked under `mdm_dep_enrollment`, `mdm_ota_enrollment` and `mdm_user_enrollment`, the object keys the audit events of the enrollments already use, so both kinds of events answer to the same key.

The `mdm_request` events are left out for now: they are posted on every checkin and every command result, and no view fetches the events of an enrollment yet.

## Tests

Three existing assertions were pinning the truncated payload and are updated. Added: a re-enrollment assertion for each of the three enrollment types, which is what pins the derived key; an assertion on the payload of a real `dep_enrollment_request` event and of a real `ota_enrollment_request` event; and the linked objects of the three event classes, with and without an enrollment in the payload.

Patch coverage is 100% (23 of 23 changed lines). The 3 "coverage regressions" Coveralls reports are an artifact of comparing line numbers across the diff: this branch deletes four small `serialize_for_event` overrides, so the lines below each deletion shift up by three, and `OTAEnrollmentSession.get_blueprint` — untouched here, and uncovered on main too — lands on a line number that used to hold a covered one.

### DIFF
--- a/tests/mdm/test_dep_enrollment.py
+++ b/tests/mdm/test_dep_enrollment.py
@@ -61,7 +61,10 @@ class TestDEPEnrollment(TestCase):
         self.assertEqual(session.status, "STARTED")
         self.assertEqual(
             session.serialize_for_event(),
-            {"enrollment_session": {"pk": session.pk, "type": "dep", "status": "STARTED"}}
+            {"enrollment_session": {"pk": session.pk, "type": "dep", "status": "STARTED"},
+             "dep_enrollment": {"pk": enrollment.pk,
+                                "name": enrollment.name,
+                                "uuid": str(enrollment.uuid)}}
         )
 
     def test_dep_enrollment_session_no_acme_payload(self):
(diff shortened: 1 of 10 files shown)


sebastian-zentral (dismissed):
lgtm

sebastian-zentral (approved):

PR #1614, opened 2026-08-17
Fix the wrong permission names in the templates

sebastian-zentral:
13 templates referenced permissions that no role can hold: 22 `perms.…` references, three of them
in dead code.

These fail silently, and only for the users who are supposed to see the button:
`ZentralBackend.has_perm` returns `True` for any string when the user is a superuser, and
`engine.has_legacy_perm` returns `False` for anything that is not registered in
`legacy_perm_actions`. The buttons were therefore there when testing as an admin, and missing for
every role scoped user.

Three flavours, all present:

* **another model** — `realms.delete_realmgroupmapping` guarding the realm group DELETE,
  `mdm.add_scepconfig` guarding the group → tag mapping CREATE (`scepconfig` is not even in the
  `permission_models` of the mdm app, so it cannot be granted at all)
* **typo, or the wrong app label** — `view_machinsnapshot`, `automatictableconstrucion`,
  `delete_script_check`, `update_jmespathcheck`, `perms.mbu.…`, and `perms.accounts.view_group`
  where the roles are `auth` groups
* **no app label at all** — `perms.add_pkginfo`, `perms.view_scriptcheck`, … which Django resolves
  as module checks (`has_module_perms("add_pkginfo")`), always false

## Fixed

| template | was | is | hidden button |
|---|---|---|---|
| `templates/user_menu.html` | `accounts.view_group` | `auth.view_group` | Roles menu entry |
| `realms/realmgroup_list.html` | `realms.delete_realmgroupmapping` | `realms.delete_realmgroup` | DELETE |
| `inventory/compliancecheck_detail.html` | `inventory.update_jmespathcheck` | `inventory.change_jmespathcheck` | UPDATE |
| `inventory/mbu_list.html` | `mbu.change_metabusinessunit` | `inventory.change_metabusinessunit` | UPDATE |
| `mdm/enrolleddevice_list.html` | `inventory.view_machinsnapshot` | `inventory.view_machinesnapshot` | machine link |
| `mdm/realmgrouptagmapping_list.html` | `mdm.add_scepconfig` | `mdm.add_realmgrouptagmapping` | CREATE |
| `mdm/realmgrouptagmapping_list.html` | `mdm.delete_realmgroupmapping` | `mdm.delete_realmgrouptagmapping` | DELETE |
| `monolith/pkginfo_list.html` | `add_pkginfo`, `add_pkginfoname` | `monolith.` prefixed | package info creation menu |
| `munki/index.html` | `view_configuration`, `view_scriptcheck` | `munki.` prefixed | Terraform export |
| `munki/scriptcheck_list.html` | `munki.delete_script_check` | `munki.delete_scriptcheck` | DELETE, for a delete only role |
| `osquery/automatictableconstruction_list.html` | `osquery.change_automatictableconstrucion` | `osquery.change_automatictableconstruction` | UPDATE column |
| `osquery/configuration_detail.html` | `perms.can_add_configuration_pack` | `can_add_configuration_pack` | CREATE pack |
| `stores/store_detail.html` | `accounts.view_group` | `auth.view_group` | role links |

`can_add_configuration_pack` is a context variable set by `ConfigurationView`, and it already
checks `osquery.change_configuration`, so it only lost its `perms.` prefix.

Two extras:

* `stores/index.html` and `stores/store_detail.html` had three empty `{% if %}` blocks left over
  from buttons that no longer exist — there is no store creation, update or deletion view in the
  UI — so they are gone.
* the empty group → tag mapping list called itself "Group → Tags".

## Tests

Every fix comes with a shown-with-permission / hidden-without pair in the app's existing test
module, and each of them fails without its template change.

The last commit adds `tests/server_base/test_template_perms.py`: it walks the templates of every
installed app plus the `TEMPLATES` dirs and asserts that each `perms.…` reference is registered in
the pbac engine, `legacy_perm_actions` for `perms.<app>.<codename>` and
`module_legacy_perm_actions` for the module wide `perms.<app>` form. Run against the tree before
this branch it reports all 21 bad references with their file and line. It cannot catch a
permission that exists but belongs to the wrong model — the first flavour above — so it is a net,
not a proof.

### DIFF
--- a/server/realms/templates/realms/realmgroup_list.html
+++ b/server/realms/templates/realms/realmgroup_list.html
@@ -59,7 +59,7 @@
           {% url 'realms:update_group' obj.pk as url %}
           {% button 'UPDATE' url "Edit group" %}
         {% endif %}
-        {% if perms.realms.delete_realmgroupmapping and obj.can_be_deleted %}
+        {% if perms.realms.delete_realmgroup and obj.can_be_deleted %}
           {% url 'realms:delete_group' obj.pk as url %}
           {% button 'DELETE' url "Delete group" %}
         {% endif %}
--- a/server/templates/user_menu.html
+++ b/server/templates/user_menu.html
@@ -45,7 +45,7 @@
             {% if perms.accounts.view_user %}
             <li><a class="dropdown-item" href="{% url 'accounts:users' %}">Users</a></li>
             {% endif %}
-            {% if perms.accounts.view_group %}
+            {% if perms.auth.view_group %}
             <li><a class="dropdown-item" href="{% url 'accounts:groups' %}">Roles</a></li>
             {% endif %}
             {% if perms.accounts.view_policy %}
--- a/tests/core_stores/test_store_views.py
+++ b/tests/core_stores/test_store_views.py
@@ -79,6 +79,25 @@ class StoreViewsTestCase(TestCase, LoginCase):
         self.assertContains(response, '<span class="store-backend">HTTP</span>')
         self.assertContains(response, "endpoint_url")
 
+    def test_store_authorized_role_link(self):
+        store = force_store()
+        store.instance.events_url_authorized_roles.add(self.group)
+        self.login("stores.view_store", "auth.view_group")
+        response = self.client.get(reverse("stores:store", args=(store.pk,)))
+        self.assertEqual(response.status_code, 200)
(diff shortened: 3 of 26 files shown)


np5 (changes requested):
Tests for this maybe?

np5 (approved):
LGTM

soysauceconspiracy (approved):
LGTM

PR #1624, opened 2026-08-20
Untangle the api.fqdn / api.tls_hostname accessors

np5:
## What this is

`docs/configuration/api.md` marks `api.tls_hostname` as deprecated in favour of `api.fqdn`, yet the payload and URL builders still read it. **This was not a live bug** — `APIDict` in `zentral/conf/__init__.py` derives each form from the other at config-load time, so both keys always resolve.

The real problem was that two similarly named accessors returned **different shapes**:

| accessor | returns |
|---|---|
| `settings["api"]["tls_hostname"]` | `https://zentral.example.com` (URL, with scheme) |
| `get_tls_hostname()` in `zentral/utils/osx_package.py` | `zentral.example.com` (bare FQDN — it reads `api.fqdn`) |

A reader who assumed those agreed got a doubled or a missing scheme. `zentral/contrib/mdm/payloads.py` and `server/accounts/views/user.py` each used both within a few lines of each other. Swapping the deprecated key for `api.fqdn` without prepending `https://` would have produced a scheme-less `BaseURL`, so this needed untangling rather than a find-and-replace.

## What to use

| You need | Value | Before | From now on |
|---|---|---|---|
| FQDN | `zentral.example.com` | `settings["api"]["fqdn"]` | `settings["api"]["fqdn"]` (unchanged) |
| FQDN, mTLS | `zentral-mtls.example.com` | `settings["api"]["fqdn_mtls"]` | `settings["api"]["fqdn_mtls"]` (unchanged) |
| Base URL | `https://zentral.example.com` | `settings["api"]["tls_hostname"]` (deprecated key) | `api_base_url()` |
| Base URL, mTLS | `https://zentral-mtls.example.com` | `settings["api"]["tls_hostname_for_client_cert_auth"]` (deprecated key) | `api_base_url(for_client_cert_auth=True)` |

`api_base_url` comes from `zentral.conf`. The two FQDN keys do not change: read them directly.

Package builders have their own FQDN accessor, because they already hold a builder instance. It is renamed:

| You need | Before | From now on |
|---|---|---|
| FQDN, in a builder | `self.get_tls_hostname()` | `self.get_api_fqdn()` |
| FQDN, mTLS, in a builder | `self.get_tls_hostname(for_client_cert_auth=True)` | `self.get_api_fqdn(for_client_cert_auth=True)` |

Do not import `get_api_fqdn` from `zentral.utils.osx_package` outside a builder. That module loads the package build machinery. Use `settings["api"]["fqdn"]`.

## The commits

**1. `santa: fix the client certificate auth configuration check`** — the one behaviour change, so it leads and can be reviewed alone. `ConfigurationForm.clean()` added its "mTLS not configured" error to `client_auth_certificate_issuer_cn`, a field removed from the model in `0031_auto_20220812_0925` (2022). `add_error()` raises `ValueError` for an unknown field, so ticking **Client certificate authentication** on a server with no mTLS FQDN returned a 500 instead of the validation error. The companion check on that field could never fire either: it was never in `cleaned_data`. The presence check also moves from the deprecated `api.tls_hostname_for_client_cert_auth` to `api.fqdn_mtls`. This commit carries the CHANGELOG entry.

**2. `conf: add api_base_url() and stop reading the deprecated api.tls_hostname key`** — the helper returns `https://{fqdn}` from the non-deprecated key. It lives in `zentral/conf` rather than next to `get_tls_hostname`, because `zentral/utils/osx_package.py` pulls in `subprocess`/`tempfile`/`shutil` that `accounts`, `realms` and `mdm` have no reason to import.

Migrated to it: turbo `BaseURL`, santa `SyncBaseURL`, the DEP profile `url` and `configuration_web_url`, the OTA enroll URL and profile-service `URL`, the password reset URL, the WebAuthn `expected_origin`, `MetaMachine.get_url()`, and the SAML and OIDC realm backend URLs. After this commit no application code reads a deprecated `api.tls_hostname*` key.

Two incidental cleanups fell out. The TOTP provisioning label went through `urlparse(...).netloc` to get the host back out of the URL, and reads `api.fqdn` directly now. The `.rstrip("/")` calls disappear, because a netloc never carries a trailing slash. The dead module-level `settings` import in the SAML backend goes too: a local in `get_saml2_config` shadowed it.

**3. `utils: rename get_tls_hostname() to get_api_fqdn()`** — the accessor says what it returns now. osquery's `--tls_hostname` flag and the `%TLS_HOSTNAME%` placeholders keep their names, because that is the agent's vocabulary and it does take a host. Every remaining `tls_hostname` in application code is osquery's.

**4. `conf: build the remaining base URLs with api_base_url()`** — every site that wrote `https://` in front of `api.fqdn` by hand now calls the helper. Those sites were correct and none of them read the deprecated key, but they left the shape of the value as a property of the call site instead of the accessor. The MDM `ServerURL` and `CheckInURL` selected `fqdn` or `fqdn_mtls` into an `fqdn_key` variable, which is what `api_base_url(for_client_cert_auth=...)` expresses.

Three sites keep their manual prefix on purpose. The jamf webhook and the Apps and Books notification URL read `api.webhook_fqdn`, a separate key that can point somewhere else, so `api_base_url()` would be wrong. The munki package builder already holds the bare FQDN for the `%TLS_HOSTNAME%` placeholder and the enrollment plist, and two accessors in one function would read worse than the prefix.

**The conf-layer shim is untouched.** It is what keeps deployments configured with `api.tls_hostname` working, and `test_api_base_url` pins `api_base_url()` to the value the deprecated key yields, so the two cannot drift.

## Tests

New tests: `api_base_url()` and its mTLS variant, including that each matches the deprecated key it replaces; the turbo `BaseURL` in the plist and in the configuration profile, asserted against `https://{fqdn}` instead of the previous `assertIn("BaseURL", ...)`, which is the scheme regression this guards; the santa `SyncBaseURL` on the client certificate path, which had no coverage; `MetaMachine.get_url()` and its missing-FQDN warning; `OTAEnrollment.get_enroll_full_url()` with and without a realm; the realm `scim_root_url`; and both branches of the santa client certificate check.

The full suite passes (8315 tests), `ruff` is clean, and every edited line is covered. The coverage was verified by intersecting the `coverage json` missing_lines with the `git diff -U0` hunks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -167,6 +167,8 @@ The Santa sync operations of a machine are serialized on its enrolled machine ro
 
 Fixed the Santa preflight that overwrote what another request committed on the same enrolled machine: it saved every column of the row it read. Only the attributes the client reports are written now.
 
+Fixed the 500 error when a user saves a Santa configuration with client certificate authentication on a server with no mTLS FQDN. The form put its error on a field removed from the configuration, and Django refuses an error on a field that does not exist.
+
 The inventory history cleanup deletes in bounded batches now. One unbounded statement for each table could delete millions of rows in one transaction, and keep the database at its maximum capacity for the full run. Each batch is a transaction, and a batch that gives an integrity error is retried alone.
 
 Fixed the MDM devices list that moved devices between its pages: it sorted on the last seen timestamp, which is null until a device checks in. The sort includes the primary key now.
--- a/ee/server/realms/backends/openidc/__init__.py
+++ b/ee/server/realms/backends/openidc/__init__.py
@@ -1,6 +1,6 @@
 import logging
 from django.urls import reverse
-from zentral.conf import settings
+from zentral.conf import api_base_url
 from realms.backends.base import BaseBackend
 from realms.exceptions import RealmUserError
 from zentral.utils.json import remove_null_character
@@ -18,11 +18,10 @@ class OpenIDConnectRealmBackend(BaseBackend):
         path = reverse("realms_public:openidc_ac_redirect", args=(self.instance.uuid,))
         if self.legacy_public_endpoints_mounted:
             path = path.removeprefix("/public")
-        return "{}{}".format(settings["api"]["tls_hostname"].rstrip("/"), path)
+        return f"{api_base_url()}{path}"
 
     def idp_initiated_login_uri(self):
-        return "{}{}".format(settings["api"]["tls_hostname"].rstrip("/"),
-                             reverse("realms_public:login", args=(self.instance.uuid,)))
+        return f'{api_base_url()}{reverse("realms_public:login", args=(self.instance.uuid,))}'
 
     def extra_attributes_for_display(self):
         config = self.instance.config
--- a/ee/server/realms/backends/saml/__init__.py
+++ b/ee/server/realms/backends/saml/__init__.py
@@ -9,7 +9,7 @@ from saml2.client import Saml2Client
 from saml2.config import Config as Saml2Config
 from saml2.saml import NAMEID_FORMAT_EMAILADDRESS
 
-from zentral.conf import settings
+from zentral.conf import api_base_url
 from zentral.utils.json import remove_null_character
 
 logger = logging.getLogger("zentral.realms.backends.saml")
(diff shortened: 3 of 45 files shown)


soysauceconspiracy (approved):
LGTM

PR #1627, opened 2026-08-20
santa: two machine rule ledger accounting fixes

np5:
Two fixes to the machine rule ledger, one commit each.

## A rule staged over a committed one is not counted

`committed_rules()` — the queryset the reported rule counts are compared with — left out a rule staged over a committed one, so a machine with a sync session open when it preflights compares one rule short of what it reports. `sync_ok()` then logs an error and writes `last_sync_ok = False`, and nothing can clear it: a rule update opens the gap, and it survives the session being settled either way.

Its `cursor__isnull=True` clause left out the staged removals and the replacements of the last batch of a session for the same reason. A cursor is only ever set together with a sync session, so the clause excluded nothing the session clause did not already exclude.

The gap has become visible: `2058d36a` removed the severity gate that kept the comparison from running on a stock configuration, and `1054ddc62` added sync health metrics that read its result. They currently report machines that are perfectly in sync as out of sync.

`MachineRule` now keeps the policy and the version the client confirmed when a send replaces them — `committed_policy` and `committed_version`, both null when the client has no other version of the rule, either because it confirmed this one or because it never received any. `committed_rules()` then counts the rule the client still holds.

The columns are added nullable and nothing backfills them, so the migration is safe to apply during a rolling deploy: a session already open when it runs is accounted for the way it is today, and the next one is correct.

A discarded session no longer deletes the row of a rule that replaced a committed one, it puts it back to the version the client holds. **A deleted row left the client with a rule the ledger did not know about anymore** — if the rule then went out of scope, no removal was ever sent for it and the client enforced it forever. That one is a rule distribution bug rather than a reporting one, and it has its own regression test.

A session the client never confirmed that only replaced rules is now discarded rather than committed. The two hypotheses have the same rule counts there, so they cannot be told apart, and committing assumes the client wrote rules it may never have received — which would leave it on the previous version of every one of them, with matching counts and nothing left to contradict the ledger.

## The removal of a rule staged during the same session

A `REMOVE` staged over a rule the session itself had staged carries no committed policy, which made it indistinguishable from the removal of a committed rule, so a discarded session restored it as a rule the client holds. The client never received it: the counts can no longer match, and if the rule is created again at the same version it is never sent — the ledger already has it — and the machine stays out of sync until a clean sync.

A staged removal now captures the committed policy and version like a staged replacement does, and a removal without them is for a rule of the same session: it is dropped with the discarded session. The removals staged before this change are marked by migration `0045` as covering a committed rule, which is what the previous code assumed, so the rolling deploy stays safe.

The two rollbacks are separate methods now, because they no longer treat that row the same way: `discard_session` deletes it with the rest of the session, but `_unstage` — the rollback of a batch the client never acknowledged — puts it back to the staged rule the removal covered, so that the removal is generated again. **The client received that rule with an acknowledged batch and holds it in its download buffer**: dropping the row would leave nothing to regenerate the removal from, and the client would apply a rule the ledger knows nothing about.

## Out of scope

A machine reporting *fewer* rules than the ledger claims it synced is a different problem: neither clause can make the ledger over-count, so that direction is drift that predates the comparison running on every preflight. Left alone here.

## Tests

For the first fix, five sync session tests driving real preflight / rule download / postflight cycles — an open session holding a replaced rule, an open session holding a staged removal, a genuine mismatch that must still be caught and logged, the interrupted session being discarded, and the removal of a rule whose session was discarded — plus the `discard_session()` and rule staging unit tests. Four of them fail on `main`.

For the second, the add-then-remove and the replace-then-remove sequences, each interrupted and discarded, and each completed and committed — the discarded add-then-remove one re-creates the rule and requires it to be sent again — plus the batch retry that must send the removal of a same-session rule again, and the `discard_session()` / `_unstage()` unit tests. The discarded add-then-remove sequence fails without the fix, with the exact signature seen in production.


soysauceconspiracy (approved):
LGTM

PR #1629, opened 2026-08-20
santa: serialize the sync operations of a machine on its row lock

np5:
The `select_for_update()` in `get_next_rule_batch` was dead code: the queryset was only ever consumed through `.update()` and `.delete()`, which never emit `FOR UPDATE`, and the batch query is raw SQL without one. Nothing serialized concurrent sync requests for one machine.

## The failure

Santa gives a rule download 30 seconds before it retries it with the same cursor, so a server transaction that stalls past that is overlapped by a duplicate of the request it is still serving. The retry cannot see the batch the original transaction has staged but not committed: its cleanup rolls back nothing, its acknowledgment update matches zero rows once the original's row locks are released, and its batch query — running after the original committed — treats the staged batch as already sent and returns the next one. The client's download buffer never holds the skipped batch, but the postflight commits the whole session, so the ledger permanently counts up to `batch_size` rules the client never received. The machine then reports fewer rules than the ledger on every preflight, and nothing short of a clean sync converges it. This is one concrete mechanism behind the over-counting direction #1627 left out of scope.

## The fix

New `EnrolledMachine.acquire_sync_lock()` takes `select_for_update()` on the machine's own row — never more than that one row — and `ATOMIC_REQUESTS` holds the lock until the end of the request. It is taken at the top of `get_next_rule_batch`, `commit_session` and `discard_session`, and in `reconcile_sync_session` once it holds a session to settle. The common preflight path, with no dangling session, stays lock free, and a test pins that.

The rule download also adopts the sync session found on the locked row: the wait for the lock can outlast a concurrent preflight, and staging under the session loaded with the request would put rules under a session no postflight can commit. `commit_session` deliberately does not adopt — the client's postflight confirms the session it downloaded, not one a concurrent preflight just started.

## Tests

A `TransactionTestCase` replays the overlap with two threads: the original request stalls after staging the second batch, and the retry must wait on the row lock, receive the second batch again, and hand the client every rule exactly once across the batches it kept. It fails without the lock — the retry skips to the third batch. The lock emission of each entry point, the lock-free preflight path and the session adoption are pinned by unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ Fixed Santa machines being reported out of sync when a rule sent during the curr
 Fixed a Santa rule that was added and removed during the same sync session being recorded as held by the machine when the session was lost: the discarded session turned the staged removal into a rule the client never received. If the rule was then created again, it matched the phantom ledger entry and was never sent, and the machine stayed out of sync until a clean sync. The removal of a rule staged during the same session is now dropped with the session it belongs to.
 
 
+The Santa sync operations of a machine are serialized on its enrolled machine row lock now. Santa retries a rule download after 30 seconds with the same cursor, and the retry could overlap the transaction of the original request, miss the batch it was staging, and skip it for good: the ledger counted rules the client never received. The rule downloads also pick up the sync session started by a concurrent preflight, instead of staging rules under a stale one.
+
+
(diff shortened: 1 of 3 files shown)


soysauceconspiracy (approved):
LGTM

PR #1637, opened 2026-08-21
mdm: keep the declarations when no software update is available to enforce

np5:
## The problem

A software update enforcement in latest mode raised a `DeclarationError` if the server found no software update for the device. The paths that build the declaration manifest do not catch this exception. Therefore the server sent a 500 response.

These are the results for a device that runs an OS version above the maximum target OS version:

| Request | Before | After |
|---|---|---|
| `PUT /checkin`, endpoint `tokens` | 500 | 200 |
| `PUT /checkin`, endpoint `declaration-items` | 500 | 200 |
| `PUT /connect` | 500 | 200 |

The 500 response on `/connect` is the most severe result. The function `_trigger_declarative_management_sync` is the fourth step in the chain of `get_next_command_response`. The exception occurred before these steps:

- the installation of the artifacts
- FileVault
- the recovery password
- the DEP account configuration
- the extra inventory

Thus one enforcement that the server cannot resolve removed the device from management.

Two conditions cause the problem. Both conditions are common:

1. The OS version of the device is above the maximum target OS version. A user updated the device, or the device came with a higher version. For example, the maximum target OS version is `26`, and the device runs `26.3`.
2. The table `SoftwareUpdate` is empty. This condition removes all the devices in scope at the same time. The query for a candidate update has the filter `availability__contains=today`. The function `sync_software_updates` deletes the records that are no longer in the feed. Zentral does not supply a schedule for this function.

## The correction

The function `build_specific_software_update_enforcement` now returns `None`. It does not raise an exception. The activation includes the configuration only if this function returns a declaration.

The second part is necessary. Before the correction, the activation made sure only that an enforcement was in scope. It did not make sure that the server found an update. If you remove the configuration from the declaration items only, the activation keeps a `StandardConfigurations` entry that has no declaration item. The device then rejects the complete activation and all of its configurations.

The activation and the declaration items now read the same `Target.software_update_enforcement_declaration`. This also decreases the number of queries of the feed to one for each request.

The parameter `missing_ok` has no more callers, thus it is removed. The new function `get_specific_software_update_enforcement` keeps the two event reasons for the fetch path. Thus `build_declaration_response` remains a simple dispatcher, as it is for all the other declaration types.

## The conditions that give no declaration

| Condition | Log level |
|---|---|
| The local time or the delay in days is absent | error |
| The device did not report a `SoftwareUpdateDeviceID` | info |
| No candidate update, and the feed is empty | error |
| No candidate update, and the feed is not empty | error |
| The best available update is below the OS version of the device | info |

The message for an empty feed is different from the message for a device without a candidate update. If the feed is empty, the server can enforce an update on no device. This is a problem of the deployment, not a problem of the device.

The two conditions at the info level are normal. The device is not yet in the inventory, or the device is up to date. An error message for these two conditions hides the real errors, because the server writes one message for each device in the fleet.

The check for an empty feed does not use the platform. The feed of Apple has no key for `iPadOS`. Therefore the field `SoftwareUpdate.platform` never has this value, and a filter on the platform shows an empty feed for all the iPads. The server does this check only after it finds no candidate update.

## The comparison of the OS versions

A device that runs the same version as the target keeps the declaration. The server removes the declaration only if the target is below the version of the device. The test `test_..._latest_same` makes sure of this behavior.

The comparison uses the new property `SoftwareUpdate.full_comparable_os_version`. It does not use `comparable_os_version`, because that property removes an empty `extra`. For a device that runs `26.3`, and for an update to `26.3` that has no `extra`:

| Value | Result |
|---|---|
| `EnrolledDevice.comparable_os_version` | `(26, 3, 0, "")` |
| `SoftwareUpdate.comparable_os_version` | `(26, 3, 0)` |
| `SoftwareUpdate.full_comparable_os_version` | `(26, 3, 0, "")` |

A tuple of three elements compares lower than a tuple of four elements that has the same first three elements. Therefore an update that the device already runs looked older than the device.

The property `comparable_os_version` does not change. Two of its four callers need the tuple of three elements. One of these two callers is important. [dep.py:88](zentral/contrib/mdm/public_views/dep.py:88) sends a 403 response `com.apple.softwareupdate.required` if the update compares higher than the OS version that the device reported. If the tuple gets a fourth element, the server tells a device that is up to date to install an update, and the device cannot enroll. The function `best_available_software_updates` is not affected, because its `<=` operator gives the same result with both tuples.

To make the number of elements the same everywhere, you must also change `make_comparable_os_version` and the six comparisons with the value `(0, 0, 0)`. Three of these six comparisons are validators of the OS version. One comparison in munki changes an empty `max_os_version` from "no limit" to "removes all the items". This work is a different change.

## The API

The API now rejects a null value for `delay_days` and for `local_time` if `max_os_version` has a value. These two fields can be null, because a one time enforcement does not use them. But a latest enforcement that has no target date cannot build a declaration. Before the correction, the code raised a `TypeError` in `datetime.combine` on the same manifest paths. The form always required these two fields. Therefore only the API had the problem.

If you do not include the two fields, the behavior does not change. The server uses the defaults of the model. All the Zentral clients continue to work:

- The request structure of `goztl` has no `omitempty` option for these two pointers. Therefore `goztl` always sends the key, and it cannot remove one.
- The Terraform provider sends `14` and `09:30:00` if `max_os_version` has a value.

The acceptance tests of both clients pass without a change.

## Malformed client capabilities

The last commit corrects the same type of problem on the same paths. The server stores `client_capabilities` exactly as the device reported it in its status report. There is no validation. The function `build_target_management_status_subscriptions` caught only a `KeyError` for the lookup of the supported status items. But a value of an incorrect type raises a `TypeError`, and a status item that is not a string raises an `AttributeError`. Nothing catches these two exceptions while the server builds the manifest. Therefore `tokens`, `declaration-items` and `/connect` sent a 500 response, and the device received nothing.

These are the results before the correction:

| Value | Exception |
|---|---|
| `['supported-payloads']` | `TypeError: list indices must be integers` |
| `'nope'` | `TypeError: string indices must be integers` |
| `42` | `TypeError: 'int' object is not subscriptable` |

The function now uses the status items that all the clients support, as it did before for capabilities that have no `status-items` key. It also writes a warning. A value that is not an object has a different message, because the server can look up no key in it. The function `supports_software_update_enforcement_specific` has the check `isinstance(..., dict)`. This function was the only function without the check.

## The tests and the checks

- The complete test suite passes: 7958 tests.
- The patch coverage is 100% on all the changed lines. Method: get the intersection of the `missing_lines` of `coverage json` and the sections of `git diff -U0`.
- `mkdocs build --strict` passes.
- `ruff` finds no error in the changed files.
- Each commit passes `tests.mdm`. Therefore you can bisect the branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,15 @@ The data asset API only accepts XML property lists now, the way the upload form 
 A malformed base 64 source on the MDM profile and provisioning profile API endpoints is answered with a 400 now. Decoding it raised, nothing caught it, and the request ended in a 500.
 
 
+Fixed an MDM software update enforcement in latest mode taking a device out of management when it had no update to enforce on it. Finding none raised, and nothing caught that on the paths building the declaration manifest: the `tokens` and `declaration-items` check-ins and `/connect` all answered with a 500, and since the declarative management sync is scheduled before them, the device also stopped receiving its artifacts, its FileVault configuration, its recovery password and its extra inventory. A device whose OS was above the *Maximum target OS version* — because it was updated past it, or shipped above it — was enough to trigger it, and so was a deployment that had never synchronized the Apple software lookup service, which took out every device in scope at once. The configuration is left out of the declarations the device is served now, and Zentral logs an error, or an information for the devices that are simply already up to date or not inventoried yet. The device kept getting a no-op enforcement targeting the version it already ran, and it still does; only a target below what the device runs is dropped.
+
+
+Malformed MDM client capabilities no longer take a device out of management. They are stored exactly as the device reported them, so the lookup of the status items it supports could fail on a type rather than on a missing key, and that was not caught: the `tokens` and `declaration-items` check-ins and `/connect` answered with a 500, and the device stopped receiving anything at all. The status subscriptions fall back to the items supported by every client now, and Zentral logs a warning.
+
+
+An explicit null delay in days or local time is rejected on the MDM software update enforcement API when a maximum target OS version is set. Both fields are nullable for the one time enforcements, which have no use for them, and a latest enforcement without them has no target date to build a declaration with. Omitting them is unchanged and still falls back to 14 days and 09:30.
+
+
 ## 2026.5
 
 
--- a/docs/apps/mdm.md
+++ b/docs/apps/mdm.md
@@ -294,6 +294,8 @@ Zentral offers two variants for setting up a Software Update Enforcement configu
 
   This type automatically enforces the latest available OS version up to a **Maximum target OS version**. Zentral will use the *device identifier* and match information from the *Apple Software Lookup Service* to return the latest OS version (e.g., `16` to install all macOS 15 updates on your fleet, but to stop before installing 16). Set the **Delay in days** following the software release and a **Target local time** to configure the enforcement time (e.g.`7` for 7 days and `09:30:00` for 9:30 a.m.).
 
+    A device only receives the enforcement when Zentral finds an update for it. A device already running a version *newer* than the one Zentral picked - because it is past the *Maximum target OS version*, or because it shipped above it - is left out, and is enforced again as soon as a newer update below the maximum is published. This mode depends on the software updates being synchronized from the *Apple Software Lookup Service*: while none are known, no device can be enforced, and Zentral logs an error for each of them.
+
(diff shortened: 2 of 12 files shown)


soysauceconspiracy (approved):
LGTM

PR #1640, opened 2026-08-22
Do not put the error from S3 in the enterprise app API response

np5:
The enterprise app API is the last of the three callers of `download_external_resource` that sent everything boto3 raises to the client in `str(e)`. Those messages can contain the bucket, the key, the region or the state of the credentials. The package API and the data asset API already answer in the correct way, so this makes the three of them the same.

```python
except ValueError as e:
    # ValueError carries operator-facing validation messages
    # (unknown URI scheme, unsupported extension, hash mismatch).
    raise serializers.ValidationError({"package_uri": str(e)})
except Exception:
    # Surface a generic message; the actual exception may carry
    # internal state (boto3 traceback, file paths, …). The full
    # exception is logged server-side for diagnosis.
    logger.exception("Could not download or validate enterprise app from %s", package_uri)
    raise serializers.ValidationError(
        {"package_uri": "Could not download or validate the enterprise app."}
    )
```

`read_package_info` runs in the same `try`, and it raises a `ValueError` for its own message about the file extension. That message still reaches the client. The tests that check the messages about the URI scheme, the file extension and the hash do not change.

### The test

The test used a `ValueError` for the error from S3, but no error from boto3 is a `ValueError`, so it did not exercise the branch it looked like it exercised. It raises a plain exception now, and it makes sure that:

- the client gets the generic message,
- `HeadObject` is not in the body of the answer,
- the message from S3 is in the log.

### Tests

The full test suite, as CI does it: **7977 tests, OK**. All the added lines in `zentral/` have test coverage: 11 added lines, and 0 lines without coverage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,7 +57,7 @@ The MDM data asset API accepts only XML property lists now, like the upload form
 
 A malformed base 64 source on the MDM profile and provisioning profile API endpoints gives a 400 now, and not a 500.
 
-The MDM data asset API gives a generic message when it cannot download the file now, like the package API already did. The message of an error from S3 can contain the bucket, the key, the region or the state of the credentials. Zentral writes it to its logs instead. The messages about the file extension, the URI scheme and the hash are unchanged.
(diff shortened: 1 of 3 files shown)


soysauceconspiracy (approved):
LGTM

PR #1642, opened 2026-08-23
Publish audit events from the Osquery management views

np5:
Osquery was the last large module with no `zentral_audit` event. Creating, updating or deleting a configuration, a pack, a query, an automatic table construction, a file category, an enrollment or a distributed query run from the web console left no record. No test in the module asserted a single event.

The Zentral Terraform provider manages seven of these objects: the automatic table constructions, the configurations, the configuration packs, the enrollments, the file categories, the packs and the queries. They are managed as code, so the event that earns its place is the one for a change that did not come from the Terraform configuration. Distributed query runs are the exception, and section *Decisions* explains why they are audited too.

This is the first of four changes. It covers the models and the management views. The API views, the standard pack import and the missing endpoint for the runs come later.

## Commits

**1. `inventory:` put the version and the update time in the enrollment event payload**

`BaseEnrollment.serialize_for_event()` left the version out, so the audit event of a version bump showed the same value before and after: an event that recorded no change. The Turbo view that only increases the version published one of those on every use. The fix is in the base method, so the five enrollments built on `BaseEnrollment` — Monolith, Munki, Osquery, Santa and Turbo — report the change the same way.

The `enrollment` attribute of the `turbo_request` and `turbo_result` events does not change: the Turbo enrollment answers `keys_only` requests before it calls the base method. The MDM enrollments are separate models.

**2. `osquery:` fix the query payload of the events**

`Query.serialize_for_event()` wrote the version of the query under the `version` key, then replaced it with the minimum Osquery version when the query had one. The two values are unrelated. A consumer of the `osquery_pack_query_update` events read the wrong number for every query that restricts its Osquery version. The payload also had no `pk`, no `name`, no tag and no compliance check, so an event could not name the query that changed.

**3. `osquery:` publish audit events from the management views**

Twenty-four views move to the `CreateViewWithAudit` / `UpdateViewWithAudit` / `DeleteViewWithAudit` bases. The two enrollment views that drive two forms or implement `post()` themselves call `post_audit_event()`. Five models had no `serialize_for_event()` at all, and `AuditEvent.build()` calls it with no fallback, so they come first.

## Two defects found while the events were connected

**"Halt current runs" ended other runs with no record.** The option is on by default. It closed every other active run of the query with a queryset update: no `save()`, no signal, no instance, and so no event from a view that only knows its own object. Creating one run stopped three others and the trail recorded one change. `DistributedQuery.objects.halt_for_query()` reads the rows before the update and reports them, so each stopped run gets its own event.

**`QueryForm` put a database expression in the payload.** It increases the version with an `F()` expression when the SQL changes, and never refreshed the instance. `sync_query_compliance_check()` refreshes it, but only when the compliance check is on, so every other update carried a `CombinedExpression` where a number belongs.

## Decisions

Distributed query runs are audited although the Terraform provider does not manage them. A run executes operator SQL on the machines, which is the most audit-worthy action in the module. The `osquery_distributed_query` resource does not exist, and this change does not ask for one: a run is an action, and not a declared state.

Two effects of a change stay silent, and the documentation says so: a configuration change increases the version of each of its enrollments, and a pack query change sets the update time of its pack. They are consequences of a change that already has an event. Stopping a run is different — it changes another object that the operator did not name — so it gets its own event.

The pack upload view keeps the `osquery_pack_update` events for now. It goes through `update_or_create_pack`, which needs its own change.

## Tests

7977 tests pass. Patch coverage is 100%, verified line by line against the diff. Each of the three commits passes the full suite on its own.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,11 @@ The enrolled machine and sync state metrics are bucketed by the age of the last 
 ### Backward incompatibilities
 
 
+#### 🧨 Osquery query payload in events
+
+The `version` attribute of an Osquery query in an event payload is the version of the query again. It was the minimum Osquery version of the query, which replaced the query version when the query had one. The minimum Osquery version has its own `minimum_osquery_version` attribute now. The query payload also has the `pk`, `name` and `tag` attributes, and the compliance check of the query. This payload is a part of the `osquery_pack_query_update` events.
+
+
 #### 🧨 Santa enrolled machines metrics
 
 `zentral_santa_enrolled_machines_total` and `zentral_santa_enrolled_machines_sync_total` are replaced by `zentral_santa_enrolled_machines_bucket` and `zentral_santa_enrolled_machines_sync_bucket`, which have an `le` label. The `le="+Inf"` bucket gives the value of the two removed gauges: in a dashboard or an alert, change the metric name and select that bucket. A query that sums the buckets counts each machine seven times.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ A data asset can be created and updated with its content in the request now, bas
 The `file_uri` and `file_sha256` attributes of the data asset endpoints are not required fields anymore. A request without `file_uri` and without `source` gives one error for the two of them, and not one "This field is required." for each attribute.
 
 
+#### Osquery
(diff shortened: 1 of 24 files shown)


soysauceconspiracy (approved):
LGTM

PR #1644, opened 2026-08-24
Publish audit events from the Osquery API views

np5:
The web console records every change to an Osquery object since [#1642](https://github.com/zentralopensource/zentral/pull/1642). The API changed the same objects with no record of it, so an operator could not tell which of the two made a change, or that a change came from a service account at all.

This is the second of four changes. The models and the management views came first. The standard pack import and the missing endpoint for the runs come later.

## Commits

**1. `osquery:` publish audit events from the API views**

The seven list views move to `ListCreateAPIViewWithAudit`, and the seven detail views to `RetrieveUpdateDestroyAPIViewWithAudit`. The bases set the permission classes and the filter backend, so those lines come out of each view. The two views that refuse to delete an object still in use keep their `perform_destroy()`: the audited base deletes and then publishes, in that order.

**2. `osquery:` paginate the API list endpoints**

The seven list endpoints answered with the whole table. They answer with a page now: 50 objects by default, 500 at most.

## PATCH is a 405

The audited base only accepts the full update. This is a fix and not only a rule, because the serializers assume every declared field is there:

| Endpoint | What a PATCH did |
|---|---|
| `/queries/<pk>/` | `KeyError` on `compliance_check_enabled` → **500** |
| `/enrollments/<pk>/` | `KeyError` on `secret` → **500** |
| `/file_categories/<pk>/` | slug computed from a name that is absent → **the object was renamed to `none`** |
| `/packs/<pk>/` | the same |

`slugify(None)` gives `"none"`, because `slugify()` calls `str()` on its argument first. A `PATCH` that carried only a description therefore renamed the object, or was refused because another object already had the slug `none`. No test in the suite sent a `PATCH`.

## Orderings

Pagination needs a total ordering. An ordering on a column that is not unique lets a row move between two pages, and the client reads it twice or not at all.

| Endpoint | Ordering | Total because |
|---|---|---|
| ATCs, configurations, file categories, packs, queries | `name` | the name is unique on those models |
| Enrollments, configuration packs | `pk` | those models have no name |

## Effect on the open pull requests

The orderings and the list assertions are the ones of [#1628](https://github.com/zentralopensource/zentral/pull/1628), taken from that branch. Its Osquery part is a deletion now, and not a merge. The Osquery part of [#1641](https://github.com/zentralopensource/zentral/pull/1641) is a deletion as well: the audited base sets the same method list.

## Tests

7985 tests pass. Patch coverage is 100%. The 21 create, update and delete tests assert the event, and one test walks the seven detail endpoints to check the 405. The first commit passes the full suite on its own.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ Zentral publishes a `zentral_audit` event when a user creates, updates or delete
 
 A run that stops other runs of the same query — the "Halt current runs" option — now publishes one event for each stopped run. Before, the other runs ended with no record of it.
 
+The Osquery API publishes the same `zentral_audit` events as the web console now. The nine endpoints of the automatic table constructions, the configurations, the configuration packs, the enrollments, the file categories, the packs and the queries were the last ones that changed an object with no record of it.
+
 
 #### Santa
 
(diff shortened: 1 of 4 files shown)


soysauceconspiracy (approved):
LGTM

PR #1645, opened 2026-08-24
Split the Osquery pack import events

np5:
An import of a standard pack changes a pack, its pack queries, the queries they schedule and their compliance checks, in one request. It reported that with two event types of its own, written two years before `AuditEvent` existed, under a comment that called them the audit trail.

This is the third of four changes. The models, the management views and the API views came first. The missing endpoint for the runs comes last.

## The two events did two jobs

**A job only they can do.** The report of the import: the result as a whole, including `present`, which says that the import ran and changed nothing, and the counts of the pack queries created, updated, deleted and left alone. `AuditEvent` has no action for a change that did not happen, and it describes one object at a time.

**A job `AuditEvent` does better.** The record of each object. The removed events carried a partial difference instead of the two values, they never declared their linked objects, so they did not reach the page of the object they reported, and `osquery_pack_query_update` reported a pack query and the query it schedules together, with the SQL of the query under a `query` key.

So the report keeps the first job, and audit events take the second. `osquery_pack_query_update` is removed. An import publishes one report and one audit event per changed object, under one event UUID, the report first, so a reader can gather the whole import from any one of its events.

The upload in the web console goes through the same code, so it publishes the same events. It was the last way to change a pack with no audit trail.

## The response does not change

`PUT /api/osquery/packs/<slug>/` still answers with `result`, `query_results`, `pack` and `updates`. `updates` comes out of the *event* only — the audit event of the pack carries the whole value before and after, which says more.

## A write that should not have happened

The import adopts a query that already exists when a pack query is missing. It saved that query whatever the file said, because the flag that recorded the change was set before the comparison and not inside it:

```python
for attr, new_val in query_defaults.items():
    query_updated = True          # before the comparison
    old_val = getattr(query, attr)
    if old_val != new_val:
        ...
```

That write was invisible. As an audit event it is a change whose value before and after are the same, published on every import. `test_import_pack_adopting_an_unchanged_query_does_not_report_a_change` covers it, and fails when the flag is put back where it was.

## One more thing

The first commit fixes a defect from [#1642](https://github.com/zentralopensource/zentral/pull/1642). The compliance check status events have always linked a distributed query under `osquery_run`. The audit events of the runs used `osquery_distributed_query`, the key built from the verbose name, so the two halves of the history of a run never met. The model declares `linked_objects_key` now — the attribute added in [#1643](https://github.com/zentralopensource/zentral/pull/1643) for exactly this. `osquery_run` wins because it is the older key, because the web console calls these objects runs, and because no released version wrote the other one.

Tell me if you want that commit in its own pull request.

## Tests

7994 tests pass. Patch coverage is 100%.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ Zentral publishes a `zentral_audit` event when a user creates, updates or delete
 
 A run that stops other runs of the same query — the "Halt current runs" option — now publishes one event for each stopped run. Before, the other runs ended with no record of it.
 
+An import of a standard Osquery pack publishes a `zentral_audit` event for each object it changes, next to the `osquery_pack_update` event that reports the import. The import changed packs, queries and pack queries with no audit trail, and it is the only way to change several of them at once.
+
 The Osquery API publishes the same `zentral_audit` events as the web console now. The nine endpoints of the automatic table constructions, the configurations, the configuration packs, the enrollments, the file categories, the packs and the queries were the last ones that changed an object with no record of it.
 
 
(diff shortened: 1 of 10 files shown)


soysauceconspiracy (approved):
LGTM

PR #1647, opened 2026-08-24
Publish audit events from the Monolith API views

np5:
The Monolith API changed a catalog, a condition, an enrollment or a manifest enrollment package with no record of it. The web console publishes a `zentral_audit` event for the same objects, so the API was the only way to change them without a trace. This is the same change as the Osquery API in #1644.

### The views

Eight views move to `ListCreateAPIViewWithAudit` and `RetrieveUpdateDestroyAPIViewWithAudit`:

| model | list | detail |
|---|---|---|
| Catalog | `/api/monolith/catalogs/` | `/api/monolith/catalogs/<pk>/` |
| Condition | `/api/monolith/conditions/` | `/api/monolith/conditions/<pk>/` |
| Enrollment | `/api/monolith/enrollments/` | `/api/monolith/enrollments/<pk>/` |
| ManifestEnrollmentPackage | `/api/monolith/manifest_enrollment_packages/` | `/api/monolith/manifest_enrollment_packages/<pk>/` |

The bases give the permission classes and the filter backend, so the views do not declare them again. The three `perform_destroy` methods that refuse to delete an object in use call `super()`, so their check runs first and the audit event follows.

### The delete of a manifest enrollment package

`ManifestEnrollmentPackageDetail.perform_destroy` calls `delete(delete_enrollment=False)`, so it could not use the `perform_destroy` of the base, and its delete had no audit event.

`RetrieveUpdateDestroyAPIViewWithAudit` calls `delete()` with the keyword arguments of the new `get_perform_destroy_kwargs()` now. The default is an empty dict, and `DestroyModelMixin.perform_destroy` calls `delete()` with no argument, so nothing changes for the other views. Two tests in `tests/utils/test_audit_views.py` keep this contract: the default is empty, and an override is used.

Why the argument is necessary, from `e4c85c5` (#974): the console creates the enrollment together with the package, and deletes the two of them together. The API takes an `enrollment_pk` for an enrollment that already exists, made by the Munki enrollments API or by Terraform, so it must only release it. The delete test asserts the audit event **and** that the Munki enrollment is still there, so a change to a plain `super()` call cannot pass.

### PATCH

These endpoints answer a `PATCH` with a `405` now, because the bases accept only the full updates. This is the same change as the Osquery API. The new `test_patch_is_not_allowed` covers the four detail endpoints.

### Tests

The full test suite, as CI does it: **8026 tests, OK**. All the added lines in `zentral/` have test coverage: 23 added lines, and 0 lines without coverage.

Twelve audit assertions were added to the tests: three for the catalogs, three for the conditions, three for the enrollments, and three for the manifest enrollment packages. To make sure that they can fail, one view was put back on its old base, and the test of the create of a catalog failed because there was no event.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ A sub manifest package can use more than one key, one time for each key, and eac
 
 Zentral adds a `default_installs` package to the `optional_installs` of the sub manifest, because munki installs it only if it is an optional install too. If the sub manifest has its own `optional_installs` package with the same name and the same condition, Zentral keeps that one, with its own scope.
 
+The Monolith API publishes the same `zentral_audit` events as the web console now. The catalogs, the conditions, the enrollments and the manifest enrollment packages changed with no record of it.
+
 
 #### Osquery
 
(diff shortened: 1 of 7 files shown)


soysauceconspiracy (approved):
LGTM

PR #1648, opened 2026-08-24
Publish audit events from the Munki and Santa enrollment API views

np5:
The Munki API and the Santa API created, updated and deleted an enrollment with no record of it. The web console publishes a `zentral_audit` event for the same object. This continues the work of #1644 for the Osquery API and #1647 for the Monolith API.

One commit for each module.

### The views

| module | list | detail |
|---|---|---|
| Munki | `/api/munki/enrollments/` | `/api/munki/enrollments/<pk>/` |
| Santa | `/api/santa/enrollments/` | `/api/santa/enrollments/<pk>/` |

The four views move to `ListCreateAPIViewWithAudit` and `RetrieveUpdateDestroyAPIViewWithAudit`, which give the permission classes and the filter backend. The Santa `perform_destroy` that refuses to delete an enrollment in use calls `super()`, so its check runs before the audit event.

`generics` is not used in the Munki API views anymore, so the import goes.

### PATCH

These endpoints answer a `PATCH` with a `405` now, because the bases accept only the full updates. This is the same change as the Osquery API and the Monolith API.

### The Santa rules keep their views

A rule change already publishes a `santa_rule_update` event, and a rule set publishes one for each rule it changes. An audit event there is a decision about one event or two, and not a new base class, so the rules are not part of this PR.

### Tests

The full test suite, as CI does it: **8026 tests, OK**. All the added lines in `zentral/` have test coverage: 4 added lines, and 0 lines without coverage.

Six audit assertions were added, three for each module. Each module has an `assert_audit_event` helper in its test utils, like the Monolith and the Osquery ones.

The test of the update also checks the version in the payload of the event, before and after. `BaseEnrollment.serialize_for_event()` puts the version there since #1642, and an update of an enrollment increases it. The helper compares the event with the same `serialize_for_event()` that the event used, so it cannot find a payload that lost an attribute. To make sure that this new check can fail, the version was taken out of the base method, and the two tests of the update gave a `KeyError`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,8 @@ The Osquery API publishes the same `zentral_audit` events as the web console now
 
 #### Santa
 
+The Santa API publishes the same `zentral_audit` events as the web console now, for the enrollments.
+
 A clean sync — `CLEAN` or `CLEAN_ALL` — can be queued for an enrolled machine from its machine page or from the API, and cancelled before its next preflight. The new `Santa::Action::"forceCleanSync"` PBAC action gives the permission, and accepts a meta business unit and a sync type.
 
 New `/api/santa/enrolled_machines/` endpoint.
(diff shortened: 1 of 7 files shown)


soysauceconspiracy (approved):
LGTM

PR #1649, opened 2026-08-24
Monolith: document the install and the uninstall of the same package

np5:
Two follow-ups to #1646.

## The install and the uninstall of the same package

`unique_together = (sub_manifest, pkg_info_name)` made it impossible to add a package with the `managed_installs` key and with the `managed_uninstalls` key in one sub-manifest. `UniqueConstraint(sub_manifest, pkg_info_name, key)` allows it now. Across two sub-manifests it was always possible.

The documentation said not to do it. It did not say that nothing prevents it, or what munki does. Munki keeps the install: `processRemoval()` reads `processed_installs`, and it does not remove a package that is in that list. It writes a warning instead.

There is no check in the form or in the API. A check can only cover one sub-manifest, and the same conflict is possible across two, so it gives partial protection that reads as complete.

The paragraph also held the catalog warnings, which are a different subject. "You also see warnings" read as a result of the conflict. They move to their own paragraph.

The scope list of the first paragraph left out the excluded tags, and it offered the featured flag for every key. The form accepts that flag only for the `default_installs` and the `optional_installs` keys.

## The shards of a managed uninstall

The `managed_uninstalls` key carries the excluded tags and the shards since #1646. A test covered the excluded tags in a served sub manifest, but nothing covered the shards. The two new tests match the ones of the `managed_updates` key: a machine out of the shard keeps the package, and a machine in the shard removes it.

## Tests

Full suite: 8028 tests, OK. `mkdocs build --strict` is clean.

### DIFF
--- a/docs/apps/monolith.md
+++ b/docs/apps/monolith.md
@@ -62,13 +62,15 @@ It's important to note that, unlike how tagging works for catalogs, enrollments 
 
 ### Add software via sub-manifests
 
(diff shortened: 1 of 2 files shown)


soysauceconspiracy (approved):
LGTM

PR #1650, opened 2026-08-24
Publish audit events from the inventory JMESPath check views

np5:
The inventory JMESPath compliance check had three event types of its own, where every other object of the console and of the API publishes a `zentral_audit` event. The Munki script checks and the Turbo MSCP checks, which are compliance checks too, already publish `zentral_audit` events.

Two commits.

### 1. The audit events of the compliance checks reach the check

An audit event of a Munki script check, a Turbo MSCP check, an Osquery query or a Turbo script that is a compliance check, or an inventory JMESPath check was linked to the check itself and not to its compliance check. A query for the events of a compliance check found only a part of them. The five models declare it in `linked_objects_keys_for_event()` now.

The Osquery query and the Turbo script declare it only when they have one, because they are compliance checks only when the user asks for it.

These are the five models that have a compliance check and views that publish an audit event. `MachineStatus` has one too, and it is the result of a check on a machine and not a check, so it keeps no link.

The test of the creation of a compliance check query published no event assertion, so the link of the Osquery query has a new one.

Two object keys were also corrected, and one of them was a real defect:

| model | key before | key now |
|---|---|---|
| `ComplianceCheck` | `compliance_checks_compliance_check` | `compliance_check` |
| `JMESPathCheck` | `jmes_path_check` | `inventory_jmespath_check` |

The events page of a JMESPath check asks for `inventory_jmespath_check`. Without this correction, the audit events of the check would go to a key that no page reads, and the page would show nothing, with no error.

**A behaviour that was called deliberate is reversed.** The Turbo test of a script with a tag and a compliance check said the audit event links the tag and "deliberately NOT the compliance check". That comment came with the Turbo test suite in `dd2e569f`, it gives no reason, and no other compliance check makes that difference. The five checks behave the same way now.

**A link that did the same work two times is removed.** The Turbo script also declared a link to itself. `AuditEvent.build()` adds the link to the object of the event when `linked_objects_keys_for_event()` gives no link to it, and the key that it computes for the model is `turbo_script` — the key that the line wrote. The Turbo tests that compare the full `objects` dictionary of the event show that the result is the same.

Nine other models declare a link to themselves in the same way, and they are not a part of this change: `mdm.DEPDevice`, `monolith.PkgInfoName`, `monolith.PkgInfo`, `monolith.SubManifestPkgInfo`, `osquery.ConfigurationPack`, `osquery.Enrollment`, `turbo.Enrollment`, `turbo.RecurringJob` and `turbo.OneTimeJob`. `mdm.DeviceCommand` also declares one, but that one is necessary: it links itself by uuid and not by pk, which is the reason why `build()` lets a model keep control of its own key.

### 2. The JMESPath check views publish audit events

The six views — three of the console, three of the API — publish `zentral_audit` events, and the three event types are removed.

The two views of the console drive two forms each, so they call `post_audit_event()`, like the Osquery enrollment views. On the update, the previous value is read in `dispatch()`: a `ModelForm` writes the posted data onto its instance while it validates, so a value read in `forms_valid()` is the new one. The update still publishes an event only when a form changed.

The delete of the API deleted the compliance check and let the cascade remove the JMESPath check. `JMESPathCheck.delete()` does that now, like `ScriptCheck.delete()` in Munki, so the views only delete the object and the shared bases publish the event.

### Tests

The events of the console views had no test at all. The three tests of the console check the audit event now:

- the create and the delete check that the event still reaches the compliance check,
- the delete checks that the compliance check goes with the JMESPath check,
- the update checks that the previous value is not the new one, which is the only way to find the `ModelForm` trap.

The full test suite, as CI does it: **8028 tests, OK**. All the added lines in `zentral/` have test coverage: 45 added lines, and 0 lines without coverage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 ### Features
 
 
+#### Inventory
+
+The inventory JMESPath compliance check publishes the same `zentral_audit` events as the other objects now, from the web console and from the API.
+
+
 #### MDM
 
 A data asset can be created and updated with its content in the request now, base 64 encoded in the new `source` attribute, and not only with a `file_uri` that points to an object in an S3 bucket. `file_uri` and `source` are mutually exclusive, and one of them is required. Zentral computes `file_sha256` from a `source`, and requires it only with a `file_uri`. A `file_sha256` given with a `source` is verified against the content. A data asset created from a `source` has no filename.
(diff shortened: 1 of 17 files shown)


soysauceconspiracy (approved):
LGTM

PR #1651, opened 2026-08-24
Publish audit events from the munki and santa console enrollment views

np5:
## What this does

The web console publishes a `zentral_audit` event for the enrollment changes that had no record. Munki and santa were the last two modules without them.

| module | view | before | now |
|---|---|---|---|
| munki | create enrollment | no event | `created` |
| munki | delete enrollment | no event | `deleted` |
| munki | bump enrollment version | no event | `updated` |
| santa | create enrollment | no event | `created` |

The other modules already do this: the six MDM enrollment views (DEP, OTA and user, create and update), osquery and turbo.

## How

The create views and the munki bump version view drive two forms, so they are plain `TemplateView`s and not the audited `CreateView`. They use `post_audit_event()`, like the osquery and MDM equivalents. The munki delete view moves from `DeleteView` to `DeleteViewWithAudit`.

The bump version view serializes the enrollment before the save, to give the `prev_value`.

## Why the API entries in the changelog were not the full story

#1648 gave the audit events to the munki and santa enrollment API endpoints. The changelog said that the API publishes "the same events as the web console". For the create, the delete and the version bump, the console published nothing, so there was nothing to be the same as. This change makes the sentence true.

## The munki enrollment gave no linked objects

`munki.Enrollment` declared no `linked_objects_keys_for_event()` at all, so its audit events were only on the page of the enrollment. The osquery, santa and turbo enrollments each give their configuration, so the events of an enrollment are on the page of the configuration too. The munki enrollment gives it now, with the key `munki_configuration` that the configuration events page reads.

It gives no link to itself. `AuditEvent.build()` adds the link to the object of the event when `linked_objects_keys_for_event()` gives no link to it, and the key that it computes, `munki_enrollment`, is correct. The osquery and turbo enrollments still write their own key, which does the same work two times; they are not a part of this change.

## Tests

The four tests that already exercised these views now assert the event: the action, the model, the pk, the payload and the tags.

Each test also asserts that the enrollment secret does not appear in the payload. `EnrollmentSecret.serialize_for_event()` leaves the secret out on purpose, because the secret is the credential that enrolls a machine: a regression there puts a live credential in each event store. The MDM enrollment audit tests make the same check.

- full suite: 8028 tests, OK (the count does not change, because the four tests already existed)
- the four tests compare the full `objects` dictionary of the event, and not only the payload
- `tests.munki` and `tests.santa`: 1111 tests, OK
- patch coverage: 11 added executable lines in `zentral/`, 0 uncovered

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,10 @@ The Monolith API publishes the same `zentral_audit` events as the web console no
 
 The Munki API publishes the same `zentral_audit` events as the web console now, for the enrollments.
 
+The Munki web console publishes a `zentral_audit` event when a user creates an enrollment, deletes an enrollment, or increases the version of an enrollment. These three changes had no record.
+
+The audit events of a Munki enrollment are also on the page of its configuration now, like the Osquery, Santa and Turbo enrollments. The Munki enrollment gave no link to its configuration, so its events were only on the enrollment.
+
 
 #### Osquery
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,8 @@ The Osquery API publishes the same `zentral_audit` events as the web console now
 
 The Santa API publishes the same `zentral_audit` events as the web console now, for the enrollments.
 
+The Santa web console publishes a `zentral_audit` event when a user creates an enrollment. This change had no record.
+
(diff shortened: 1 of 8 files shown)


soysauceconspiracy (approved):
LGTM

PR #1652, opened 2026-08-24
Remove the redundant self keys of the linked objects

np5:
## What this does

`AuditEvent.build()` links the object of the event when `linked_objects_keys_for_event()` gives no link to it. Nine models wrote their own key with their own pk, which is the value that `build()` computes, so the line did the same work a second time.

| module | models |
|---|---|
| mdm | `DEPDevice` |
| monolith | `PkgInfoName`, `PkgInfo`, `SubManifestPkgInfo` |
| osquery | `ConfigurationPack`, `Enrollment` |
| turbo | `Enrollment`, `RecurringJob`, `OneTimeJob` |

Each model keeps the links to its parents. Only the link to itself goes away.

## One model keeps its self key

`mdm.DeviceCommand` also writes its own key, and it keeps it: it links itself by uuid and not by pk. That is the reason why `build()` lets a model keep control of its own key, and the comment in `zentral/core/events/base.py` gives it as the example.

## Why the events do not change

`linked_objects_keys_for_event()` has two callers only: `AuditEvent.build()`, which supplies the missing self key, and `DeviceCommand`, which calls `super()` on its own parent. No serializer and no view reads the hook, so there is no other path where the removed line did work.

The tests of the four modules compare the full `objects` dictionary of the event, so they show the result directly. `tests/turbo/test_api_recurring_jobs.py` compares the set of the keys and finds `turbo_recurring_job` in it, and `tests/munki/test_api_views.py` compares the whole dictionary. They pass with no change.

There is no changelog entry, because a user sees no difference.

## The new test

`tests/core_events/test_audit_event_objects.py` walks the models and finds a model that writes its own key with its own pk.

The test reads the source of the hook, because the hook needs an instance to run. It finds the two spellings that the code uses, `[(self.pk,)]` and `((self.pk,),)`, and not every possible one — it is a guard and not a proof. `mdm.DeviceCommand` needs no exception in it, because the rule is the pk spelling and that model writes `str(self.uuid)`.

A second test makes sure that the traversal finds the hooks. A guard that stops traversing passes for the wrong reason.

To show that the guard works, I put one self key back and ran it: it failed with `['osquery.Enrollment osquery_enrollment'] != []`.

## Tests

- full suite: 8030 tests, OK (8028 before, plus the two new ones)
- patch coverage: 8 added executable lines in `zentral/`, 0 uncovered

### DIFF
--- a/tests/core_events/test_audit_event_objects.py
+++ b/tests/core_events/test_audit_event_objects.py
@@ -1,4 +1,7 @@
-from django.test import TestCase
+import inspect
+import re
+from django.apps import apps
+from django.test import SimpleTestCase, TestCase
 from zentral.core.events.base import AuditEvent, linked_objects_key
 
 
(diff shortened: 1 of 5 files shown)


soysauceconspiracy (approved):
LGTM

PR #1656, opened 2026-08-26
monolith: keep the file name of a rebuilt enrollment package

np5:
`build_manifest_enrollment_package` deleted the current file of the package, then saved the new content under the canonical name. When an orphan file used that name, a storage that does not overwrite files saved the package under a name with a random suffix. A rolled back transaction leaves such an orphan: the storage keeps the file, and the database keeps the old row.

The build now also deletes the file at the target name before it saves the package. The package always keeps its expected name. The name contains the primary key of the package, so the file cannot belong to another object. The delete of a name that does not exist is a no-op, on the file system storage and on S3. The S3 storage already overwrites the file at that name.

This also made the `rebuild_manifest_enrollment_packages` command tests flaky. A test database repeats the same primary keys on each run, and the rolled back test transactions do not remove the uploaded files. When the media volume keeps the files of a previous run, the fixture files collide with them, and `test_rebuild_with_a_storage_without_a_local_path` fails: the command prints the new file name, and the test compares it with the old one. The runs on a fresh CI volume did not show it.

The new test makes the package file an orphan, rebuilds, and checks that the package takes its name back. It fails without the fix, on a clean volume and on a dirty volume.

Verification, in an isolated compose stack:

- The old code passes the 10 existing command tests on a clean volume, and fails the new test.
- The old code on a dirty volume reproduces the reported failure.
- The fixed code passes all 11 command tests on a dirty volume.
- `tests.monolith`: 606 tests OK. Full suite with coverage: 8112 tests OK. All the changed lines of `zentral/contrib/monolith/utils.py` are covered.


soysauceconspiracy (approved):
LGTM

PR #1658, opened 2026-08-27
Expire the event store cache after 300s

np5:
## Problem

The store cache is loaded one time and then invalidated only by the `stores.store` notifier callback. The notifier gives no guarantee of delivery, and a notification that never arrives leaves the cache in place for ever: the instance keeps sending events to a store with the configuration it read when it started, until an operator restarts the process.

`PoliciesCache` carries a fallback for exactly this reason, with the comment "Fallback for missed notifier notifications". `MonolithConf` has an equivalent, `reload_interval`, of one hour. `Stores` had none.

## Change

`max_age_seconds = 300` on `Stores`, with the same shape as `PoliciesCache`: a load timestamp, a freshness test, and `clear()` resets the timestamp. The value is the one `PoliciesCache` uses.

## Probe views: moved to a follow-up

The first version of this pull request gave the same fallback to `ProbeView`. @Labeeb2339 found a defect in it: the probe views form a tree, `all_probes_dict` is a child of `all_probes`, and giving each view an age of its own lets a child stay fresh while its parent expires and reloads, so the child serves a mapping that misses the change. I reproduced it, and the test I had written missed it because it moved the parent and the child timestamps together.

The correction needs a parent and its child to agree on the data they serve, which is a change of design rather than an added condition. Repairing it on top of two earlier commits made the result harder to follow than the design itself, so the probe views come in a separate pull request that makes the change one time. This one keeps the part that is simple and complete on its own.

## Tests

`Stores` goes to **100%**, not only the changed lines: it decides which store an event goes to, so the parts this change does not touch are worth covering too. It was at 89%.

The admin console resolution (several admin console stores, none, no store at all), `iter_events_url_store_for_user` for each of its four outcomes, `iter_queue_worker_stores`, `_start_sync` with and without sync, and the max age itself. `force_store()` takes `admin_console` now, which the store cases need.

`tests.core_stores` and `tests.core_probes`: 420 tests pass.


andreas-zentral (dismissed):

zentral/core/probes/conf.py
```diff
@@ -57,14 +70,16 @@ def __init__(self, parent=None, item_func=None, unique_key=True, with_sync=False
...
         self._start_sync()
-        if self._probes is None:
-            self._probes = {}
-            for probe in self.iter_parent_probes():
-                for key, val in self.item_func(probe):
-                    if self.unique_key:
-                        self._probes[key] = val
-                    else:
-                        self._probes.setdefault(key, []).append(val)
+        if self._cache_is_fresh():
```
Labeeb2339: Could you add a regression test for staggered parent/child expiry? all_probes_dict is a child of all_probes, but ProbesDict._load() returns as soon as the child cache is fresh. If the child reloads at t=299s while the parent is still fresh, then all_probes expires and reloads at t=301s, the child remains fresh until roughly t=599s and can serve a mapping that omits probe changes made after the parent reload. The current test_expired_child_reloads_through_expired_parent backdates both timestamps, so it does not cover this skew. Please make child freshness depend on the parent and reload the parent before serving, or add equivalent invalidation, then assert the staggered case.

Labeeb2339 (commented):
The cache-expiry coverage is useful, but I found one derived-cache invalidation gap in the new parent/child cache behavior.

zentral/core/probes/conf.py
np5: You are right on both points, and thank you - I reproduced it before changing anything. The test you describe fails on the old code with a `KeyError` on the new probe: the child keeps the mapping it built from the parent that has since reloaded.

You are also right that `test_expired_child_reloads_through_expired_parent` hid it. It moved both timestamps together, which is the one arrangement where the bug cannot appear.

I took the first of the two options you give, because pinning the skew with a test would have made a documented compromise permanent. In e3341bec a derived view has no age of its own any more. It records the generation of the data its parent served and reloads when that number changes. Only the root keeps the max age, so the bound is now the max age for the whole tree, at any depth, and not twice it at depth one.

Asking the parent for its generation makes the parent apply its own max age first, so a view is never built on data its source should already have replaced. Only the parent lock is taken while a child holds its own, which is the direction the comment on `ProbeList.clear()` asks for; the cascade in the other direction still happens outside the lock.

One case is left, and it is safe: the parent can reload between the call for its generation and the iteration that follows. The child then stores a number older than the data it built from and reloads one more time. That costs work, never staleness.

The new `test_child_follows_parent_expiry_when_staggered` moves the parent timestamp alone, with the child fresh, which is exactly the arrangement you describe. `zentral/core/probes/conf.py` stays at 100%.

np5: Update: the probe views are out of this pull request.

Repairing the tree semantics on top of the two commits that were already here made the result harder to read than the design is on its own, so the probe views move to a follow-up that makes the change one time, with your case as one of its tests. This pull request keeps only the store cache, which has no parent and no child and is complete on its own.

Your finding stands and is what caused the split. Thank you again.

sebastian-zentral (approved):

zentral/core/probes/conf.py
np5: The follow-up is open: zentralopensource/zentral#1659.

It does the tree semantics once instead of as a correction. A view now has a version number that increases on every build, only a root has an age, and a derived view takes a snapshot of its parent, which returns the version and the probes together. The bound is one max age for the whole tree, at any depth.

Your case is `test_child_follows_parent_expiry_when_staggered` there, and it fails without the change. There is a second one next to it, `test_child_has_no_age_of_its_own`, which moves the child alone and expects nothing to change - the assertion the earlier design could not make.

PR #1660, opened 2026-08-27
Restrict the task result endpoints to the user of the task

np5:
## Problem

`/api/task_result/<uuid:task_id>/` and `/api/task_result/<uuid:task_id>/download/` set `authentication_classes` but no `permission_classes`, so the project default `IsAuthenticated` was the whole check. Neither view filtered on a user. Any authenticated principal that had a task ID could read the state of any task, and download any exported file, whatever the model permissions that gated the export in the first place. The endpoint is shared by every export in Zentral: inventory, MDM, Monolith, Osquery, Santa, Intune and Workspace ONE.

The link to make the check with exists. `accounts.UserTask` joins a user to a `TaskResult`, and the console task list and task pages already use it. It was written but never read for authorization on the API.

## Part 1 — the tasks with no user

There are 22 launch sites. `apply_async` is the only entry point: there is no `.delay()`, no `chain`, `group`, `chord` or `signature`, and no Celery beat schedule, so no periodic tasks. 19 of the 22 pass `task_user`. The other three:

| Site | Kind | Decision |
|---|---|---|
| `mdm/views/dep_enrollments.py` `define_dep_profile_task` | user-initiated | **Fixed.** It passes `task_user` now, and the task takes `**kwargs` like the others. Nothing polls it — the view redirects — so it only ever appeared in the task list, and it did not appear there. |
| `osquery/public_views.py` `build_file_carving_session_archive` | device-initiated | **Correct as is.** An Osquery agent finishes a carve upload and there is no user. The task writes to `FileCarvingSession.archive` and returns no `filepath`, so it never reaches the download endpoint. |
| `mdm/tasks.py` `assign_dep_virtual_server_default_enrollment_task` | chained | **Correct as is.** It inherits the user of the synchronization that scheduled it. Its only parent is an API view that always passes one. |

Two of the 26 tasks had no `**kwargs`. `define_dep_profile_task` needed it; `build_file_carving_session_archive` does not, because it is never given a user.

A task with no user is superuser-only from here on. Nothing in the console or the API polls one, so no flow breaks: every `.task` link in the templates points at an API view that passes `task_user`.

## Part 2 — the enforcement

Both views read through `task_results_for_user()`, which scopes to the `UserTask` row. A superuser sees every task, and a service account never does, the rule of `ZentralBackend.has_perm`. `User.save()` already forces `is_superuser = False` on a service account; the check does not depend on that.

A task that is hidden is a task that does not exist. The state endpoint answers `UNKNOWN`/`unready`, which is what it already answered for an unknown ID, and the download endpoint gives a `404`. Neither confirms that the ID exists. The console views now share the queryset, so the two surfaces cannot drift apart.

## The edge cases

**Best-effort ownership.** `create_task_result_on_publish` logged any failure and published the task anyway. That silence becomes an unexplainable `404` once the endpoints check ownership, so the write fails the launch now. This is safe: `before_task_publish` runs before `producer.publish`, and Django signals do not swallow exceptions, so the message never reaches the broker. Inside a request that is a clean rollback under `ATOMIC_REQUESTS`.

A missing user is the exception, and is still only logged. The cascade already removed the ownership row with the user, so there is nobody left to attribute the task to and nobody to keep it from — and failing there would kill a DEP assignment mid-retry for no gain.

**Retries.** `self.retry()` republishes the same task ID, so `UserTask.objects.create` hit the `OneToOne` constraint and logged an error on every retry of the DEP default enrollment assignment. It is a `get_or_create` now, which also keeps the first user.

**`ATOMIC_REQUESTS` orphans.** The `UserTask` row is written inside the request transaction while the broker publish is not transactional, so a view that raised after `apply_async` would leave a queued task with no ownership row. No launcher can: 20 of them return a `Response` or a `redirect` immediately, and the two remaining ones publish from `transaction.on_commit`, after the commit.

**Old task results.** A task from before v2025.11 has no `UserTask` at all, so only a superuser can read it. Exports are temporary — launching the export again gives a new file. Noted as a backward incompatibility.

**Service accounts** launch and download as the same principal, so they keep working. Covered by a test.

## Testing

Full suite: 8220 tests, green. 100% patch coverage, verified by intersecting `coverage json` `missing_lines` with the `git diff -U0` hunks.

Unrelated: `tests/monolith/test_rebuild_manifest_enrollment_packages_cmd.py::test_rebuild_with_a_storage_without_a_local_path` is order-dependent and failed in one full run. It asserts on a filename captured before the rebuild, so it breaks whenever the media storage has to rename the file. Deleting the stale files under `/var/zentral/monolith` makes it pass. It is not touched here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,12 @@ The `inventory_jmespath_check_created`, `inventory_jmespath_check_updated` and `
 
 `zentral_santa_enrolled_machines_total` and `zentral_santa_enrolled_machines_sync_total` are replaced by `zentral_santa_enrolled_machines_bucket` and `zentral_santa_enrolled_machines_sync_bucket`, which have an `le` label. The `le="+Inf"` bucket gives the value of the two removed gauges: in a dashboard or an alert, change the metric name and select that bucket. A query that sums the buckets counts each machine seven times.
 
+#### 🧨 Privilege escalation hardening — task results
+
+A task belongs to the user or the service account that launched it, and `/api/task_result/<task_id>/` and `/api/task_result/<task_id>/download/` answer only for that principal. A superuser reads all of them, as before. Until now the two endpoints only asked for a valid session or API token: any authenticated principal that had a task ID could read the state of that task and download its file, and an export carries everything the principal that launched it was allowed to see. The state endpoint gives the `UNKNOWN` status for the task of a different principal, like it does for a task that does not exist, and the download endpoint gives a `404`. The task list and the task pages of the web console already applied this rule.
+
+A task that a device launched, and a task that Zentral launched before version 2025.11, has no user, so only a superuser can read those. The exports are temporary: launch the export again to get a file that you can download.
+
 
 ### Bug fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -125,6 +125,8 @@ The task that pushes an updated DEP profile to Apple Business Manager belongs to
 
 A retried task does not write an error to the logs about its user record anymore. Celery publishes a retry with the same task ID, Zentral tried to write the record a second time, and the unique constraint refused it. The task that assigns the default enrollment of a DEP virtual server retries when Apple throttles it.
 
+The task list of the web console reads the user of each task together with the tasks now. It made one more query for each row of the page, and a superuser sees the tasks of all the users.
+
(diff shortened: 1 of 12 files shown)


sebastian-zentral (dismissed):

soysauceconspiracy (approved):
LGTM

PR #1661, opened 2026-08-27
mdm: fix the re-enrollment completion and the user short name of the check-in view

np5:
A re-enrollment session was never completed, so `/public/mdm/checkin/` answered with a 500 and posted no event on each token update of the device. The failure repeats for as long as the device stays enrolled. Two commits: that fix, and one optional key of the same message that Zentral wrote unconditionally.

The links to the schemas point at [Release-v26.4](https://github.com/apple/device-management/tree/67045e2fa06f528b196c01edee6a8bf88b844beb) of `apple/device-management`.

## A re-enrollment completes on its token update

`set_completed_status` required the `AUTHENTICATED` status, so a re-enrollment session could stay in `STARTED` for ever:

```
File "zentral/contrib/mdm/public_views/mdm.py", in do_token_update
File "zentral/contrib/mdm/models.py", in set_completed_status
File "zentral/contrib/mdm/models.py", in _set_next_status
    raise EnrollmentSessionStatusError(self, next_status)
zentral.contrib.mdm.exceptions.EnrollmentSessionStatusError: (<ReEnrollmentSession: ...>, 'COMPLETED')
```

Every traceback on this path is a `ReEnrollmentSession`. None are DEP, OTA or user sessions, because only a re-enrollment sends no authenticate. `build_mdm_configuration_profile` builds every enrollment profile of a deployment with `get_payload_identifier("mdm")`, and Apple makes that value the selector ([com.apple.mdm.yaml](https://github.com/apple/device-management/blob/67045e2fa06f528b196c01edee6a8bf88b844beb/mdm/profiles/com.apple.mdm.yaml#L315)):

> The system uses this value to determine whether to replace an existing profile or add a new one.

The device replaces the profile it holds, updates the MDM payload, and does not run the installation the authenticate belongs to. Apple gives the rest of the rule in the `Authenticate` notes ([authenticate.yaml](https://github.com/apple/device-management/blob/67045e2fa06f528b196c01edee6a8bf88b844beb/mdm/checkin/authenticate.yaml#L227-L232)):

> When the device successfully installs the MDM payload, it sends a `Token-Update` message.

`ReEnrollmentSession.set_completed_status` accepts `STARTED` now. Each session class defines its own method, so the change applies to a re-enrollment only. `DEPEnrollmentSession`, `OTAEnrollmentSession` and `UserEnrollmentSession` keep theirs, and those still require `AUTHENTICATED`. The `self.enrolled_device == enrolled_device` test stays as well, and `ReEnrollmentSessionManager` is the only manager that sets `enrolled_device` when it creates the session. A stale re-enrollment session is safe too: the view reads the session from the common name of the client certificate, so only a token update over the new identity reaches it.

The token update also writes the certificate attributes now, because `cert_not_valid_after` starts the next re-enrollment (`commands/scheduling.py::_reenroll`) and only the authenticate wrote it. A re-enrollment that never authenticated kept the old date, and the device re-enrolled every four hours.

Both status call sites abort now instead of raising, as the rest of the view already does, and carry the session status on the failure event. `do_authenticate` needs it for a reason of its own: an authenticate against a session that is already authenticated raised the same exception, and a device does send one. The refusal stays, because at that point `do_authenticate` has staged the clearing of `token` and `push_magic` and a `purge_state(full=True)` on an enrolled device, and only the rollback of the request keeps that out of the database. Only the report changes.

## The short name of a user is optional

Apple gives `UserShortName` `presence: optional` ([tokenupdate.yaml](https://github.com/apple/device-management/blob/67045e2fa06f528b196c01edee6a8bf88b844beb/mdm/checkin/tokenupdate.yaml#L121-L136)), and `EnrolledUser.short_name` is a `models.TextField()`, so the column is not nullable. `do_token_update` wrote it from the payload with no test, so a token update without the key raised `IntegrityError: null value in column "short_name" of relation "mdm_enrolleduser" violates not-null constraint`, and gave a 500 with no event.

The view writes the column only when the payload carries the key now. That is what it already does with the two other optional keys of the message, the [`UnlockToken`](https://github.com/apple/device-management/blob/67045e2fa06f528b196c01edee6a8bf88b844beb/mdm/checkin/tokenupdate.yaml#L184-L201) and the [`AwaitingConfiguration`](https://github.com/apple/device-management/blob/67045e2fa06f528b196c01edee6a8bf88b844beb/mdm/checkin/tokenupdate.yaml#L204-L214). A user created without the key gets the empty default of the column, which `EnrolledUser.__str__` already expects: `self.long_name or self.short_name`.

## Tests

Seven new tests, five for the first commit and two for the second. Each commit is green on its own. A revert of the source of a commit, with its tests kept, reproduces its failures: the same exception arguments for the first, and the integrity error above for the second.

`tests.mdm`: 2847 tests, OK.

## Noted, not fixed

[`Token`](https://github.com/apple/device-management/blob/67045e2fa06f528b196c01edee6a8bf88b844beb/mdm/checkin/tokenupdate.yaml#L176-L179), [`PushMagic`](https://github.com/apple/device-management/blob/67045e2fa06f528b196c01edee6a8bf88b844beb/mdm/checkin/tokenupdate.yaml#L180-L183) and [`UserLongName`](https://github.com/apple/device-management/blob/67045e2fa06f528b196c01edee6a8bf88b844beb/mdm/checkin/tokenupdate.yaml#L161-L175) are written unconditionally, and all three are `presence: required`. A payload without one of them clears a column or gives a 500, but Apple requires the keys and none of that was ever observed, so nothing here tests for them.

`AwaitingConfiguration` from the user channel. Apple gives the key `userchannel: false` on macOS, and says a `true` from the user channel is for Shared iPad, where it means the device waits for a `UserConfigured` command. The token update writes the value to `EnrolledDevice.awaiting_configuration` on the two channels, and `DeviceConfigured.verify_channel_and_device` then sends a `DeviceConfigured` on the device channel. Zentral has no `UserConfigured` command. This predates the change.


soysauceconspiracy (approved):
LGTM

PR #1663, opened 2026-08-28
santa: reject the compiler policy on an incompatible target type

np5:
## Problem

`Rule.Policy.ALLOWLIST_COMPILER` was accepted on any target type. Nothing tied the policy to a compatible target: not the model, not the forms, not the serializers.

The Santa client only accepts the compiler state on three rule types. `SNTPolicyProcessor.mm` looks the `(type, state)` pair up in a decision map, and that map has an entry for the compiler state only under `SNTRuleTypeCDHash`, `SNTRuleTypeBinary` and `SNTRuleTypeSigningID`:

```objc
{{SNTRuleTypeCDHash, SNTRuleStateAllowCompiler}, SNTEventStateAllowCompilerCDHash},
{{SNTRuleTypeBinary, SNTRuleStateAllowCompiler}, SNTEventStateAllowCompilerBinary},
{{SNTRuleTypeSigningID, SNTRuleStateAllowCompiler}, SNTEventStateAllowCompilerSigningID},
```

There is no such pair for `SNTRuleTypeTeamID` or `SNTRuleTypeCertificate`. The lookup misses, and the client logs an invalid type/state combination and returns `NO` — it treats the rule as if it were not found.

So a compiler rule on a team ID or a certificate target passed validation, synchronized cleanly, and then did nothing on the client. The failure is silent, and worse than a rejection: the target ends up with no rule at all, so it is blocked in lockdown mode. The `NSException` in the `SNTRuleStateAllowCompiler` branch further down is not what a server can trigger — the map lookup already returned before it, and the three cases it handles are the three valid ones.

## The rule

`Target.Type` carries it, the way it already carries `is_native`, and the message enumerates the members the way `rule_choices()` already does:

```python
@property
def compatible_with_compiler_policy(self):
    # the client only accepts the compiler state on these rule types
    return self.value in (self.SIGNING_ID, self.BINARY, self.CDHASH)

@classmethod
def compiler_policy_error(cls):
    return "Only available for {} targets".format(
        ", ".join(member.value for member in cls if member.compatible_with_compiler_policy)
    )
```

`Rule.Policy` holds the other half — that only the compiler policy constrains the target type at all:

```python
def compatible_with_target_type(self, target_type):
    # only answers for the policy itself: a CEL expression can return ALLOWLIST_COMPILER
    # too, and the client resolves the CEL state before the (type, state) lookup, so a CEL
    # rule reaches the same invalid combination. The expression is opaque to the server.
    if self.value != self.ALLOWLIST_COMPILER:
        return True
    return Target.Type(target_type).compatible_with_compiler_policy
```

The message runs off the same predicate the check does, so it cannot name a set that is not enforced. It enumerates in declaration order and renders as `Only available for SIGNINGID, BINARY, CDHASH targets`.

## Where it is checked

| Entry point | Where |
|---|---|
| Model | `Rule.clean()` |
| Create form | `RuleForm.clean()` |
| Update form | `Rule.clean()`, through `ModelForm._post_clean` |
| Rule API, create and update | `RuleSerializer.validate()` |
| Ruleset API | `RuleUpdateSerializer.validate()` |

All five report `Target.Type.compiler_policy_error()` on the `policy` field.

`RuleForm` is a plain `forms.Form` and never touches model validation, so it needs the explicit check. `UpdateRuleForm` is a `ModelForm`, so `Rule.clean()` already reaches it and maps onto its `policy` field — a second check there would report the error twice, and `assertFormError` compares the error list exactly, so the test proves one arrives and only one.

**A check constraint is not an option.** `policy` is a column of `santa_rule` and the type is a column of `santa_target`; a Postgres `CHECK` cannot span the two tables. `clean()` is the model-level mechanism available.

Voting rules cannot produce one: the upsert in `utils.py` only ever writes policy `1` or `2`.

## Drive-by fix

`RuleForm.clean()` read `policy` on a path where it was never bound:

```python
try:
    policy = int(cleaned_data.get("policy"))
except (TypeError, ValueError):
    pass

if policy:  # UnboundLocalError
```

A rule form submitted without a policy makes `int(None)` raise, so `cleaned_data` has no `policy`, so the `except` runs and `if policy:` raised `UnboundLocalError` — a 500 instead of "This field is required." on the field. It is one line above the check this PR adds, so it is fixed here, with `test_create_configuration_rule_missing_policy` to hold it. Confirmed by removing the fix and watching the test fail on the `UnboundLocalError`.

## Tests

Each entry point asserts that the two incompatible types are rejected. All of them except the rule `PUT` also assert that the three compatible types are accepted. The `PUT` shares `RuleSerializer.validate()` with the `POST`, and the `POST` covers the accept path.

- `test_models.py` — `Rule.clean()`, and an incomplete rule, since `clean()` also runs after the field validation failed and the policy can be missing or unknown there
- `test_setup_views.py` — the create and update rule views
- `test_api_views.py` — rule `POST` and `PUT`, and `ruleset_update`
- `test_serializers.py` — `RuleUpdateSerializer`, next to its other validation tests

They assert against `Target.Type.compiler_policy_error()` rather than a copy of the string. `test_compiler_policy_error` pins what it renders to and `test_compatible_with_compiler_policy` pins the members and their order, so a broken format string or a silently widened set still fails.

Full suite: 8281 tests, OK. `flake8` and `ruff` clean on the changed files. `mkdocs build --strict` passes. 100% coverage on the edited lines.

## Docs

The `policy` row of the ruleset API rule attributes gets the constraint. The `rule_type` and `identifier` rows of the same table were stale, from before the cdhash and signing ID types: they listed `BUNDLE`, which the serializer rejects (there is a test for it), and omitted `CDHASH` and `SIGNINGID`. Corrected, so the table does not contradict the new note.

A second commit corrects the *Definitions* section, one screen above the table. It listed four target types and four policies, from before the cdhash, the signing ID and the CEL rules, and gave the sha256 hexdigest as the identifier of all the targets. It is the first description of a rule that a reader meets. The **Allowlist compiler** paragraph carries the constraint now, and a short paragraph describes the CEL policy, which the page did not document at all.

## Existing rows

This is validation, so the rows already in a database keep their policy, and there is no data migration. Nothing writes them again though. A ruleset that carries such a rule gives a 400 on the next post, and so does a `PUT` on the rule API. `santa/terraform.py` exports the policy through `get_translated_policy()`, so a Terraform configuration that carries such a rule fails on the next apply. The update form of the web console rejects a change to any other attribute of the rule as well, because `Rule.clean()` validates the policy and the target type together.

The CHANGELOG entry is in *Backward incompatibilities* for that reason, and not in *Features*.

## Noted, not fixed

**A CEL rule reaches the same invalid combination.** `Policy.CEL` is allowed on any target type, and `cel_expr` is opaque to the server. On the client the CEL result replaces the state before the map lookup (`state = celResult.resultState`), and `ReturnValue::ALLOWLIST_COMPILER` gives `SNTRuleStateAllowCompiler` in the v1 and the v2 traits. A CEL rule on a team ID or a certificate whose expression returns `ALLOWLIST_COMPILER` therefore drops in the same way, and Santa's own documentation gives a team ID rule as an example of where to attach a CEL expression. The server cannot see this without an evaluation of the expression, so the check does not cover it. `Policy.compatible_with_target_type()` carries a comment that says so: it answers for the policy alone.

A partial `PATCH` to `/api/santa/rules/<pk>/` raises `KeyError: 'target'` on the first line of `RuleSerializer.validate()`, which reads `data["target"]` unconditionally. It predates this change and fails before any policy check. The endpoint is a `RetrieveUpdateDestroyAPIView`, so `PATCH` is routed but only ever works with a full payload.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,10 @@ The `inventory_jmespath_check_created`, `inventory_jmespath_check_updated` and `
 
 `zentral_santa_enrolled_machines_total` and `zentral_santa_enrolled_machines_sync_total` are replaced by `zentral_santa_enrolled_machines_bucket` and `zentral_santa_enrolled_machines_sync_bucket`, which have an `le` label. The `le="+Inf"` bucket gives the value of the two removed gauges: in a dashboard or an alert, change the metric name and select that bucket. A query that sums the buckets counts each machine seven times.
 
+#### 🧨 Santa compiler rules on team ID and certificate targets
+
+A rule with the `ALLOWLIST_COMPILER` policy is rejected on a target that is not a cdhash, a binary or a signing ID, in the web console, the rule API and the ruleset API. The Santa client only accepts the compiler state on those three rule types. On a team ID or a certificate it drops the rule when it evaluates it, so the rule synchronized cleanly and then left its target with no rule at all, which blocks the target in lockdown mode. The rules already in the database keep their policy, but Zentral rejects each of them at the next write: a ruleset that you post again, a `PUT` on the rule API, a Terraform configuration that you apply again, and a change to any other attribute of the rule in the web console.
+
 #### 🧨 Privilege escalation hardening — task results
 
 A task belongs to the user or the service account that launched it, and `/api/task_result/<task_id>/` and `/api/task_result/<task_id>/download/` answer only for that principal. A superuser reads all of them, as before. Until now the two endpoints only asked for a valid session or API token: any authenticated principal that had a task ID could read the state of that task and download its file, and an export carries everything the principal that launched it was allowed to see. The state endpoint gives the `UNKNOWN` status for the task of a different principal, like it does for a task that does not exist, and the download endpoint gives a `404`. The task list and the task pages of the web console already applied this rule.
--- a/docs/apps/santa.md
+++ b/docs/apps/santa.md
@@ -118,7 +118,7 @@ The payload is an array of [NotificationSettingItem](https://developer.apple.com
 
 ### Definitions
 
-Zentral Santa rule combine a *target* and a *policy*. The target can be of type `Binary`, `Bundle`, `Certificate` or `Team ID`. A target is uniquely identified by its type, and its identifier (sha256 hexdigest for `Binary`, `Bundle` and `Certificate` targets). The policy can be one of `Allowlist`, `Blocklist`, `Silent blocklist` and `Allowlist compiler`.
+A Zentral Santa rule combines a *target* and a *policy*. The target can be of type `cdhash`, `Binary`, `Signing ID`, `Certificate` or `Team ID`. A target is uniquely identified by its type, and its identifier: a sha256 hexdigest for `Binary` and `Certificate`, a 40 character hexdigest for `cdhash`, `TEAMID:SIGNINGID` for `Signing ID`, and the 10 character team ID for `Team ID`. `Bundle` is a sixth type that only exists in Zentral, see below. The policy can be one of `Allowlist`, `Allowlist compiler`, `Blocklist`, `Silent blocklist` and `CEL`.
 
 To avoid conflicts, there is at most **one rule per target** for each configuration.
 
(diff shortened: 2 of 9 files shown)


soysauceconspiracy (approved):
LGTM

PR #1664, opened 2026-08-28
monolith: add a pkginfo data view

sebastian-zentral:
The version of a PkgInfo in the *PkgInfos* list and on a PkgInfo name page
links to its data now: the pkginfo as stored, the pkginfo as served in the
catalogs, and the rewrites between the two.

Gated by a new `Monolith::Action::"viewPkgInfoData"` PBAC action, because the
raw pkginfo carries the installation checks and the pre/postinstall scripts.
It is admin-only – not part of the viewer action groups – and takes User
principals only, since the data has no API endpoint.

The action takes the PkgInfo as its resource, with its repository and its
active catalogs as parents, and the repository with its meta business unit as
parent. Nothing goes in the context. A policy can be scoped to one PkgInfo, to
a catalog, to a repository, or to the meta business unit that already scopes
the machines.

A scope on a catalog is existential, so a restricted catalog is taken back
with a `forbid` policy. Every active catalog of a PkgInfo is a parent of the
resource, whatever the list is filtered on.

One decision per PkgInfo on the lists, batched in a single call.

Docs and CHANGELOG included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,10 @@ Zentral adds a `default_installs` package to the `optional_installs` of the sub 
 
 The Monolith API publishes the same `zentral_audit` events as the web console now. The catalogs, the conditions, the enrollments and the manifest enrollment packages changed with no record of it.
 
+The version of a PkgInfo in the *PkgInfos* list and on a PkgInfo name page links to its data now: the pkginfo as it is stored in Zentral, the pkginfo as it is served in the catalogs, and the differences between the two. Use it to see the rewrites monolith applies, for example the `installer_item_location` it points at its own repository endpoint.
+
+The raw pkginfo carries the installation checks and the pre/postinstall scripts, so the new `Monolith::Action::"viewPkgInfoData"` PBAC action gates the page, and it is not part of the viewer action groups. It takes the PkgInfo as its resource, with its repository and its active catalogs as parents, and the repository with its meta business unit as parent, so a policy can be scoped to one PkgInfo, to a catalog, to a repository, or to the meta business unit. A `forbid` policy on a catalog takes a restricted catalog back. The action has no API endpoint, and no service account can reach it.
+
 
 #### Munki
 
--- a/docs/apps/monolith.md
+++ b/docs/apps/monolith.md
@@ -104,6 +104,45 @@ First Zentral checks the `excluded_tags`. If a machine has any one of the tags l
 
 **IMPORTANT** Make sure that the tags used in the PkgInfo exist in Zentral **before** the repository is synced.
 
+## PkgInfo data
+
+The version of a PkgInfo in the *PkgInfos* list and on a PkgInfo name page links to its data: the pkginfo as it is stored in Zentral, and the pkginfo as it is served in the catalogs, with the differences between the two – the *rewrites* monolith applies, for example the `installer_item_location` it points at its own repository endpoint, or the `catalogs` key it drops.
+
+* PBAC action: `Monolith::Action::"viewPkgInfoData"`
+
+The raw pkginfo carries the installation checks and the pre/postinstall scripts, so this action is not part of the viewer action groups, unlike the other monolith view actions: `Monolith::Action::"ViewerActions"` does not grant it, and neither does `monolith.view_pkginfo`. It has to be granted explicitly, and only to a role, not to a service account – the pkginfo data has no API endpoint.
+
+The action takes the PkgInfo as its resource, with its repository and its active catalogs as parents, and the repository with its meta business unit as parent. Nothing goes in the context. A policy can be scoped to one PkgInfo, to a catalog, to a repository, or to the meta business unit that already scopes the machines:
+
+```
+permit (
+  principal in Role::"6",
+  action == Monolith::Action::"viewPkgInfoData",
+  resource in Monolith::Repository::"3"
+);
+```
+
+`resource in Monolith::Catalog::"7"` matches the PkgInfos of that catalog, `resource in Inventory::MetaBusinessUnit::"2"` the PkgInfos of every repository of that meta business unit, and `resource == Monolith::PkgInfo::"12"` one PkgInfo.
+
+A scope on a catalog is existential: a PkgInfo of the `testing` **and** of the `production` catalog matches a policy scoped on `testing`. Take a restricted catalog back with a `forbid` policy, which wins over every `permit`:
+
+```
+permit (
+  principal in Role::"6",
+  action == Monolith::Action::"viewPkgInfoData",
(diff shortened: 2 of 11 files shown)


np5 (commented):
I think `viewPkgInfoData` must take the **PkgInfo** as its resource, with the Repository and the Catalogs as parents.

## The two Cedar rules the current shape breaks

**[Don't use the context field to hold information about the principal, action, and
resource](https://docs.cedarpolicy.com/bestpractices/bp-using-the-context.html).**
`repositoryName`, `catalogIDs` and `catalogNames` are all resource information. Cedar keeps
the context for the *how* and the *when* of a request: action parameters and request
metadata. The entities carry the *who* and the *what*.

**[Every resource lives in a
container](https://docs.cedarpolicy.com/bestpractices/bp-resources-containers.html).** This
rule gives the anchor point of a permission. It does not replace the resource of the
request. You grant "create a file" on the folder, because the file does not exist yet. A
PkgInfo that you read does exist.

## The precedent is already in the repository

[`inventory/pbac.py:29`](https://github.com/zentralopensource/zentral/blob/c829a335/zentral/contrib/inventory/pbac.py#L29):

```python
MACHINE_RESOURCE_TYPE = ResourceType("Machine", get_namespace(), parents=(MBU_RESOURCE_TYPE,))
```

`get_meta_machine_resource()` gives a Machine **several** MetaBusinessUnit parents. This is
the same many-to-many shape as PkgInfo/Catalog, and nothing goes in the context.

The other two typed actions do not disagree with this:

* `createMachineTag` takes the Machine, because the MachineTag does not exist yet. This is
  the container rule above.
* `forceInstallArtifact` puts the artifact in the context, because the artifact is a
  parameter of the action. Cedar gives the same example: the recipient when you share a
  folder.

`viewPkgInfoData` is different. The catalogs are not parameters of the action. They are the
containers of the PkgInfo.

The engine supports the hierarchy already: `EntityType(parents=…)` becomes `memberOfTypes`,
and `_serialize_entity` collects the parents and removes the duplicates across a batch.

## What I verified with cedarpy

Hierarchy: `PkgInfo → [Repository, Catalog*]`, `Catalog → Repository`,
`Repository → Inventory::MetaBusinessUnit`. Action `resourceTypes: [Monolith::PkgInfo]`.

| policy scope | schema validation | evaluation |
| --- | --- | --- |
| `resource == Monolith::PkgInfo::"12"` | OK | – |
| `resource in Monolith::Catalog::"3"` | OK | matches a PkgInfo in `[2,3,4]`, not one in `[5,6,7]` |
| `resource in Monolith::Repository::"1"` | OK | matches through the Catalog parent, transitively |
| `resource in Inventory::MetaBusinessUnit::"7"` | OK | matches, two levels up |
| `resource in System::"any"` | rejected at write time | – |

A policy can be scoped to a repository, which is what this PR wants. It can also be scoped
to a catalog, to one PkgInfo, and to the meta business unit. The last one connects monolith
to the scope that machines already use. None of these are possible when the repository is
the resource.

## The cost

The note in the docs is correct: this makes one decision per PkgInfo, and `PkgInfosView` is
not paginated. Measured with `is_authorized_batch`, median of 7 runs, 4 policies:

| PkgInfos on the page | one decision per (repository, catalog set) | one decision per PkgInfo |
| --- | --- | --- |
| 100 | 0.26 ms | 2.7 ms |
| 1 000 | 0.26 ms | 26 ms |
| 3 000 | 0.24 ms | 83 ms |
| 10 000 | 0.25 ms | 294 ms |

Decomposed, this is approximately 22 µs for each request and 4 µs for each entity, thus
approximately **26 µs for each PkgInfo**. The page already runs a full table SQL aggregate
and renders each row through a Django template. The cost is real at 10 000 PkgInfos. I think
the correct answer there is to paginate the list, and not to change the model.

## The difference in meaning, and how to keep it

`resource in Catalog::"x"` is existential. A PkgInfo in `testing` **and** in `production`
matches a policy scoped on `testing`. The `["testing","canary"].containsAll(context.catalogNames)`
form in the docs is an allow list, and a scope cannot express it.

A `forbid` policy gives the same result. I verified this:

```
permit(principal in Role::"6", action == Monolith::Action::"viewPkgInfoData",
       resource in Monolith::Repository::"1");
forbid(principal, action == Monolith::Action::"viewPkgInfoData",
       resource in Monolith::Catalog::"9");
```

A PkgInfo in `[2,3,4]` is authorized. A PkgInfo in `[3,9]` is denied. There are no set
operators, and there is no list of catalog IDs in the policy.

The two forms behave differently when a catalog is added:

* With `containsAll`, a new catalog hides its PkgInfos until you edit every policy.
* With `forbid`, a new restricted catalog shows its PkgInfos until you add a `forbid`.

This is a choice to make. But note that the `contains()` / `containsAll()` examples also put
`catalogNames` into the policies, and that is
[bp-mutable-identifiers](https://docs.cedarpolicy.com/bestpractices/bp-mutable-identifiers.html):
a catalog that is renamed changes who can read the data, and nothing reports it. Entity IDs
are primary keys, and they do not change.

## Proposed shape

```python
REPOSITORY_RESOURCE_TYPE = ResourceType("Repository", get_namespace(), parents=(MBU_RESOURCE_TYPE,))
CATALOG_RESOURCE_TYPE = ResourceType("Catalog", get_namespace(), parents=(REPOSITORY_RESOURCE_TYPE,))
PKG_INFO_RESOURCE_TYPE = ResourceType("PkgInfo", get_namespace(),
                                      parents=(REPOSITORY_RESOURCE_TYPE, CATALOG_RESOURCE_TYPE))
```

`applies_to` then has an empty context, `ViewPkgInfoDataRequest` takes the PkgInfo, and
`add_view_data_pbac_requests` builds one request for each PkgInfo. It already reads the
catalogs and the repository that this needs.

`PkgInfo.objects.alles()` also rehydrates the catalogs with `select_related('repository')`
at
[`models.py:423`](https://github.com/zentralopensource/zentral/blob/c829a335/zentral/contrib/monolith/models.py#L423).
The list path can possibly use them, and do fewer queries. The catalog filter case in the
docstring stays an exception, because the aggregate only reports the catalog of the filter.

## Two other points

* The `views.py` diff contains a large import re-sort that is not related to the feature. It
  makes the five functional lines difficult to find. The same applies to the change in
  `tests/server_accounts/test_pbac_engine.py`.
* `_serialize_entity` at
  [`cedar.py:82`](https://github.com/zentralopensource/zentral/blob/c829a335/server/pbac/cedar.py#L82)
  keys the collected entities on `(entity.type, entity.id)`, and not on `full_type`. Two
  types with the same name in different namespaces collide, and one entity is discarded.
  This does not occur today, but this PR adds resource types. It is a separate fix.

PR #1666, opened 2026-08-29
santa: fix the target type of the cdhash event store redirect view

np5:
## What changed

`CDHashEventsStoreRedirectView` had `target_type = Target.Type.SIGNING_ID`. It is `Target.Type.CDHASH` now.

The two sibling views for the same target, `CDHashEventsView` and `FetchCDHashEventsView`, both use `CDHASH`, and this view reads its events with the `"file"` / `"cdhash"` pair.

## Why

`EventsMixin` builds every URL from `target_type`: `get_fetch_url`, `get_redirect_url`, `get_store_redirect_url` and the `target_url` context variable all interpolate `self.target_type.lower()`. With the wrong value, all of them resolve a `santa:signingid_*` name and carry a cdhash identifier.

`EventsStoreRedirectView.get()` calls `get_redirect_url()` when the selected store gives no event URL. A user who clicked a store link on the cdhash events page went to the Signing ID events page. That page searches on `apple_signing_id`, so it shows no event.

The events the view read were always the correct ones. They come from `object_key` and `identifier_key`, not from `target_type`.

## The other target views

I compared the class name, the `target_type`, the `object_key` and the `identifier_key` of each view against its siblings, for the seven target types. I also compared the `object_key` and `identifier_key` pairs against the linked object keys that the Santa events publish, and the `add_rule_link_key` and `add_rule_link_attr` of each `TargetView` against what `CreateConfigurationRuleView` reads. The cdhash view is the only one that was wrong.

## The test

`SantaTargetEventsViewsURLsTestCase` walks the Santa URL configuration for the `EventsMixin` subclasses and makes three checks:

- the `target_type` of each view agrees with the URL name that it is routed under;
- each native target type has its three event views, so a walk that finds nothing cannot pass;
- the URLs that `get_fetch_url`, `get_redirect_url` and `get_store_redirect_url` build, and the target detail URL, all resolve to a view of the same target.

With the bug put back, the two tests fail and name the class and the wrong page:

```
CDHashEventsStoreRedirectView SIGNINGID routed as santa:cdhash_events_store_redirect
CDHashEventsView.get_store_redirect_url links to a SIGNINGID page
```

## Tests

`tests.santa` passes: 804 tests. `ruff` is clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -193,6 +193,8 @@ Munki installs the Osquery enrollment package on a machine that has the Turbo ag
 
 The `rebuild_manifest_enrollment_packages` command takes a `--manifest` option with one or more manifest IDs now, to rebuild the enrollment packages of a part of the manifests only. It also takes a `--new-version` option, which increases the version of each enrollment package that it rebuilds, and gives it a new filename. The enrollment versions do not change with it, so the machines that are up to date do not install the packages again.
 
+Fixed the links of the Santa cdhash event store redirect, which had the Signing ID target type. The view sends the user back to the events page of the target when the store gives no event URL. It sent them to the Signing ID events page instead, with a cdhash identifier that this page cannot find. The events the view read were the correct ones.
+
 
 ## 2026.5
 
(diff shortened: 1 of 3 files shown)


soysauceconspiracy (approved):
LGTM

PR #1667, opened 2026-08-29
mdm: align the linked objects of the events with the models

np5:
## What changed

Three fixes to the objects that the MDM events and models link in `metadata.objects`. One commit for each.

### 1. The events use the key of the model

Some MDM events wrote a linked object under a hand-written key that did not match the key that `linked_objects_key()` derives from the model:

| Old key | New key | Model |
| --- | --- | --- |
| `mdm_artifactversion` | `mdm_artifact_version` | `ArtifactVersion` |
| `mdm_enrolleddevice` | `mdm_enrolled_device` | `EnrolledDevice` |
| `mdm_enrolleduser` | `mdm_enrolled_user` | `EnrolledUser` |
| `mdm_command` | `mdm_device_command` | `DeviceCommand` |
| `mdm_depenrollmentsession`, … | `mdm_dep_enrollment_session`, … | `DEPEnrollmentSession`, … |

The events that carry the new keys are `mdm_download`, `target_artifact_update`, `admin_password_updated`, `filevault_prk_updated`, `recovery_password_set`, `recovery_password_updated`, `recovery_password_cleared`, `device_lock_pin_set` and `device_lock_pin_cleared`.

`serialize_needles()` makes one needle for each key, `_o:<key>:<value>`. An old key put the events of one object in a second token: the `zentral_audit` events of that object used the key of the model, and these events used the other one. A search for the events of an artifact version found one half only.

No `EventsMixin` view exists for those objects today, so nothing queries either key yet. The split becomes visible as soon as something does.

### 2. A user command links itself by its uuid

`DeviceCommand` writes its own key in `linked_objects_keys_for_event()`, so an audit event links it by the uuid. The uuid is the identifier that the API and the web console show. `UserCommand` wrote no key, so `AuditEvent.build()` used the primary key instead: the needle of a user command held an auto-increment id, and the needle of a device command holds a CommandUUID.

Nothing builds an audit event for a user command today. The API lists them and reads them, and it does not delete them. No stored event changes.

### 3. The asset events carry the asset in the `asset` key

`BaseAppsBooksEvent.get_linked_objects_keys()` reads the asset from the `asset` key of the payload. Every apps & books event put it there, except `mdm_asset_created` and `mdm_asset_updated`. Those two were posted with a flat `asset.serialize_for_event()`: the attributes of the asset were at the top level, so the two events linked no object at all.

They carry an `asset` object now, like the location asset events and the notification events. `notification_id` keeps its place at the top level.

## Notes for the reviewer

- `mdm_command` was the worst of the set. It named no model at all, while `DeviceCommand` links itself under `mdm_device_command`. The four events that used it post from `Channel.DEVICE` commands only, so `mdm_device_command` is correct for all of them.
- The enrollment session key was built from `_meta.model_name`, which the token carries. It comes from the model now. `linked_objects_key()` accepts a class as well as an instance, which `tests/core_events/test_audit_event_objects.py` already relied on, so the lookup makes no throwaway instance.
- The token is signature-verified before that point, so the model name is not attacker-controlled. A model that is removed from the code, and whose content type is removed too, makes the session lookup raise `DoesNotExist` upstream. The key is then skipped and logged, and the `404` stands.
- The other thirteen descendants of `BaseAppsBooksEvent` were already correct. The seven notification events set `payload["asset"]` in the preprocessor. The location asset and the device assignment events serialize the location asset, which nests the asset. `DeviceAssignmentRequestEvent` and `DeviceAssignmentRequestErrorEvent` are registered, and no code posts them.
- The other hand-written keys agree with their model: `mdm_artifact`, `mdm_package`, `mdm_location`, `mdm_dep_enrollment`, `mdm_ota_enrollment`, `mdm_user_enrollment`, `mdm_dep_virtual_server`, `mdm_device_command`.
- No backfill. The events already in an event store keep the old key and the old payload, and they go away with the retention of the store. The CHANGELOG has an entry for each of the two incompatibilities.
- An asset keeps the `adamId` and pricing parameter pair as its only value. A primary key beside it was in an earlier version of this branch, and it is out. A view reads one key and one value — `fetch_object_events(key, val)` — so one needle must carry every event of an asset. The seven notification events hold the pair only, and they always will: Apple sends the notification before Zentral creates the asset. So the pair is the value that can cover all of them, and the primary key is not. The `zentral_audit` events of an asset are the ones to move, with a `linked_objects_keys_for_event()` on the model, and no audit event of an asset exists yet.

### The same defect in other apps

`linked_objects_key()` does not agree with the hand-written key in four more places. None of them splits anything today, because no audit event exists for those models:

| Hand-written | Model key | File |
| --- | --- | --- |
| `incident`, `machine_incident` | `incidents_incident`, `incidents_machine_incident` | `zentral/core/incidents/events/__init__.py` |
| `realm_user` | `realms_realm_user` | `zentral/contrib/santa/events/__init__.py` |
| `santa_ruleset` | `santa_rule_set` | `zentral/contrib/santa/events/__init__.py` |
| `munki_pkginfo`, `munki_pkginfo_name` | `monolith_pkg_info`, `monolith_pkg_info_name` | `zentral/contrib/munki/events/__init__.py` |

The munki pair carries a natural key, `(name, version)`, so it looks intentional. For the other three, a `linked_objects_key` attribute on the model is the cheaper fix: it moves the model to the event, and it breaks no stored event. `zentral/core/incidents/views.py` already queries `incident`, so that one has a consumer.

## Test

`tests/mdm/test_event_linked_objects_keys.py` writes no key spelling of its own. It builds the expected keys with `linked_objects_key()`, so the events and the models can only move together:

- `test_event_linked_objects_keys` pins 15 payload cases to the model classes they link. Two of them are real asset payloads: the notification, and the asset event, which carries a primary key that it does not link.
- `test_every_mdm_event_is_pinned` walks the events package and fails on an event that no case covers. A class needs a case where it writes the keys itself, and also where it changes what an inherited key function writes: the enrollment request events set `enrollment_payload_key` only, and their key is built from it, so a class that overrides an attribute of the class the key function belongs to counts too. The walk uses `walk_packages()`, and it reads the package itself first. `iter_modules()` does not descend into a sub-package, and neither function yields an `__init__` — where the other Zentral apps keep their event classes.
- `test_hand_written_keys_are_model_keys` parses the key functions with `ast` and rejects a written key that names no model in any app. It counts the written keys only: a subscript that is read is a payload lookup, and a dict literal counts where it maps a key to its values, `[(...)]`. A dict that maps a key to anything else is a payload dict — a `get()` default, a lookup table — and its keys are not object keys. The rule keeps the one real case, `mdm_dep_virtual_server`, which a key function returns as a dict literal.
- `test_unknown_enrollment_session_model` covers the lookup failure.

`tests/mdm/test_apps_books_assets_assignments_sync.py` asserts the keys that `mdm_asset_created` and `mdm_asset_updated` link, and not only the shape of the payload.

Mutation proves the guard. Four changes, all caught:

| Mutation | Test that fails |
| --- | --- |
| `mdm_artifact_version` back to `mdm_artifactversion` | `test_event_linked_objects_keys` and `test_hand_written_keys_are_model_keys` |
| a new event class, with a key that names no model, in `events/sub/inner.py` | `test_every_mdm_event_is_pinned` and `test_hand_written_keys_are_model_keys` |
| the same class in `events/__init__.py` | `test_every_mdm_event_is_pinned` and `test_hand_written_keys_are_model_keys` |
| a key written through another name, `out["mdm_usercommand"]` | `test_hand_written_keys_are_model_keys` |

The two walk mutations are the reason for that walk. With `iter_modules()` alone, both classes pass both guards.

The guard was measured for false alarms too. A `get()` default with contents, `self.payload.get("asset", {"adam_id": None, ...})`, is an ordinary line that writes no key, and it does not fail the test. A real bad key in the same function, `keys.update({"mdm_not_a_model": [(1,)]})`, still does.

A key that is composed at run time leaves no literal. `EnrollmentRequestEvent` builds `f"mdm_{self.enrollment_payload_key}"`, so a case is the only guard on those three keys — and `test_every_mdm_event_is_pinned` is what makes a fourth subclass ask for one. A subclass with `enrollment_payload_key = "not_a_model_at_all"` fails it.

## Verification

- `tests.mdm`, `tests.core_events` and `tests.core_stores`: 3237 tests, all pass, on the rebased branch.
- Patch coverage: no uncovered added line. `coverage json` missing lines ∩ `git diff -U0` hunks.
- `ruff` is clean on all the changed files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -125,6 +125,10 @@ A task that a device launched, and a task that Zentral launched before version 2
 
 The `UserActions` action groups contain the view actions of the models now. A role with a permit on `Action::"GlobalUserActions"`, or on the `UserActions` group of one namespace, reads the pages and the API endpoints of the models in that scope. Review your policies before you update.
 
+#### 🧨 MDM event object keys
+
+Some MDM events linked an object under a key that did not match the key of the model, so a search found one half of the events only: the `zentral_audit` events of the object always used the model key. `mdm_artifactversion` is `mdm_artifact_version` now, `mdm_enrolleddevice` is `mdm_enrolled_device`, `mdm_enrolleduser` is `mdm_enrolled_user`, `mdm_command` is `mdm_device_command`, and `mdm_depenrollmentsession` is `mdm_dep_enrollment_session`, with the other enrollment sessions. Update a probe or a query that filters on an old key. No backfill is necessary: the stored events keep the old key, and they disappear with the retention of the event store.
+
 
 ### Bug fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,6 +129,10 @@ The `UserActions` action groups contain the view actions of the models now. A ro
 
 Some MDM events linked an object under a key that did not match the key of the model, so a search found one half of the events only: the `zentral_audit` events of the object always used the model key. `mdm_artifactversion` is `mdm_artifact_version` now, `mdm_enrolleddevice` is `mdm_enrolled_device`, `mdm_enrolleduser` is `mdm_enrolled_user`, `mdm_command` is `mdm_device_command`, and `mdm_depenrollmentsession` is `mdm_dep_enrollment_session`, with the other enrollment sessions. Update a probe or a query that filters on an old key. No backfill is necessary: the stored events keep the old key, and they disappear with the retention of the event store.
 
+#### 🧨 MDM asset event payloads
+
+The `mdm_asset_created` and `mdm_asset_updated` events put the asset in an `asset` object now, like the other apps & books events. The attributes of the asset were at the top level of the payload, so the two events linked no object at all: `get_linked_objects_keys()` reads the asset from the `asset` key. Update a probe or a query that reads `adam_id`, `pricing_param`, `name` or another attribute of the asset at the top level of these two payloads.
+
 
 ### Bug fixes
 
--- a/tests/core_events/test_audit_event_objects.py
+++ b/tests/core_events/test_audit_event_objects.py
@@ -78,7 +78,7 @@ class RedundantSelfKeysTestCase(SimpleTestCase):
 
     A model that writes its own key with its own pk in linked_objects_keys_for_event() does the
     same work a second time. A model that writes its own key with something else keeps control of
-    it — the MDM device commands link themselves by uuid — so only the pk spelling is a defect.
+    it — the MDM commands link themselves by uuid — so only the pk spelling is a defect.
     """
 
     def iter_hooks(self):
--- a/tests/mdm/test_account_configuration_command.py
+++ b/tests/mdm/test_account_configuration_command.py
@@ -300,5 +300,5 @@ class AccountConfigurationCommandTestCase(TestCase):
         )
         metadata = event.metadata.serialize()
         self.assertEqual(metadata["machine_serial_number"], enrolled_device.serial_number)
-        self.assertEqual(metadata["objects"], {"mdm_command": [str(cmd.uuid)]})
+        self.assertEqual(metadata["objects"], {"mdm_device_command": [str(cmd.uuid)]})
         self.assertEqual(set(metadata["tags"]), {"mdm", "admin_password"})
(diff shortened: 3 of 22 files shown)


soysauceconspiracy (approved):
LGTM

PR #1668, opened 2026-08-30
accounts, monolith: fix the loop variable of the deferred callbacks

np5:
## Summary

This PR takes the fix from #1617, adds the regression tests it needs, and removes one of the two closures. The first commit is the commit of @harshadkhetpal, with the original author.

Two `transaction.on_commit()` callbacks were defined in a loop, and they read the loop variable when they ran, not when they were defined. Django defers a callback to the commit of the outermost atomic block, so a callback can run after the loop, when the loop variable holds the value of the last iteration.

**`sync_monolith_repositories`**: the `transaction.atomic()` block contains the full loop, so all the callbacks run after the loop. A command run with N repositories sent N notifications, and each one carried the primary key of the last repository. The only subscriber of the `monolith.repository` channel is `MonolithConf._reset_repositories()`. It ignores the payload and clears the full cache, so no repository cache became stale. The notifications were wrong, and nothing read them. A default argument binds the primary key at definition time.

**`remove_expired_api_tokens`**: the `transaction.atomic()` block is in the loop, so each callback runs at the end of its own iteration. A standalone command run is correct. In an outer transaction, all the callbacks run at the end, and the audit event of the last token is posted N times.

The third commit removes the closure of this second command. `AuditEvent.post` takes no argument, and a bound method carries its own instance, so `transaction.on_commit(event.post)` defers the event of the current token. Nothing reads the loop variable anymore, so there is no binding left to get wrong.

The `sync_monolith_repositories` command keeps its closure. Its callback needs the primary key of the repository, so there is no bound method to give to `transaction.on_commit()`, and the value must be captured at definition time.

## Tests

The two command test suites built one repository and one token only, so neither one could show the problem.

This PR adds one test to each suite with two objects:

- `test_sync_two_repositories_ok` verifies that each notification carries the primary key of its own repository.
- `test_remove_two_tokens` verifies that each audit event carries its own token.

Both tests fail on the two commands before the fix, and pass after it. `tests.monolith` and `tests.server_accounts` are green (1112 tests).

The monolith test reads the order of the sync from the queryset that the command iterates. The repository names are random, and the `en_US.utf8` collation of the database does not order them like python does. A sort in python made the test fail 3 times out of 12.

No CHANGELOG entry: the behavior of Zentral does not change.

### DIFF
--- a/server/accounts/management/commands/remove_expired_api_tokens.py
+++ b/server/accounts/management/commands/remove_expired_api_tokens.py
@@ -111,9 +111,6 @@ class Command(BaseCommand):
 
                     token.delete()
 
-                    def on_commit_callback(event=event):
-                        event.post()
-
-                    transaction.on_commit(on_commit_callback)
+                    transaction.on_commit(event.post)
         finally:
             queues.stop()
--- a/server/accounts/management/commands/remove_expired_api_tokens.py
+++ b/server/accounts/management/commands/remove_expired_api_tokens.py
@@ -111,7 +111,7 @@ class Command(BaseCommand):
 
                     token.delete()
 
-                    def on_commit_callback():
+                    def on_commit_callback(event=event):
                         event.post()
 
                     transaction.on_commit(on_commit_callback)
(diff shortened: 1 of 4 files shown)


soysauceconspiracy (approved):
LGTM

PR #1669, opened 2026-08-30
monolith: sync each repository in its own transaction

np5:
## Summary

Follow-up of #1668. The `sync_monolith_repositories` command wrapped the full loop in one `transaction.atomic()` block, and caught the errors inside that block. This gives the command two failure modes.

**A database error stops the command.** The `except Exception` block looks like it isolates each repository, but it only does that for the errors that do not touch the database. A database error puts the connection in an aborted state. The command catches it, writes to stderr, and continues, and then the advisory lock of the next repository fails with `current transaction is aborted, commands ignored until end of transaction block`. That exception is outside the `try`, so the command stops. The repositories that the command synchronized before the error are rolled back with it. One bad catalog can stop the synchronization of every repository.

**An error keeps the partial changes of the repository that failed.** The `except` block catches the error inside the atomic block, so `__exit__` never rolls it back. `BaseRepository.sync_catalogs()` imports the pkg infos, archives the pkg infos and the catalogs, and then reads the client resources. The S3 backend raises a `RepositoryError` on that read, so the archived pkg infos stay, the synchronization date does not change, and the manifests do not get their new version.

## Changes

Each repository gets its own transaction, and the `try` block is outside of it, like in `sync_repository_task`:

- An exception reaches `__exit__`, which rolls that repository back. The command synchronizes the next repository with a healthy connection.
- The advisory lock is transaction scoped, so it is released at the end of the synchronization of its own repository. Before, the command held the locks of all the repositories until the end of the run, and an operator who used the *Sync* button on a repository the command was done with got "a sync is already running".
- The audit events of a repository are published when that repository commits, and not at the end of the command.

The notification callback carries its arguments in a `functools.partial`, so nothing reads the loop variable when the callback runs. The callback of #1668 used a default argument for this. The default argument is not necessary with one transaction for each repository, but only when Zentral runs the command outside of a transaction. A `partial` is correct in the two cases.

## Tests

Two new tests, both fail before this change:

- `test_sync_error_rolls_the_repository_back`: the sync of a repository writes a catalog and raises. The catalog must not exist after the command. Before: `AssertionError: True is not false`.
- `test_sync_database_error_on_first_repository`: the sync of the first of two repositories makes a database error. The second repository must be synchronized, and get its notification. Before: `InternalError: current transaction is aborted, commands ignored until end of transaction block`.

`tests.monolith` and `tests.server_accounts` are green (1114 tests). The command test file passed 10 runs out of 10.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -209,6 +209,8 @@ The `rebuild_manifest_enrollment_packages` command takes a `--manifest` option w
 
 Fixed the links of the Santa cdhash event store redirect, which had the Signing ID target type. The view sends the user back to the events page of the target when the store gives no event URL. It sent them to the Signing ID events page instead, with a cdhash identifier that this page cannot find. The events the view read were the correct ones.
 
+The `sync_monolith_repositories` command uses one transaction for each repository now, and releases the lock of a repository at the end of its own synchronization. With one transaction for all of them, a database error stopped the command and rolled back the repositories already synchronized, and a failed synchronization kept its partial changes.
+
 
 ## 2026.5
 
--- a/tests/monolith/test_sync_monolith_repositories_cmd.py
+++ b/tests/monolith/test_sync_monolith_repositories_cmd.py
@@ -1,9 +1,10 @@
 from io import StringIO
 from unittest.mock import call, patch
 from django.core.management import call_command
-from django.db import connections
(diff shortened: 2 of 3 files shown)


soysauceconspiracy (approved):
LGTM

PR #1675, opened 2026-09-01
Drop autoprefixer and postcss-loader

np5:
## Summary

This PR removes `autoprefixer`, `postcss` and `postcss-loader` from the frontend build. It also removes the `postcss-loader` rule from `webpack.config.js`.

Autoprefixer added no vendor prefix to the compiled CSS. Bootstrap and bootstrap-icons write every prefix that Zentral uses in their own SCSS sources.

## Evidence

The built `server/static/dist/main.css` contains 30 different vendor-prefixed properties. The SCSS sources contain the same 30. The difference between the two sets is empty.

To confirm this, build `main` and this branch in the same container, then compare the two results:

| output | result |
| --- | --- |
| `main.css` | identical, 311697 bytes |
| `user_portal.css` | identical, 1528 bytes |
| `main.js`, `theme.js`, `user_portal.js` | identical |
| full `server/static/dist/` tree | identical |

The build gives the same bytes with autoprefixer and without it.

## Effect on the lockfile

20 packages leave `package-lock.json`: `autoprefixer`, `postcss-loader`, the `cosmiconfig` configuration loader that `postcss-loader` uses, and their dependencies. No package is added. No package that stays changes its version.

`postcss` stays in the tree, because `css-loader` depends on it. `browserslist` and `caniuse-lite` also stay, because `webpack` depends on them. Only the removal of webpack can remove those.

## The browserslist commit

The second commit updates the browserslist data. It changes six lockfile entries: `baseline-browser-mapping`, `browserslist`, `caniuse-lite`, `electron-to-chromium`, `node-releases` and `update-browserslist-db`. No package is added or removed.

The update changes no built byte. The repository has no `.browserslistrc` and no `browserslist` key in `package.json`, so no tool reads the browserslist data during the build.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,38 +12,12 @@
         "chart.js": "^4.5.1"
       },
       "devDependencies": {
-        "autoprefixer": "^10.4.14",
         "css-loader": "^6.8.1",
         "mini-css-extract-plugin": "^2.7.6",
-        "postcss": "^8.5.25",
-        "postcss-loader": "^7.3.3",
         "sass": "^1.63.6",
         "sass-loader": "^13.3.2",
         "webpack": "^5.88.1",
         "webpack-cli": "^5.1.4"
-      }
-    },
-    "node_modules/@babel/code-frame": {
-      "version": "7.26.2",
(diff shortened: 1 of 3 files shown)


soysauceconspiracy (approved):
LGTM

PR #1678, opened 2026-09-02
turbo: record the run of a results entry that is set aside

np5:
The Turbo results endpoint discards the outcome of an entry that it cannot use. It discarded the run with it.

`ResultsView._process` leaves the loop with `continue` on a kind mismatch. This skipped `_record_job_status` together with the scoring and the tagging. `last_result_at` on the per-machine row is what stops the configuration endpoint from serving a one-time job again. A one-time job with results that the server always set aside was therefore served again after every configuration refresh. It ran, and the server discarded the outcome again. There is no end to this, because `not_after` is optional.

The cost is milliseconds for a script. It becomes a repeated collection when command kinds arrive, so it is better to fix it before them.

### Changes

- `_record_job_status` becomes the public `record_run(job_machine, job, version, ran_at)`. `ResultsBatch.add` calls it, and the view calls it directly before it sets an entry aside.
- `unknown_schedule` still discards everything. There is no row to record against, and the job is not being served either.
- The results request event carries a `skipped_counts` map. An entry that the server sets aside makes no `turbo_result` event, so a shot that the server consumed and discarded left nothing in the event stream to find it by.

### Behavior change

A one-time job with a systematically unusable result stops after one attempt now. The operator sees the failure and schedules a new `OneTimeJob`. This is the retry model that the endpoint already documents for a failed upload.

### Verification

- Full suite: 8398 tests, all pass.
- Six new tests. Two of them separate the two rules that are easy to confuse: a mismatched one-time entry closes the gate, and a stale-version entry does not, whether the server sets it aside or not.
- Every line of the change is covered. The one uncovered line in the touched files is in `_prune_cc_out_of_scope` and is not part of this diff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -226,6 +226,8 @@ Fixed the links of the Santa cdhash event store redirect, which had the Signing 
 
 The `sync_monolith_repositories` command uses one transaction for each repository now, and releases the lock of a repository at the end of its own synchronization. With one transaction for all of them, a database error stopped the command and rolled back the repositories already synchronized, and a failed synchronization kept its partial changes.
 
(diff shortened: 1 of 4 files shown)


soysauceconspiracy (approved):
LGTM

PR #1680, opened 2026-09-02
turbo: add commands, the third kind of job

np5:
A command collects something from a machine instead of producing a verdict. Two kinds land with it: sysdiagnose, which takes no options, and file_export, which takes a list of absolute path patterns and an uncompressed max_size.

One Command model on BackendInstance with a backend registry, not one model per kind. Script and MSCPCheck each earn a table because Postgres enforces their identity — FKs with DB semantics, typed ODV columns and their constraints. A command has neither: its variation is agent behaviour, which is Python. That is the split stores.Store and probes.Action already use, and a command that grows an FK graduates to its own definition model under the same Job anchor.

Job.kind grows one value per backend, flat, so a new command is a config change for an agent rather than a protocol change: an agent already drops a kind it does not know, element by element. Job.definition returns None for a kind the release does not know, and the config endpoint skips such a job instead of dereferencing it — during a rolling deploy an older instance can read a row a newer one wrote, and every machine holding that job would otherwise get a 500. Job.definition_relations() replaces the eleven select_related sites that listed the definition relations by hand.

A command runs one time only. A backend declares its allowed_modes, the recurring form narrows its job choices to the kinds that accept a recurring schedule, and the recurring serializer says so. A recurring collection is a log shipper, not a job.

A backend also declares the artifacts a run produces — name, filename stem, extension, content type, and whether it is optional. file_export declares two: a required manifest of what was collected, and an optional archive, absent when a run matched nothing. The declaration is the server's, never the agent's.

The console is read-only for commands, like it is for stores and probe actions: the API validates the per-backend options, and the detail page exists because every job list links a definition's get_absolute_url.

New Turbo::Action::"scheduleCommand", the first typed PBAC action in the module. It carries the kind and, on the machine flow, the machine — so a policy can permit a sysdiagnose and refuse a file_export, which one turbo.add_onetimejob cannot express. It is an additional gate rather than a replacement: the configuration flow targets a scope expression with no machine to name, so a machine-keyed policy applies to one path and not the other, and replacing the permission would make that path look covered. The plain CRUD actions stay on the legacy-perm path, where the console gates its buttons.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -317,6 +317,14 @@ Recurring and one-time job schedules on a job anchor, with a per-machine result 
 
 Management UI and API for the configurations, enrollments, scripts, mSCP checks and jobs, the agent protocol endpoints, the script and mSCP compliance checks, and the agent request and result events.
 
+New **commands**, the third kind of job: a command collects something from a machine instead of producing a verdict. Two kinds are available. `sysdiagnose` collects a `sysdiagnose` archive and takes no options. `file_export` collects the files that match a list of path patterns, plus a manifest of what it collected, and takes `patterns` and an uncompressed `max_size`.
+
+A command is created with the `/api/turbo/commands/` endpoint, and *Turbo > Commands* lists them. The kind cannot be changed after the command is created. A change to the options bumps the version of the job, and the agent runs the command again.
+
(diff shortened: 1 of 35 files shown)


np5:
superseded by https://github.com/zentralopensource/zentral/pull/1682

PR #1682, opened 2026-09-02
turbo: add commands, the third kind of job

np5:
Commands are the third kind of Turbo job. A command collects something from a machine instead of producing a verdict. Two kinds land with it:

* `sysdiagnose` collects a `sysdiagnose` archive and takes no options.
* `file_export` collects the files that match a list of absolute path patterns, plus a manifest of what it collected. It takes `patterns` and an uncompressed `max_size`.

This is the definition and scheduling half. The upload of the collected files comes in the next change.

## Reading it

Three commits, each of which stands on its own — 1 and 2 were checked out and tested in isolation, not just at the tip:

| | Commit | Files | |
|---|---|---|---|
| **1** | `pbac: partial evaluation, and entities that a policy can actually reach` | 5 | Shared `server/pbac/`. **Worth the hardest look**: it changes authorization for every module, and it is where four silent bugs were. 979 tests OK on its own. |
| **2** | `turbo: add commands, the third kind of job` | 35 | The definition kind, its backends, the wire, API and console. No PBAC. 1047 tests OK on its own. |
| **3** | `turbo: type the one-time job actions, and retire their permissions` | 21 | The typed actions, the permission retirement, and the view / template / API rewiring. |

The [review of 2026-09-03](https://github.com/zentralopensource/zentral/pull/1682#pullrequestreview-5101479723) is answered in place: each fix is folded into the commit that owns the code, so there is no separate *address the review* commit to read against the other three. Every thread has a reply. Two of them changed the shape of the fix and are worth reading rather than trusting — the rolling-deploy guard and the picker narrowing, both marked below.

---

# 1 — `pbac: partial evaluation, and entities that a policy can actually reach`

Three additions to the engine. Nothing that exists today exercises any of them, which is why the bugs below were invisible.

## Partial evaluation, so a context attribute can stay required

A `Request` carries `unknown_context`, set by a caller that cannot fill the context in yet — a view deciding whether to offer an action before the user has picked the object that would complete it. `Engine.authorize_request` routes those to cedarpy's `is_authorized_partial`, which residualizes what a policy reads instead of failing on a record that does not have it. `Deny` is the one definitive answer: no context could make the request pass.

The routing lives on the request, so `PBACViewMixin`, `PBACPermission` and every check helper get the right evaluator without knowing there are two — and a policy writer never learns which one ran.

The alternative was to declare every such attribute optional and write `!(context has x) || …` in every policy that reads one. That fails **asymmetrically**: a `permit` missing the guard is skipped and the action vanishes for the roles it grants, while a `forbid` missing it is skipped and decides correctly anyway — so the footgun only bites the policy an operator writes first.

`{}` is not the same as absent: an empty context is a record with no attributes and errors in *both* evaluators, which is why a preview serializes its context as `None`. `authorize_requests` splits a batch, since partial evaluation has no batch form and needs none.

## Four latent bugs, all of which failed silently

An attribute can be typed as an entity, and an entity can be named in a context. Both were expressible before and neither worked:

* `register_entity_type` auto-registered parents but not the types an attribute refers to. Cedar then answers `failed to resolve type`, which fails the **whole** schema — every policy save in the deployment, not just the action that declared it. Six existing accounts tests caught this one.
* `_serialize_entity` copied attributes verbatim, so an entity-valued attribute reached Cedar as a reference to something absent: `resource.x.y` evaluated to `entity … does not exist` and a `forbid` **failed open**, with nothing but a diagnostic.
* `_serialize_requests_entities` walked only the principal, the action and the resource, so an entity named in a **context** never reached the slice either — the same failure, on the path commit 3 depends on.
* `register_action` registered the types of the principals and the resources an action applies to, but not the ones named in its **context**. Same failure as the first bullet, same blast radius, reached by the other door. Found in review — it did not bite here only because `Turbo::Job` is also reachable through `OneTimeJob.job`.

All of them now recurse through `SetOf`, `RecordOf`, dicts and lists. The registry and the collector are populated before recursing, so a self-referential attribute or a cycle between two entities terminates.

---

# 2 — `turbo: add commands, the third kind of job`

## One model, one registry

`Command` is a `BackendInstance` with a backend registry, not one model for each kind. `Script` and `MSCPCheck` each earn a table because Postgres enforces their identity: foreign keys with database semantics, typed ODV columns and their constraints. A command has neither, and its variation is agent behaviour, which is Python. This is the split that `stores.Store` and `probes.Action` already use. A command that grows a foreign key graduates to its own definition model under the same `Job` anchor.

`Job.kind` grows one value for each backend, flat. A new command is then a configuration change for an agent and not a protocol change, because an agent already drops a kind it does not know, element by element.

## Rolling deploys — **changed in review**

`Job.definition` returns `None` for a kind the release does not know. During a rolling deploy an older instance reads rows a newer one wrote, so every endpoint an unknown kind can reach has to hold that shape. The review found the guard covered `/config/` only; it now covers all three, and the wire no longer throws the report away.

* **config** skips the job. Without it, every machine holding that job gets a `500` until the refresh completes.
* **results** sets the entry aside as `unknown_job_kind` and still records the run — the same reason `kind_mismatch` does: the run happened, only the outcome is unusable. `definition_wire_ref()` read `.pk` on `None` before, which `500`d the whole batch, taking every valid entry in it with it, on every retry.
* **status** writes the tracker *first* and only then skips the ref. Skipping the entry outright would stop the tracker update, and the next `cleanup` sweep would mark a job the agent is still holding as removed — worse than a missing acknowledgment.
* **the wire** takes `kind` as a short string, not a `ChoiceField` over the kinds *this* release knows. During a rollout the instances do not share a set of kinds, and a report of a run that happened is not a malformed entry: rejecting it drops the entry whole, so the agent gets no acknowledgment to read and the shot never closes. The resolved job decides, which is where the authority already was — results only compares the wire kind against it, and status never reads it at all.

Worth noting how the two halves hid each other: with the `ChoiceField` in place, an entry *naming* an unknown kind never reached the view, so my first tests for the guard passed for the wrong reason. The reachable path was an entry with no `kind` at all. Both shapes are pinned now.

`Job.definition_relations()` replaces the eleven `select_related` sites that listed the definition relations by hand.

## One time only

A command cannot be attached to a recurring job. Collecting the same files every hour is a log shipper, not a job. A backend declares its `allowed_modes`, the recurring form narrows its job choices to the kinds that accept a recurring schedule, and the recurring serializer gives an error.

## fnmatch patterns

The agent matches the `file_export` patterns with `fnmatch(3)`. macOS gives it that for free.

`fnmatch` has no recursive wildcard, and it does not refuse `**` — it reads two stars as one. `/Library/Logs/**/*.log` matches `/Library/Logs/a/b.log`, and it does not match `/Library/Logs/a/b/c.log`. A pattern that an operator writes for recursion collects one directory level and gives no error.

The server refuses a pattern that contains `**`. An operator names each level instead:

```
/Library/Logs/Foo/*.log
/Library/Logs/Foo/*/*.log
```

The pattern cap is 32, because one tree now costs one pattern for each level.

## Declared artifacts

A backend declares the artifacts a run produces: name, filename stem, extension, content type, and if it is optional. `file_export` declares two, a required manifest and an optional archive that a zero-match run omits. The declaration belongs to the server, never to the agent. The upload plane validates against it in the next change.

## Read-only console

*Turbo > Commands* lists the commands and shows what each one collects. A command is created with the API, like an event store or a probe action, because the API validates the options of each kind. The detail page exists because every job list links the `get_absolute_url` of a definition.

---

# 3 — `turbo: type the one-time job actions, and retire their permissions`

`Turbo::Action::"createOneTimeJob"`, `Turbo::Action::"updateOneTimeJob"` and `Turbo::Action::"deleteOneTimeJob"` become **typed** PBAC actions.

|Action|Resource|
|---|---|
|`createOneTimeJob`|The **configuration** — the schedule does not exist yet.|
|`updateOneTimeJob`|The **schedule**.|
|`deleteOneTimeJob`|The **schedule**.|

A schedule is a member of its configuration, so one configuration-scoped policy covers all three, and it carries the job it runs, so an update or delete policy can read `resource.job`. All three carry the job as an **entity** in the context, which is the only thing all three share — create has no row to read it from:

```
forbid (
  principal,
  action in [Turbo::Action::"createOneTimeJob",
             Turbo::Action::"updateOneTimeJob",
             Turbo::Action::"deleteOneTimeJob"],
  resource
) when { context.job.kind == "file_export" };
```

An entity rather than a kind string because it composes: `context.job.kind` says what a `job_kind` key said, and anything else the job comes to declare is readable without reshaping every action that carries it. The closed set of kinds is not lost — the `values` list is declared on `Turbo::Job`'s own `kind` attribute, where the schema browser still shows it.

**`context.job` is required, and a policy reading it needs no guard** — that is what commit 1 buys. This grant works on its own:

```
permit (
  principal in Role::"<role id>",
  action == Turbo::Action::"createOneTimeJob",
  resource
) when { context.job.kind == "sysdiagnose" };
```

It schedules a `sysdiagnose`, refuses a `file_export`, and still offers the action before a job has been picked — no `has` guard, no companion `forbid`. A test pins exactly that, in the console and through the API. (The review caught the earlier version of that test passing for the wrong reason: it pinned a `Turbo::JobKind` entity that no longer exists, which only stayed in the fixture because `set_policy` bypasses `Policy.clean()`.)

**The permissions are retired.** `turbo.add_onetimejob`, `turbo.change_onetimejob` and `turbo.delete_onetimejob` map to no action, so nothing can ask these actions with `System::"any"` and an empty context: every request carries a resource. There is a `🧨` changelog entry. A converted `Role <name>` policy keeps working, because it names the actions and leaves the resource open — so no role loses access it had.

**`viewOneTimeJob` stays legacy-mapped, deliberately.** A typed view action would have to scope a *list*, which means filtering a queryset by policy rather than deciding one request — a separate piece of work. A test asserts it stays that way so nobody types it by halves.

Consequences worth reviewing:

* The console gates its buttons on view-computed booleans instead of `perms.turbo.*`. The per-row buttons differ by job, so `authorize_one_time_job_rows` decides `can_update` and `can_delete` per row and batches them through `engine.authorize_requests`. That batching hands `_object_jobs.html` a **list** where it used to get a queryset, so the heading counts with `|length` — a list has no argument-free `count()`, and the template engine renders the empty string for a call that needs an argument instead of raising. Caught on the command page after the review, where the heading read `One-time job ()`; it was all three definition pages, since they share the mixin. The recurring heading keeps `.count`, since that context is still a queryset.
* The API empties `perms_map` for `POST`, `PUT` and `DELETE` and decides in `perform_create` / `perform_update` / `perform_destroy`, on validated objects rather than primary keys off the wire.
* The shared base views expose the configuration as a `cached_property`, so an authorization mixin can use it as its resource without any code reading the database before the mixin runs. Resolving it eagerly — which is what the first version of this commit did — turned an anonymous request for an unknown pk into a `404` instead of the login redirect. They no longer carry `PermissionRequiredMixin`; each concrete view declares its own gate, which is why the recurring views change too.
* **The pickers offer only the kinds a policy would permit — changed in review.** A kind-scoped grant used to walk an operator into a `403` with the form lost. Only the **choices** are narrowed, not the field's queryset: narrowing the queryset moves the gate from authorization to form validation, where a refused kind comes back as *select a valid choice* without ever reaching the typed check, so nothing is logged as a permission denial. I tried it that way first and two tests asserting `403` started returning `200`, which is how I noticed. A forged post still reaches the check and is refused there; those two tests are unchanged, which is the proof. One batched evaluation per render, and on an update the field keeps its current value even if a policy would refuse that kind today.
* Scheduling from a machine page also requires `turbo.view_enrolledmachine`. Its configuration comes from the enrolled machine, so it is only known once the serial resolves and cannot be the gate that runs first — and a 404 against a 403 must not reveal which serials are enrolled.
* Every kind is covered, not only the commands. A script runs arbitrary code as root, and `turbo.add_onetimejob` never distinguished it from an mSCP check either.

The plain CRUD of a command stays on the legacy permission path: a global definition list has no resource to scope by.

**The mechanism is here, the policies are not.** Until someone writes one, a role that holds the Turbo admin action group can schedule a `file_export` against any machine, as before.

---

## Builds on

zentralopensource/zentral#1678, already in `main`, which records the run of a results entry that is set aside. Without it, a one-time command whose results are systematically unusable would be collected and discarded on every configuration refresh.

## Tests

The full suite passes: 8552 tests. Every added line is covered (495 executable added lines, 0 uncovered). Commits 1 and 2 were checked out and run on their own after the fold, not only at the tip: 979 and 1047 tests.


server/pbac/engine.py
```diff
@@ -70,20 +73,41 @@ def __init__(self) -> None:
...
         key = (et.name, et.namespace.id if et.namespace else None)
         existing = self.entity_types.get(key)
         if existing is None:
+            # in the registry before recursing, so a type whose attribute points back at it — or at
+            # itself — terminates
             self.entity_types[key] = et
             for parent in et.parents:
                 self.register_entity_type(parent)
+            for attr in et.attrs.values():
+                self._register_attr_entity_types(attr.type)
```
np5, line 110: **issue:** `register_action` (lines 218-219, unchanged) still registers the types of `applies_to.principals` and `applies_to.resources` only. An entity type that is named only in an action's `context` never reaches the schema, and the schema then fails for every policy. That is the same class of bug this commit closes for attributes and parents.

It does not bite in this PR because `Turbo::Job` is also reachable through `OneTimeJob.job`. Checked on this branch in a scratch container: an action with `context={"widget": AttrSpec(Widget)}` and no other path to `Widget` makes `validate_policies` fail for an unrelated `permit` with `Schema parse error: failed to resolve type: Probe::Widget`. With `Widget` reachable through a resource attribute, the same policy validates.

Suggested fix in `register_action`, with a test next to `test_register_entity_type_registers_an_entity_typed_attr`:
```python
for et in (*applies_to.principals, *applies_to.resources):
    self.register_entity_type(et)
for attr in (applies_to.context or {}).values():
    self._register_attr_entity_types(attr.type)
```

CHANGELOG.md
```diff
@@ -317,10 +317,26 @@ Recurring and one-time job schedules on a job anchor, with a per-machine result
 
 Management UI and API for the configurations, enrollments, scripts, mSCP checks and jobs, the agent protocol endpoints, the script and mSCP compliance checks, and the agent request and result events.
 
+New **commands**, the third kind of job: a command collects something from a machine instead of producing a verdict. Two kinds are available. `sysdiagnose` collects a `sysdiagnose` archive and takes no options. `file_export` collects the files that match a list of path patterns, plus a manifest of what it collected, and takes `patterns` and an uncompressed `max_size`.
```
np5: **issue:** These paragraphs, and the 🧨 entry at line 332, are in the `## 2026.5` section. 2026.5 is tagged (`v2026.5` on 2026-08-18, `v2026.5.4` on 2026-08-26) and both tags carry Turbo. The commands and the permission retirement ship in 2026.6, so they belong under `## 2026.6`: a `#### Turbo` subsection under *Features*, and the 🧨 under its *Backward incompatibilities*. As written, an operator who upgrades from 2026.5.x and reads the 2026.6 notes does not see the retirement.

zentral/contrib/turbo/forms.py
```diff
@@ -299,7 +335,7 @@ class OneTimeJobSearchForm(forms.Form):
     def get_queryset(self):
         qs = (
             OneTimeJob.objects
-            .select_related("configuration", "job__script", "job__mscp_check")
+            .select_related("configuration", *Job.definition_relations("job__"))
```
np5, line 356: **issue:** The `q` filter below (lines 344-348, label *Job or configuration name*) still matches `job__script__name` and `job__mscp_check__rule_id` only. A one-time job that schedules a command cannot be found by the command's name: the list shows it, the search does not.

```python
| Q(job__command__name__icontains=q)
```

`RecurringJobSearchForm` has the same shape, but no command can be recurring, so only this one matters today.

zentral/contrib/turbo/public_views/config.py
```diff
@@ -14,10 +14,17 @@
...
-    def _wire(job, schedule):
+    def _wire(self, job, schedule):
+        # a kind this release does not know has no definition to serve — during a rolling deploy an
+        # older instance can read a row a newer one wrote. Skip it instead of dereferencing None, or
+        # every machine holding that job gets a 500 here until the refresh finishes.
+        definition = job.definition
+        if definition is None:
+            logger.error("Turbo config for %s: unknown job kind %r, skipping job %s",
+                         self.serial_number, job.kind, job.pk)
+            return None
```
np5, line 25: **suggestion:** This guard covers `/config/` only. The results and status endpoints still dereference the definition of a job they do not know: `_wire_ref` (results.py:102) and status.py:62 call `job.definition_wire_ref()`, which reads `definition.pk` on `None`. That is an `AttributeError`, so a 500 for the whole batch.

The sequence is the one this comment describes, one step later: the machine takes its config from a new instance, runs the job, and posts the results or the status to an old instance. The agent retries, so nothing is lost, but every valid entry in that batch waits with it, with a traceback on each attempt, until the refresh completes.

A `definition is None` check next to the `kind_mismatch` branch would close it. `raise WireError("unknown_job_kind")` keeps the retry (the agent retries a 4xx too) without the traceback. Setting the entry aside with `record_run` consumes the shot instead. Either is fine, the crash is not.

np5 (commented):
Read all three commits with the full files at the head. Ran the docs build (`--strict`, clean), ruff (clean, apart from an unused import that is already on main), and a scratch-container check of the schema behaviour behind the first finding.

Mergeable once two things are fixed: the engine gap and the changelog placement. The rest can follow.

- `server/pbac/engine.py`: `register_action` does not register the entity types named in an action's `context`. The bug class commit 1 closes for attributes stays open for contexts. It is masked here only because `Turbo::Job` is also reachable through `OneTimeJob.job`. Verified on the branch, details inline.
- `CHANGELOG.md`: the entries are in the closed 2026.5 section. They belong under 2026.6.
- `zentral/contrib/turbo/forms.py`: the one-time job search does not match a command's name.
- `tests/turbo/test_api_one_time_jobs.py`: one test and three comment blocks describe the earlier `JobKind` design and say the opposite of the PR's headline claim.
- `zentral/contrib/turbo/public_views/config.py`: the rolling-deploy guard covers config only. Results and status still 500 on a kind they do not know. Bounded to the refresh window and self-healing, so a suggestion.

What looked good: the three engine fixes are real and each is pinned by a test; routing the preview on the request keeps every caller ignorant of the two evaluators; `Job.definition_relations()` and the `allowed_modes` declaration remove two classes of forgotten-site bugs.

tests/turbo/test_api_one_time_jobs.py
```diff
@@ -140,7 +143,93 @@ def test_update_one_time_job_job_immutable(self):
...
+        # System as its resource, so a permit scoped to a JobKind does not authorize it, and the
+        # refusal lands in the DRF permission layer before the typed check is even reached.
+        configuration = force_configuration()
+        command = force_command(backend=CommandBackend.SYSDIAGNOSE)
+        self.set_policy("permit ("
+                        f' principal in Role::"{self.group.pk}",'
+                        ' action == Turbo::Action::"createOneTimeJob",'
+                        ' resource in Turbo::JobKind::"sysdiagnose"'
+                        ");\n")
+        response = self.post(reverse("turbo_api:one_time_jobs"),
```
np5, line 237: **issue:** This test pins a design that is gone. `Turbo::JobKind` is not an entity type in the schema. `Policy.clean()` would reject this source, and the test only stores it because `set_policy` bypasses `clean()`. The mechanism in the comment is gone too: `POST` has an empty `perms_map` now, so the 403 comes from `perform_create`, not from the DRF permission layer. The test passes because no permit matches, whatever the policy says.

The API counterpart of `test_a_kind_scoped_permit_needs_no_guard_and_no_forbid` would pin the real behaviour, and the headline claim of the PR: a `permit … when { context.job.kind == "sysdiagnose" }` alone gives 201 for a sysdiagnose and 403 for a file_export.

tests/turbo/test_api_one_time_jobs.py
```diff
@@ -140,7 +143,93 @@ def test_update_one_time_job_job_immutable(self):
...
         response = self.delete(reverse("turbo_api:one_time_job", args=(pk,)))
         self.assertEqual(response.status_code, 204)
         self.assertFalse(OneTimeJob.objects.filter(pk=pk).exists())
+
+    # PBAC: the typed createOneTimeJob / updateOneTimeJob, which the legacy permission alone cannot
+    # express. This path had no typed check at all before.
+    #
+    # A kind is refused with a FORBID on top of the broad grant, not with a kind-scoped permit: the
+    # same actions are still asked resourcelessly through the legacy permission path (a template
+    # deciding whether to render a link), and a kind-scoped permit does not answer that question.
```
np5: **nit:** This block says a kind-scoped permit *cannot* answer the question, and that the actions are still asked through the legacy path. Both are the opposite of what this commit does: the actions are unmapped, and `test_a_kind_scoped_permit_needs_no_guard_and_no_forbid` pins the permit-only grant. Stale from an earlier iteration, together with the test at line 218.

tests/turbo/test_setup_enrolled_machines.py
```diff
@@ -204,3 +218,33 @@ def test_schedule_machine_one_time_job_post(self):
         one_time_job = OneTimeJob.objects.get(job=script.job, configuration=configuration)
         self.assertEqual(one_time_job.serial_numbers, [serial_number])
         self.assertEqual(one_time_job.tags.count(), 0)
+
+    # PBAC: a forbid on a JobKind refuses the schedule even though the role holds the model
+    # permission. The forbid rides on top of the broad grant login() writes, because the same actions
+    # are also asked resourcelessly through the legacy path and a kind-scoped permit cannot answer
+    # that question.
```
np5: **nit:** Stale rationale: there is no `JobKind`, `login()` writes no broad grant here (these tests use `login_with_policy`), and the actions are no longer asked through the legacy path. A kind-scoped permit does answer the question now. The forbid is just the shape these tests chose.

tests/turbo/test_setup_one_time_jobs.py
```diff
@@ -278,3 +281,47 @@ def test_delete_one_time_job(self, post_event):
         self.assertEqual(len(audit_events), 1)
         self.assertEqual(audit_events[0].payload["action"], "deleted")
         self.assertEqual(audit_events[0].payload["object"]["model"], "turbo.onetimejob")
+
+    # PBAC: a forbid on a JobKind refuses the schedule even though the role holds the model
+    # permission. The forbid rides on top of the broad grant login() writes, because the same actions
+    # are also asked resourcelessly through the legacy path and a kind-scoped permit cannot answer
+    # that question.
```
np5: **nit:** Same stale block as in `test_setup_enrolled_machines.py`: no `JobKind`, no legacy path, and a kind-scoped permit works on its own now.

tests/turbo/test_pbac.py
```diff
@@ -0,0 +1,407 @@
...
+        refused = force_command(backend=CommandBackend.FILE_EXPORT)
+        allowed = force_command(backend=CommandBackend.SYSDIAGNOSE)
+        self.assertFalse(
+            self._authorized(CreateOneTimeJobRequest(self.user, configuration, refused.job)))
+        self.assertTrue(
+            self._authorized(CreateOneTimeJobRequest(self.user, configuration, allowed.job)))
+
+    def test_a_forbidden_kind_still_offers_the_action(self):
+        # the has guard is what keeps the console link alive: the question with no job must not be
+        # refused by a forbid that reads job_kind
```
np5: **nit:** There is no `has` guard any more. The preview (partial evaluation) is what keeps the link alive, as the comment on `test_the_context_job_is_required` says.

zentral/contrib/turbo/views/one_time_jobs.py
```diff
@@ -10,21 +14,50 @@ class OneTimeJobListView(SearchFormListView):
...
+    # form_valid asks again with the job, and that decision is the one that counts.
+    def get_pbac_request_kwargs(self, kwargs):
+        return {"configuration": self.configuration}
     model = OneTimeJob
     form_class = OneTimeJobForm
     anchor = "one-time-jobs"
 
+    def form_valid(self, form):
+        check_create_one_time_job(self.request, self.configuration, form.cleaned_data["job"])
+        return super().form_valid(form)
```
np5, line 34: **suggestion:** The form offers every job, and a kind the policy refuses ends in a 403 page after submit, with the form lost. Two options: filter the `job` queryset with one batched `CreateOneTimeJobRequest` per candidate through `engine.authorize_requests`, or catch the refusal here and add it as an error on `job`. Not blocking, the decision that counts is the right one, but it is the first thing an operator with a kind-scoped grant meets. Same for `ScheduleMachineOneTimeJobView.form_valid`.

zentral/contrib/turbo/views/base.py
```diff
@@ -5,25 +5,28 @@
...
+    # configuration page and redirect back to it. Subclasses set model, form_class, anchor (the
+    # configuration-page fragment the success URL jumps to) and their own authorization mixin —
+    # PermissionRequiredMixin for the recurring views, PBACViewMixin for the one-time ones.
     anchor = None
 
+    def setup(self, request, *args, **kwargs):
+        super().setup(request, *args, **kwargs)
+        # resolved in setup() and not in dispatch() so that an authorization mixin running in dispatch
+        # can use it as the resource of its request
+        self.configuration = get_object_or_404(Configuration, pk=kwargs["configuration_pk"])
```
np5: **nit:** `setup()` runs before any authentication check. An anonymous request for an unknown configuration pk now gets a 404 where it got the login redirect before (`PermissionRequiredMixin.dispatch` ran first), and this holds for the recurring views too. UUID pks make it academic, but the PR argues 404-vs-403 ordering for serials two files over. A `cached_property` would keep the lookup lazy: `PBACViewMixin.dispatch` calls `get_pbac_request_kwargs` after its `is_authenticated` check, so that would be the first reader.

zentral/contrib/turbo/api_views/one_time_jobs.py
```diff
@@ -1,17 +1,47 @@
...
 from ..serializers import OneTimeJobSerializer
 
 
+class OneTimeJobPermissions(DefaultDjangoModelPermissions):
+    # turbo.add_onetimejob, change_onetimejob and delete_onetimejob no longer map to an action, so
+    # they cannot gate anything here. The write methods are authorized in perform_create /
+    # perform_update / perform_destroy instead, on validated objects rather than primary keys off the
+    # wire. GET keeps turbo.view_onetimejob, which is still legacy-mapped.
+    perms_map = dict(DefaultDjangoModelPermissions.perms_map,
+                     POST=[], PUT=[], PATCH=[], DELETE=[])
```
np5, line 15: **question:** With the write methods unmapped, any authenticated token reaches the serializer before the typed check. A caller with no Turbo access at all gets the 400s first, including whether a configuration or job pk exists, and the 403 only once the body validates. Acceptable, or should the write methods keep a coarse gate in front of validation, for example `turbo.view_onetimejob` on `POST`, `PUT` and `DELETE`?

zentral/contrib/turbo/templates/turbo/command_detail.html
```diff
@@ -0,0 +1,48 @@
...
+    <thead>
+        <th style="width:30vw">Attribute</th>
+        <th>Value</th>
+    </thead>
+    <tbody>
+        <tr><td>Name</td><td>{{ object.name }}</td></tr>
+        <tr><td>Description</td><td>{{ object.description|default:"-"|linebreaksbr }}</td></tr>
+        <tr><td>Kind</td><td>{{ object.get_backend_display }}</td></tr>
+        <tr><td>Version</td><td>{{ object.job.version }}</td></tr>
+        {% for key, value in backend_kwargs %}
```
np5, line 28: **nit:** `{{ value }}` renders a list as its Python repr (`['/var/log/install.log*']`). `{{ value|join:", " }}` renders a list and passes a scalar through unchanged.

docs/apps/turbo.md
```diff
@@ -126,6 +127,125 @@ An mSCP check *is* a compliance check – one is created with it, named after th
...
+
+```
+/Library/Logs/Foo/*.log
+/Library/Logs/Foo/*/*.log
+/Library/Logs/Foo/*/*/*.log
+```
+
+A `file_export` run produces two files: a manifest, which lists what was matched, collected, skipped and unreadable, and an archive of the collected files. The manifest is a separate file so an operator can read what a run collected without downloading the archive. A run that matched nothing uploads the manifest and no archive.
+
+**A command runs one time only.** It cannot be attached to a recurring job — collecting the same files every hour is a log shipper, not a job. Schedule it with a one-time job, or with the *Schedule one-time job* action on a machine page.
```
np5: **nit:** The machine-page path now requires `turbo.view_enrolledmachine` as well. The CHANGELOG says so, this page does not. One sentence here, or in the actions section below, keeps them aligned.

server/pbac/engine.py
np5: Fixed in 5e5897b1. `register_action` walks `applies_to.context` through `_register_attr_entity_types`, so it recurses `SetOf` / `RecordOf` the same way the attribute walk does. Three tests next to the one you named: an entity-typed context attr, one behind a `SetOf`, and `context=None` (which must not be walked).

CHANGELOG.md
np5: Fixed in 5e5897b1. Both moved into `## 2026.6`: the feature paragraphs as a `#### Turbo` under *Features*, and the 🧨 next to `PBAC User action groups` under *Backward incompatibilities*. You are right about what 2026.5's `#### Turbo` block is — it describes the module that shipped, so appending to it was the mistake.

zentral/contrib/turbo/forms.py
np5: Fixed in 5e5897b1, on `OneTimeJobSearchForm` only, for the reason you give: no command can be recurring, so adding it to `RecurringJobSearchForm` would be a filter on a relation that can never match.

zentral/contrib/turbo/public_views/config.py
np5: Fixed in 5e5897b1, in both places, with `raise` replaced by setting the entry aside as `unknown_job_kind` — a new documented `skipped` reason.

Two things I found writing the tests, which change the shape slightly:

**The reachable path is narrower than the comment suggests.** The wire `kind` is a `ChoiceField(choices=Job.Kind.values)`, so an entry *naming* the unknown kind is rejected by the serializer before `_process` sees it — it lands in `serializer.rejected`, not in `skipped`. What gets through is an entry with **no** kind, which the wire allows (`required=False`). My first two tests passed for the wrong reason until I noticed.

**Status needed the guard lower down.** Skipping the entry outright would stop writing the tracker, and the next `cleanup` sweep would mark a job the agent is still holding as removed — worse than a missing ack. So the row is updated as usual and only the event ref is skipped. Results does record the run, for the reason you gave: the run happened, only the outcome is unusable.

tests/turbo/test_api_one_time_jobs.py
np5: You are right on every count, including why it passed. Replaced in 5e5897b1 by `test_a_kind_scoped_permit_alone_is_enough`, which pins exactly what you describe: a bare `permit … when { context.job.kind == "sysdiagnose" }`, with no broad grant underneath and no companion forbid, gives `201` for a sysdiagnose and `403` for a file_export, and creates nothing in the second case.

tests/turbo/test_api_one_time_jobs.py
np5: Fixed in 5e5897b1. The block now says these tests use one of the two shapes that work — a forbid on top of the broad grant — and points at the permit-only test for the other.

tests/turbo/test_setup_enrolled_machines.py
np5: Fixed in 5e5897b1.

tests/turbo/test_setup_one_time_jobs.py
np5: Fixed in 5e5897b1, same rewrite as in `test_setup_enrolled_machines.py`.

tests/turbo/test_pbac.py
np5: Fixed in 5e5897b1 — the comment now says the preview is what keeps the link alive, matching `test_the_context_job_is_required`.

zentral/contrib/turbo/views/one_time_jobs.py
np5: Done in 5e5897b1, and your first option — but narrowing only what the picker **offers**, not the field's queryset.

I tried the queryset first and it made things worse: with the refused job out of the queryset, a post naming it comes back as *select a valid choice* from form validation, so `check_create_one_time_job` never runs and nothing is logged as a permission denial. The two tests asserting `403` on that path started returning `200`, which is how I noticed.

So `offerable_jobs()` returns the allowed jobs, the form assigns them to `field.choices`, and `field.queryset` stays whole — a forged post still reaches the typed check and is refused there. Those two `403` tests are unchanged, which is the proof, and a new test pins that the refused kind is absent from `choices` while still present in `queryset`. One batched evaluation per render. `ScheduleMachineOneTimeJobView`'s form too.

zentral/contrib/turbo/views/base.py
np5: Fixed in 5e5897b1 with the `cached_property` you suggested. `PBACViewMixin.dispatch` is indeed the first reader, and the recurring views get it back too since they keep `PermissionRequiredMixin`. Pinned by a test: anonymous + an unknown pk is a login redirect again, not a 404.

zentral/contrib/turbo/api_views/one_time_jobs.py
np5: Acceptable, I think — I left it as is, and here is the reasoning to push back on.

Both pks are UUIDs, so the 400-vs-403 ordering only confirms an identifier the caller already holds; there is nothing to walk. Against that, putting `turbo.view_onetimejob` in front of the writes would newly require a permission those paths never required, and a role converted from `add_onetimejob` without `view_onetimejob` would stop being able to create — an odd combination, but one the migration can have produced.

If the leak matters more than that risk, the one-line `perms_map` change is the fix and I am happy to make it.

zentral/contrib/turbo/templates/turbo/command_detail.html
np5: Fixed in 5e5897b1 with `|join:", "`, which passes a scalar through unchanged.

docs/apps/turbo.md
np5: Fixed in 5e5897b1 — a sentence in the *A command runs one time only* paragraph, with the reason (the configuration is only known once the serial resolves). The new `unknown_job_kind` skipped reason is documented in the same pass.

zentral/contrib/turbo/public_views/results.py
```diff
@@ -58,6 +58,16 @@ def _process(self, entries, batch):
...
+            if job.definition is None:
+                # a kind this release does not know: an instance still on the previous release can be
+                # handed the results of a job a newer one served. definition_wire_ref() would read
+                # .pk on None and 500 the whole batch — every valid entry in it waits for the refresh
+                # to finish, with a traceback on each retry.
+                logger.warning("Turbo results from %s: unknown kind %r for schedule %s",
+                               serial_number, job.kind, schedule_pk)
+                batch.record_run(job_machine, job, entry["version"], ran_at)
+                skipped.append({**ack, "reason": "unknown_job_kind"})
+                continue
```
np5, line 70: **issue (follow-up on the resolved `config.py` thread):** This guard is reachable only by an entry with no `kind`, and the agent always sends one: `TurboAPI.swift` encodes `kind` as `Job.Kind` on every results entry and every status entry. On an instance still on the previous release, `WireResultSerializer.kind` (`wire.py:61`, a `ChoiceField` over `Job.Kind.values`) rejects the entry that names the new kind before `_process` sees it. Per the wire contract, a rejected entry appears in neither list and the agent must not resend it. So the result is dropped, the run is never recorded, the new instances re-serve the job, and the machine runs the command again. For a sysdiagnose that is a repeated collection and upload. The 500 this replaces at least made the agent retry.

`WireStatusEntrySerializer.kind` (`wire.py:86`) has the same shape, so the status entry is rejected too and the tracker is not written, which is what the status guard set out to prevent.

The wire `kind` is only compared against `job.kind` (line 71, `kind_mismatch`); the event refs use the authoritative `job.kind`. A `CharField(max_length=32)` in both wire serializers lets the entry reach this branch, which then does the right thing, and `kind_mismatch` still catches a contradiction. Plus one test per endpoint that posts `"kind": "a_kind_from_the_future"` and asserts `unknown_job_kind`.

np5 (commented):
Re-review at 5e5897b1. All 14 threads are resolved: 12 fixed as described, the API ordering question declined with a rationale I agree with (a converted role that holds `add_onetimejob` without `view_onetimejob` would stop creating), and the rolling-deploy thread fixed for the entry shape it can reach.

That last one has a follow-up inline. The agent always names the kind, and the wire `ChoiceField` rejects that entry on an old instance before the new guard runs, so the realistic path still drops the result and re-runs the command. One small change in `wire.py` closes it, and it is small enough to fold into this PR.

Verified on the new head: CI green, patch coverage 100% (492/492), `mkdocs build --strict` clean, ruff clean apart from the unused import that is already on main. The `register_action` fix, the changelog move, the search filter, the lazy configuration and the choices-not-queryset narrowing of the pickers all check out. The queryset-stays-whole reasoning is right, and the two unchanged 403 tests are the proof.

np5:
Checked `c8d93b17`. The wire change is in: both `kind` fields are a `CharField(max_length=32)`, the results and status tests post the named unknown kind (plus the unnamed variant), and the docs say what the wire accepts. The three original commits are unchanged. `run_tests` is green and ruff is clean on the amended file.

Resolved the last thread. Nothing open from my side.

PR #1683, opened 2026-09-03
Move the documentation to Hugo

np5:
Move the documentation to Hugo, and stop the mkdocs build.

The documentation will be published at https://www.zentral.com/docs/. The website is a Hugo site. It will clone this repository and mount `docs/content/` and `docs/layouts/` into its own build. A companion pull request in the website repository does that part. This one makes `docs/` a Hugo site.

`docs/` is also a complete Hugo site on its own. That is the local preview, and the only reason the standalone configuration exists.

## What changes

- The pages move from `docs/` to `docs/content/`. Each `index.md` becomes `_index.md`: an `index.md` makes its directory a leaf bundle, and every sibling page becomes a resource instead of a page.
- `title` and `weight` in the front matter come from the mkdocs nav, so the sidebar keeps the same order and the same labels.
- Internal links become source-relative `.md` targets. A render hook resolves them against the page tree.
- `docker-compose.docs.yml` serves the preview. A new workflow builds the documentation on every pull request that can change it.
- `mkdocs.yml`, `.readthedocs.yaml`, `docs/requirements.txt` and the mkdocs theme stylesheet are removed.

## Preview

```
docker compose -f docker-compose.docs.yml up
```

Then open http://localhost:1314. Port 1313 stays free for the zentral.com preview.

## A broken link fails the build now

mkdocs resolved a link against the output URL, which is one level deeper than the source file. It reported an unresolved link as `INFO` and put it in the output unchanged. Nothing failed, and nothing checked the result.

The render hook builds an absolute path from the docs root and looks up the page. If there is no such page, the build fails. The workflow adds `--panicOnWarning`, so raw HTML that goldmark drops also fails.

The hook does not give the target to `GetPage` as written. A relative lookup that misses falls back to a search by name. Such a link works in the preview, where the docs are the whole site, and fails in the website build, where the docs are one level down. An absolute path gives the same result in both.

This found one broken link: `configuration/pbac.md` pointed to `../apps/`, which resolves to `/configuration/apps/`. `apps/` also had no page of its own, so there was no target. It has one now.

## Markup that mkdocs accepted and Hugo drops

Raw HTML in markdown stays off. This repository takes public pull requests, and the output is served from the company website.

- 23 `<br>` in the table cells of the Santa and the OneLogin pages. Each one was a cosmetic line wrap inside a cell. A cell wraps on its own.
- Three angle bracket placeholders in the MDM page. A browser reads them as unknown tags. That public key filename renders today as `.zentral.cloud_public_key__.pem`.
- The `width` attribute on the welcome page logo came from the mkdocs `attr_list` extension. The stylesheet sizes the image now.

## Two configurations, one source

The website supplies the configuration for the published build. Every markup option, and the Hugo version, must be the same on both sides. A difference only shows up as a preview that renders differently from the site. `docs/hugo.yaml`, `conf/start/docker/hugo/Dockerfile` and `docs/README.md` say what has to stay in step.

## Verification

The build reports 32 pages, no warnings and no errors. All 38 internal link targets in the output resolve. A test build with a broken link, and one with a raw HTML tag, both fail.


soysauceconspiracy (approved):
LGTM

PR #1684, opened 2026-09-03
turbo: the upload round trip, and CI on stacked pull requests

np5:
Stacked on #1682 — review that one first. This PR adds the upload plane the commands of #1682 need: a command collects a file, and until now the file had nowhere to go.

The file does not travel in the result. The agent asks for a destination, sends the artifact there, and reports the upload in the result. Four commits, in dependency order.

*Rebased onto the current #1682 head, and the review round is folded into the commits it corrected — the tree is unchanged from the one CI last validated as `2bbc5ae0`. The [review](https://github.com/zentralopensource/zentral/pull/1684#pullrequestreview-5103091757) and its inline threads are still the record of what changed and why.*

## `storage: presigned PUT uploads, signed with SigV4`

`generate_presigned_put()` in `zentral/utils/storage.py`, and a `file_storage_has_presigned_uploads()` predicate that is narrower than `file_storage_has_signed_urls()` on purpose: signing a download is a read of an object the server already has, signing an upload delegates a write.

The content length, the content type and the SHA-256 travel as **signed headers**, so the storage itself refuses a body that does not match, and Zentral never reads a byte to find out.

The key goes through the storage's own name normalisation, so the client writes where django-storages reads: with a `location` set, a presigned PUT built from the bare name would land outside the namespace `exists()`, `delete()`, `url()` and a later `HEAD` all look in.

**That guarantee only holds under SigV4, and it was silently not holding.** botocore reaches for SigV4 when the signature version was set *explicitly*. Where it is merely resolved from the endpoint, `generate_presigned_url` falls back to the legacy V2 query signature: the length is not signed at all, and the checksum becomes a query parameter S3 does not read. Measured:

|Configuration|Signature|
|---|---|
|`us-east-1`, no `signature_version`|**V2**|
|custom `endpoint_url`, no `signature_version`|**V2**|
|`eu-central-1`, no `signature_version`|V4|
|explicit `s3v4`, any region|V4|

Two of the most common deployments, and nothing to see in either direction. So the version is forced rather than trusted, on a client of the storage's own session, so uploads authenticate exactly like downloads. `test_generate_presigned_put_signs_with_sigv4` pins the `us-east-1` case that used to fall back.

## `turbo: the JobUpload row, and the reads the mint needs`

One row per `(run, artifact)`, created by the first request for a destination and closed by that run's result.

**Named for the job, not the command.** A command is grouped under one model because per-kind reification is not justified for it. That has nothing to do with the reason this plane exists — some jobs produce artifacts — so a kind reified later can write here without the plane learning what a `Command` is. For the same reason the artifact declaration is read through the **definition** (`job.definition.artifacts`), never through the command backend. A `Script` and an `MSCPCheck` declare none: a verdict is not an artifact.

Keyed like the wire and not like the ORM — `schedule_pk` + `schedule_mode`, no foreign key to either tracker — so one column covers both scheduling models and a recurring upload kind needs no migration. The cost is no cascade.

`status` and `verification` are two axes on purpose. "The agent says it uploaded" and "the storage agrees" are two facts, and the second must never reopen a shot the first closed. The multipart columns are here from the start, empty: they cost nothing and save two migrations.

One row per run and artifact is the **only** uniqueness asserted, and the absence of the other one is deliberate. An index over the uploaded rows of a one-time schedule looks right and is false: a version bump re-arms a one-time job, so a machine can legitimately hold two runs of one shot and upload for both. Such an index makes the second result an `IntegrityError`, and under `ATOMIC_REQUESTS` that is a `500` on the whole batch, on every retry, until someone deletes a row by hand.

The artifact declaration **and the size ceiling** are read through the definition, as properties on `Command` rather than `getattr()` against the model — `BackendInstance.__getattr__` raises for a name it does not know, so a `getattr` with a default answers `None` silently and the kind's ceiling never applies.

`get_machine_schedule()` is the resolver for the mint, and deliberately not the batch one — it creates no tracker row, so a run that asks for a destination and never reports leaves no trace in a ledger it does not own. Its rules are the config view's: a schedule the agent could not have been served is one it cannot ask a destination for.

## `turbo: the upload round trip — mint, upload, close`

`POST /public/turbo/uploads/`, machine-token auth, and the `result.uploads[]` entry that closes the row.

**Just in time, never in the config payload.** That payload is a pure function of `(definition, version)`, cached and served again to every in-scope machine — a URL in it would be expired for the machines that needed it most, and a live write credential for the ones that did not.

**The server owns the key**, and it is stable for the life of the row: a retry re-signs the same key and replaces the object instead of leaving a half-written twin, and the filename a browser saves is settled at the first request. At ingest the echoed key is **compared** to the minted one, never taken from the wire.

Two destinations, one contract. A presigned PUT signs the length and the digest as headers. The hosted destination has to read the body, and checks both itself — the length first, then the digest, which a wrong length also fails but explains less well. That check is not redundant with S3: it is the only moment a file system storage can ever verify the bytes, because it reports no checksum to read back later. Only S3 can presign today, so a Google Cloud Storage deployment always takes the hosted path and cannot choose otherwise. The agent sees one `mode: "put"` shape and has no storage to detect.

**The ceilings are the deployment's, not ours.** Without a presigned storage the limit is `DATA_UPLOAD_MAX_MEMORY_SIZE`, because that is the wall the body actually meets; publishing anything else spends the artifact, the upload and every retry discovering it. The kind's own maximum clamps under it.

**Every wire value is checked at the boundary**, because past it they reach bounded columns and an encoder — a length check on a `sha256` is not a format check, and 64 characters of non-hex reach `unhexlify` as a `500` where the agent needed a `400`. Rows are bounded too: attempts bound a row, and `run_id` is the agent's, so a machine that never reports could otherwise mint a fresh write credential for every id it invented. The count excludes the row the request is about, so the bound falls on a new row and never on a retry.

The result closes the row **on the agent's word alone**; verification against the storage is the second axis and lands in a later PR. Partial success is first-class: a `file_export` that uploads its manifest and gives up on the archive keeps the inventory of what it collected, and an optional artifact that never happened leaves no row sitting pending. Ingest stays tolerant — an entry that cannot be matched is logged, never a reason to fail a batch whose run did happen.

`WireError` grows a status, because the code alone cannot tell the agent whether to report a failure or say nothing at all: `404` and `410` for a schedule that is gone or a shot already closed, `413` over a maximum, `429` when too much is open.

The config payload publishes `upload_max_size` and `upload_digests`, so the agent can refuse a file that is too large before it collects it, and read the file one time to compute every digest. No part size: nothing the agent computes depends on the chunk geometry.

## `ci: run the checks on every pull request, not only the ones based on main`

Repo-wide, and the one commit here that is not about uploads.

All three workflows filtered `pull_request` on `branches: [main]`, so a stacked pull request — based on the branch below it — got no checks at all. The review that most needs them, since the diff is the only thing separating it from a PR that would run them. The filter is gone from the `pull_request` trigger of `tests`, `docker` and `osv-scanner`.

`push` keeps its `main` filter: without it every branch push would run the whole suite a second time next to the pull request run. `merge_group` keeps its filter too, because a merge queue only exists for `main`.

The change applies to itself — `pull_request` runs the workflow file from the PR head — so this PR is the first one it checked.

## For the agent team

Two contract changes since the branch you tested against:

* The hosted destination's `200` has a body now (`{}`) where it was empty, and its failures carry JSON error codes: `invalid_token`, `unknown_upload`, `already_uploaded`, `size_mismatch`, `digest_mismatch`, `too_large`.
* `upload_max_size` differs between deployments rather than being a constant — 2 GiB on S3, the deployment's own body limit without it. That was always the intent; until now the number was the same everywhere because it was wrong everywhere.
* A `file_export` over 500 MiB is refused at the mint now. The cap was always declared; it just never ran.
* New `429 {"error": "too_many_pending"}` when a machine holds more than twelve destinations on one schedule that no result has closed. Reporting the runs already made is what clears them; asking again for a destination the machine already has is never refused for this reason.

## Not in this PR

Verification against the storage, multipart, and the console and API to list and download what was collected.

**Cleanup**, which nothing does yet: a `JobUpload` row is keyed like the wire and has no foreign key to either scheduling model, so deleting a schedule leaves its rows and their objects behind — and a row whose upload never completed leaves a key that was signed and never written. There is no sweeper and no lifecycle rule. It needs both before this carries a real fleet's traffic.

## Tests

Full suite green (8622), 70 new tests. 100% patch coverage on every production line.

One note on that number, because the first version of it was wrong. Patch coverage was measured with `git diff`, which omits untracked files — so the three new modules, which are the whole point of the PR, went unchecked while the report said 100%. Measured against the staged diff instead, six production lines were uncovered; they are covered now.


np5 (commented):
Read the five commits with the full files at the head, on top of the `turbo-command-kinds` base (its tree is the one I checked as `c8d93b17`). Ran ruff (clean), `mkdocs build --strict` (clean), and three checks in a throwaway container: the presigned PUT signs exactly `content-length`, `content-type`, `host` and `x-amz-checksum-sha256`, so the three headers the mint returns are the complete set the agent has to send; the two findings below about the key namespace and the per-kind cap reproduce on the branch. CI is green and coveralls reports full patch coverage.

Two things to fix before merge, one latent one-liner worth doing now, the rest suggestions.

- `zentral/contrib/turbo/public_views/uploads.py`: the per-kind maximum never reaches the mint. The attribute is read off the model, where it does not exist, so `file_export`'s 500 MiB cap is dead and only the deployment ceiling applies.
- `zentral/contrib/turbo/models.py`: the partial unique index on uploaded one-time rows conflicts with a version bump between two runs of one shot. Under `ATOMIC_REQUESTS` that is a results batch that 500s on every retry for that machine. The sequence is inline.
- `zentral/utils/storage.py`: the presigned key skips `_normalize_name`, so a storage `location` would split the namespace between what the agent writes and what django-storages reads. No deployment sets one today.
- Suggestions: check the digest on the hosted PUT (the only place a filesystem storage can ever check it), refresh the storage digests on a retry, a conservative default when Django's body limit is disabled, and a docs sentence on GCS taking the hosted path.
- One question inline on bounding mints per schedule. And one not inline: `JobUpload` rows and their objects have no cleanup path yet, and cleanup is not in the "Not in this PR" list; worth naming there so it is not lost.

What looked good: the SigV4 pinning is the right fix and the test pins the exact case that used to fall back; the mint resolver that creates no tracker row, the stable server-owned key, the compare-not-take of the echoed key, and the two-axis status are all sound; the CI trigger change is correct and applies to itself.

zentral/contrib/turbo/public_views/uploads.py
```diff
@@ -0,0 +1,158 @@
...
+
+        url, headers = build_upload_destination(upload, artifact, storage)
+        self.request_event_payload = {"artifact": artifact_name, "size": size,
+                                      "attempts": upload.attempts}
+        return {"mode": upload.mode, "url": url, "headers": headers, "key": upload.key,
+                "expires_at": (timezone.now() + timedelta(seconds=UPLOAD_URL_EXPIRY)).isoformat()}
+
+    def _kind_max_size(self, job):
+        definition = job.definition
+        return getattr(definition, "max_upload_size", None) or upload_max_size()
```
np5: **issue:** The per-kind cap is never applied. `job.definition` is the `Command` model, and `max_upload_size` lives on the backend class, not on the model. `BackendInstance.__getattr__` raises `AttributeError` for a name it does not know, so `getattr(definition, "max_upload_size", None)` is always `None` and the deployment ceiling wins: a 600 MiB `file_export` mints fine against the 2 GiB S3 ceiling. `test_too_large` uses 10 GiB, above the deployment ceiling, so it cannot see this.

Checked on the branch: `getattr(command, "max_upload_size", None)` is `None`; `command.get_backend().max_upload_size` is 524288000.

Same shape as `artifacts`: a `Command.max_upload_size` property that reads the backend, `None` on `Script` and `MSCPCheck`, and a test under `S3_STORAGE` with a `file_export` artifact between 500 MiB and 2 GiB that expects `413`.

zentral/contrib/turbo/models.py
```diff
@@ -796,6 +820,100 @@ def __str__(self):
...
+        default_permissions = ()   # the agent writes these rows, never an operator
+        constraints = [
+            models.UniqueConstraint(fields=["schedule_pk", "serial_number", "run_id", "artifact"],
+                                    name="turbo_jobupload_unique_run_artifact"),
+            # the one-time gate, in Postgres: a shot closes on its first result, so it has at most one
+            # successful upload per artifact. Recurring rows carry no gate, hence the mode.
+            models.UniqueConstraint(
+                fields=["schedule_pk", "serial_number", "artifact"],
+                condition=Q(status="uploaded", schedule_mode="one_time"),
+                name="turbo_jobupload_unique_one_time_uploaded",
```
np5: **issue:** This index turns a legitimate sequence into a results batch that never clears.

1. A one-time job is served at version 1. The machine runs it and mints (run R1).
2. The operator edits the command. The version moves to 2.
3. The machine takes the new config, runs again, and mints (run R2).
4. R1's result arrives. It is stale, so `record_run` does not close the gate, but `_close_uploads` (results.py:150) marks R1's row `uploaded`; it has no version check.
5. R2's result arrives. `_close_uploads` sets its row to `uploaded`, and `row.save()` hits this index.

`ATOMIC_REQUESTS` is on, so the `IntegrityError` rolls the whole POST back into a 500. The agent retries, the retry fails the same way, and that machine's results stop flowing until someone deletes a row. The gate this index mirrors already lives on the tracker (`last_result_at`), and the mint already refuses a closed gate and an uploaded row.

Two ways out: drop the index and let the ledger decide which upload is current by version, or make `_close_uploads` treat "another run already uploaded this artifact" as a logged failure of the row instead of a save that cannot succeed. Either way, an `IntegrityError` on this path must not reach the agent.

zentral/utils/storage.py
```diff
@@ -9,6 +12,87 @@ def file_storage_has_signed_urls(storage=None):
...
+    headers = {
+        "Content-Type": content_type,
+        "Content-Length": str(size),
+        "x-amz-checksum-sha256": _sha256_base64(sha256_hex),
+    }
+    url = client.generate_presigned_url(
+        "put_object",
+        Params={
+            "Bucket": storage.bucket_name,
+            "Key": key,
```
np5: **issue (latent):** `Key: key` bypasses the storage's namespace. `S3Storage.url()` builds its key as `self._normalize_name(clean_name(name))`, which prepends `location`. With a `location` set, the presigned PUT writes `turbo/uploads/…` at the bucket root, while every django-storages read of `upload.key` (`exists`, `open`, `delete`, `url`, and the HEAD the verification PR will do) looks under `<location>/turbo/uploads/…` and finds nothing.

Checked on the branch with `location="zentral"`: presigned path `/turbo/uploads/x/…`, `storage.url()` path `/zentral/turbo/uploads/x/…`.

No shipped deployment sets `location` today (ztl_admin passes the bucket, the region and `s3v4` only), so nothing breaks now. One line keeps the two paths on one namespace, at the cost of a private name that `url()` itself relies on:

```python
from storages.utils import clean_name
…
"Key": storage._normalize_name(clean_name(key)),
```

zentral/contrib/turbo/public_views/uploads.py
```diff
@@ -0,0 +1,158 @@
...
+            # that declared a size it did not send: without it Django answers a SuspiciousOperation
+            # with an HTML page, on the one endpoint contracted never to serve one.
+            logger.warning("Turbo hosted upload %s: body over the deployment limit", upload.pk)
+            return JsonResponse({"error": "too_large"}, status=413)
+        if len(body) != upload.size:
+            # a presigned PUT has the length in the signature and the storage enforces it for us;
+            # here there is nobody else to do it, and the size is what the mint was clamped against
+            logger.warning("Turbo hosted upload %s: %s bytes declared, %s sent",
+                           upload.pk, upload.size, len(body))
+            return JsonResponse({"error": "size_mismatch"}, status=400)
```
np5, line 173: **suggestion:** The hosted destination has the body and the declared digest, and checks only the length. The presigned path gets the digest checked by S3. A `hashlib.sha256(body).hexdigest() != upload.sha256` here, answered with `400 {"error": "digest_mismatch"}`, makes the two destinations give the same guarantee. It is also the only point where a filesystem storage can ever check the digest: there is no storage-side checksum to read back later, so the verification axis will have nothing to compare on this path.

zentral/contrib/turbo/public_views/uploads.py
```diff
@@ -0,0 +1,158 @@
...
+        if upload.attempts >= MAX_UPLOAD_ATTEMPTS:
+            raise WireError("attempts_exhausted")
+
+        # a retry re-signs the SAME key, so it overwrites in place instead of leaving a twin
+        if not upload.key:
+            upload.key = build_upload_key(upload, artifact)
+        # the digests belong to the bytes, and a retry of the same artifact carries the same ones; a
+        # retry with different bytes is a different artifact and would sign a header it cannot match
+        upload.size = size
+        upload.sha256 = sha256
```
np5: **nit:** On a retry the row takes the new `size` and `sha256` but keeps the first mint's `crc64nvme` and `crc32c` (they are in `defaults` only). If a retry carries different bytes, the stored storage digest belongs to the old bytes, and the verification axis later compares against the wrong value. Either update the digests here too, or keep all three fixed after the first mint.

zentral/contrib/turbo/public_views/uploads.py
```diff
@@ -0,0 +1,158 @@
...
+            raise WireError("too_large", status=413)
+
+        upload, _ = JobUpload.objects.get_or_create(
+            schedule_pk=schedule.pk, serial_number=self.serial_number,
+            run_id=run_id, artifact=artifact_name,
+            defaults={"schedule_mode": schedule.wire_mode, "size": size, "sha256": sha256,
+                      "crc64nvme": digests.get("crc64nvme", ""),
+                      "crc32c": digests.get("crc32c", ""),
+                      "key": ""},
+        )
```
np5, line 87: **question:** `run_id` is the agent's, and nothing bounds it. A machine token that never posts a result can mint a fresh row and a fresh 2 GiB write credential per `run_id`, for a one-time schedule too, since the gate only closes on a result. `MAX_UPLOAD_ATTEMPTS` is per row, not per schedule. Is a cap on pending rows per `(schedule_pk, serial_number)` worth having now, or is a compromised enrolled machine outside the threat model of this plane?

zentral/contrib/turbo/uploads.py
```diff
@@ -0,0 +1,163 @@
...
+# How long a minted URL is good for. Generous, because the agent may be uploading hundreds of
+# megabytes over a hotel connection, and a signature is checked at request START — a slow upload that
+# begins in time finishes.
+UPLOAD_URL_EXPIRY = 3600
+
+# The DEPLOYMENT ceiling on one artifact, which the config response publishes so the agent can fail
+# fast instead of compressing 2 GiB it will not be allowed to send. The per-kind maxima clamp below
+# it, server-side at mint.
+PRESIGNED_UPLOAD_MAX_SIZE = 2 * 2**30
+HOSTED_UPLOAD_MAX_SIZE_UNDECLARED = 32 * 2**20
```
np5: **nit:** When the deployment disables Django's body limit, this publishes 32 MiB, but the shipped nginx configurations cap at 10 MiB. It is the one remaining case where the published ceiling can exceed the wall the agent meets, the exact mismatch the last commit fixes. 10 MiB, the value the shipped proxies use, would be the conservative default; a deployment that raised nginx can raise Django too.

docs/apps/turbo.md
```diff
@@ -368,6 +368,29 @@ When `collect_inventory` is enabled, the agent posts a full machine inventory sn
...
+
+Zentral builds the object key, and the key is stable for the life of the upload. A retry signs the same key again and replaces the object, so a failed attempt never leaves a partial twin behind. The file name is built at the first request, from the artifact declaration, the serial number and the time.
+
+Where the bytes go depends on the storage backend:
+
+ * When the default storage can sign a `PUT` – S3 – the agent sends the file directly to the bucket. Zentral reads none of it, and the size and the digest are signed into the destination, so the storage itself refuses a body that does not match.
+ * When the storage cannot sign a `PUT`, Zentral hosts the destination and the bytes go through the web workers. Use this only in development, or for a small on-premises installation. The size limit is then the body limit of the deployment itself, `DATA_UPLOAD_MAX_MEMORY_SIZE`, which is 10 MiB by default. Raise it only together with the `client_max_body_size` of your reverse proxy: the smaller of the two is what the agent meets, and the published limit only knows about the first.
```
np5: **nit:** GCS cannot presign here, so every GCP deployment takes this path, with the 10 MiB ceiling, and it cannot choose otherwise. "Development or a small on-premises installation" reads as a choice. One sentence that names GCS as not supported for presigned uploads yet would set the expectation; the CHANGELOG paragraph has the same gap.

zentral/contrib/turbo/public_views/uploads.py
np5: Confirmed and fixed in b402614c. `Command` had `artifacts` but not `max_upload_size`, and `BackendInstance.__getattr__` raises, so the default won every time — checked on the branch before touching anything.

Same shape as `artifacts`, as you suggested: a property on `Command` reading the backend, `None` on `Script` and `MSCPCheck`.

The line had a second fault worth naming. The fallback called `upload_max_size()` with no storage, so on an S3 deployment it answered with the *hosted* ceiling while the `min()` beside it used the presigned one — the two halves disagreed about which deployment they were in. It is one function now that takes the storage and returns the lower of the kind's limit and the deployment's.

Two tests under `S3_STORAGE`: a 600 MiB `file_export` gets `413`, 499 MiB gets `200`, so the ceiling is the kind's own and not something smaller. Plus one that pins the clamp in the other direction — `sysdiagnose` declares 2 GiB, and against a storage that cannot presign the deployment limit still wins.

zentral/contrib/turbo/models.py
np5: Reproduced exactly as written, before changing anything:

```
IntegrityError: duplicate key value violates unique constraint
  "turbo_jobupload_unique_one_time_uploaded"
```

Fixed in b402614c by dropping the index, not by catching the conflict — because what the index asserts is false. A version bump re-arms a one-time job, so a machine holding two runs of one shot and uploading for both is the design working, not corruption. Handling the `IntegrityError` would have kept an index that forbids a state the system is built to reach, and left the next reader to wonder which of the two was wrong.

The gate stays where it can see the version, which the index never could: `last_result_at` on the tracker, read by the mint. Per-row `already_uploaded` still covers the case the index actually caught.

Migration 0003 is unreleased and in this branch, so it is edited rather than superseded. Your sequence is a regression test now — `test_two_runs_of_one_shot_across_a_version_bump`.

zentral/utils/storage.py
np5: Taken in b402614c, with your one-liner. The private name is worth it for the reason you give: `url()` depends on it, so the two paths have to agree, and there is no public accessor. Commented as such.

`test_generate_presigned_put_respects_the_storage_location` sets `location="zentral"` and asserts the presigned path and `storage.url()` produce the same path — which is the property, rather than the prefix appearing somewhere.

zentral/contrib/turbo/public_views/uploads.py
np5: Taken in b402614c. Your last sentence is the argument: on a filesystem storage there is no checksum to read back, so this is not a redundant check that S3 would repeat — it is the only moment that path can ever be checked at all.

One thing your version would have changed by accident: I put the digest check after the length check. A body of the wrong length fails the digest too, so digest-first made `size_mismatch` unreachable and cost the agent the more precise diagnosis. Length, then digest, and `digest_mismatch` now means what it should — right length, wrong bytes.

zentral/contrib/turbo/public_views/uploads.py
np5: Fixed in b402614c: the digests move with `size` and `sha256` on a retry. You are right about which way to resolve it — the digests belong to the bytes, so freezing them after the first mint would have kept a value describing bytes that are no longer at the key.

`test_a_retry_moves_the_storage_digests_with_the_bytes` covers it.

zentral/contrib/turbo/public_views/uploads.py
np5: Worth having, and implemented in b402614c.

A compromised token can already flood inventory and results, so unbounded rows are not new. A *write credential to the bucket* is a different kind of amplification, though, and your framing is exactly the gap: `MAX_UPLOAD_ATTEMPTS` bounds a row and nothing bounded the rows.

`MAX_PENDING_UPLOADS = 12` per `(schedule_pk, serial_number)`, answered `429 {"error": "too_many_pending"}`. Twelve because a version bump legitimately opens more than one run and a dozen is not legitimate.

My first cut was wrong in a way worth recording: the check ran before `get_or_create`, so it refused *retries* as well as new rows — which would strand an agent at the limit with no way to clear it, since reporting the runs it already holds is the only thing that frees a row. The count excludes the row the request is about now, so the bound falls on new rows only. `test_a_retry_of_a_bounded_schedule_is_not_refused` is the test that caught it.

zentral/contrib/turbo/uploads.py
np5: Taken in b402614c — 10 MiB, matching the shipped proxies. You are right that it was the one remaining case of the mismatch the commit beside it fixes, which makes leaving it indefensible: a deployment that raised nginx can raise Django too, and one that raised neither should not be told about a ceiling nobody honours.

docs/apps/turbo.md
np5: Taken in b402614c, in the docs and in the CHANGELOG. "Development or a small on-premises installation" did read as a choice, and for GCP it is not one — both now say that only S3 can sign a `PUT`, so a GCS deployment always takes the hosted destination and cannot choose otherwise.

np5:
Thanks — this was a good catch on both blockers. Fixed in b402614c, replies inline on each thread. Reproduced both before changing anything.

Two places where I went further than the suggestion, and one where I went a different way:

**The index is dropped, not defended.** You offered two exits; I took the first, because the index asserts something *false*. A version bump re-arms a one-time job, so a machine holding two runs of one shot and uploading for both is the design working. Teaching `_close_uploads` to swallow the conflict would have kept an index that forbids a state the system is built to reach, and left the next reader unsure which of the two was wrong.

**The dead cap had a second fault in the same line.** Its fallback asked for the ceiling without a storage, so on an S3 deployment it answered with the *hosted* number while the `min()` beside it used the presigned one — the two halves disagreed about which deployment they were in. Both are one function now.

**The digest check goes after the length check, not before.** A body of the wrong length fails the digest too, so digest-first made `size_mismatch` unreachable and cost the agent the more precise diagnosis.

On the bound: implemented, `MAX_PENDING_UPLOADS = 12`. A compromised token can already flood inventory and results, but a write credential to a bucket is a different kind of amplification, and you named the gap exactly — attempts bound a row, nothing bounded the rows. My first cut refused retries as well as new rows, which would have stranded an agent at the limit with no way to clear it; the count excludes the row the request is about now.

Cleanup is named in **Not in this PR** — the rows, the objects, and the keys that were signed and never written. There is no sweeper and no lifecycle rule, and it needs both before this carries a real fleet.

Full suite green at 8622, patch coverage complete, and the new agent-visible answers (`too_many_pending`, `digest_mismatch`) are in the docs.

np5:
Checked `b402614c`. All eight threads are fixed as described, and I resolved them.

Verified on the branch: `Command.max_upload_size` reads the backend (500 MiB for `file_export`, 2 GiB for `sysdiagnose`, `None` for scripts and mSCP checks), and the presigned path equals the `storage.url()` path under a `location`. The dropped index, the digest check on the hosted PUT, the digests moving on a retry, the pending cap that excludes the row of the request, the 10 MiB fallback and the GCS sentence all check out. Your regression test is the sequence from the thread. CI is green, patch coverage is full, ruff and the strict docs build are clean.

One small docs leftover, not worth a thread: the table under *The hosted destination* still lists the errors from before the digest check, so it has `size_mismatch` and not `digest_mismatch`, while the new paragraph above it names both. One row keeps the two in step.

Nothing open from my side.

PR #1685, opened 2026-09-03
turbo: does the storage agree? the second axis of an upload

np5:
Stacked on #1684 — review that one first.

*Rebased onto the current #1684 head (`41454e80`), and the [review](https://github.com/zentralopensource/zentral/pull/1685#issuecomment-5542100473) round is folded in. The rebase itself carried a silent break — the presign now normalises the object key and the verification HEAD did not — and the review's nit about the stat client turned up a second: `retries.max_attempts` is read by botocore as a retry count, so the ceiling allowed two attempts while its comment said one.* It established the first axis of an upload: the agent reported, and the row closed on its word. This PR establishes the second: whether the storage agrees.

They are separate on purpose. "The agent says it uploaded" and "the storage agrees" are two facts, and the second must never be able to invalidate the first.

## `storage: stat an object, and hash one the storage did not`

`stat_object()` reports the size and the whole-object digests the storage itself recorded. It returns `None` when there is nothing at the key — and only then. Every other failure raises, so a storage that is unreachable, or a key the deployment may not read, is never mistaken for an object that is not there.

Two S3 details it depends on, each silent when missed:

 * **`HeadObject` reports a checksum only when the request asks for it.** Without `ChecksumMode` the response is checksum-free, and a comparison against it would pass for any object at all.
 * **A missing key surfaces as the bare status `404`, not `NoSuchKey`** — `HeadObject` has no body to carry an error code — so an `except NoSuchKey` would never fire.

Both are pinned by tests stubbed against the real service model, so a parameter S3 does not have fails in CI rather than in production.

The HEAD gets an explicit ceiling instead of botocore's minute-scale defaults, and **one** attempt. It runs on a device endpoint inside the request transaction, so an unbounded call to a storage that has gone quiet would hold a web worker *and* a database connection; a retry would double how long that transaction stays open to buy an answer that was never urgent.

`sha256_object()` reads an object and hashes it, for the storage that keeps no digest of its own — affordable exactly because that is the storage with the small ceiling.

## `turbo: does the storage agree? the second axis`

When the agent reports an upload, Zentral asks the storage about the object at that key and compares the size and the SHA-256.

**The two numbers reach that comparison by two independent paths.** One went out at the mint and was signed into the upload, so the storage refused any body that did not hash to it, and comes back as the checksum the storage recorded. The other came home in the result. Verification is where the two copies meet, and on a storage that signs, Zentral reads no part of the file to do it.

Both legs are checked. An agent whose result contradicts its own mint is describing a different file, and comparing the stored object against the mint alone would miss it — on a signing storage that comparison is nearly tautological, since S3 enforced it on the way in.

`verified` / `mismatch` / `missing` lands on the row's own field, next to the status and never instead of it, and rides the echoed entry into the result event as `verification` — the verdict itself, not a yes or no, because `mismatch` and `missing` are different problems and a consumer of an event cannot go to the database to tell them apart. The event now also carries `run.id` — what an upload row is keyed on, without which the event describes an artifact nothing can find.

**Some answers are not answers, and none of them is recorded.** A storage that timed out, a key we may not read, and an object with no stored digest all leave the row `pending`, with a log entry. A transient blip that marked artifacts permanently missing would be worse than no answer at all. On a signing storage a missing digest is never a reason to read the object back — that would mean downloading two gigabytes to answer a question.

**A mismatch does not reopen the gate.** That is what two axes are for: the shot was consumed when the agent reported, whatever the storage says afterwards. There is a test for it.

One structural change: the uploads now close *before* the wire ref is built rather than after. The verdict has to ride the echoed entry, and enriching the outcome in place afterwards would only have worked by aliasing a dict two callers hold — the shape that has already cost this codebase one class of lost events.

## Not in this PR

Multipart, and the console views to list and download what was collected.

## Tests

Full suite green — 8647 tests. 100% patch coverage on every production line, verified by intersecting `coverage json` against the diff hunks. 32 new tests, `mkdocs build --strict` clean, all in-page anchors resolve.

The S3 paths are covered with `botocore.stub.Stubber` rather than a mock, so the expected parameters are validated against the real service model: a parameter S3 does not have fails here, not in production.

### DIFF
--- a/docs/apps/turbo.md
+++ b/docs/apps/turbo.md
@@ -389,7 +389,32 @@ The result closes the upload. It carries a `run.id`, and one entry in `result.up
 
 Partial success is a normal outcome. A `file_export` that sends its manifest and fails on its archive keeps the manifest: the two artifacts are two rows, and they close one at a time.
 
-The agent's report is the first axis, and Zentral records it immediately. Verification against the storage is a second axis, and it comes later. Verification never reopens a one-time job: a one-time job that reported an upload does not get another destination.
+The agent's report is the first axis, and Zentral records it immediately. Whether the storage agrees is a second axis.
+
+#### Verification
+
+When the agent reports an upload, Zentral compares the size and the SHA-256 of the stored object with the ones the agent declared when it asked for the destination.
+
+Those two numbers reach the comparison by two independent paths. They went out at the request for a destination and were signed into the upload, so the storage refused any body that did not match them, and they come back as the checksum the storage recorded. They also come home a second time in the result. Verification is where the two copies meet, and on a storage that signs, Zentral reads no part of the file to do it. Where the storage records no digest of its own, Zentral reads the object and hashes it — which is affordable because that is the same storage that has the low size limit.
+
+Zentral checks both paths. A result that echoes a size or a digest which is not the one it asked for is a `mismatch`, whatever is at the key: an agent that requests a destination for one file and reports another has described two files, and only one of them was stored. Echoing them at all is optional — the request for the destination is what was signed either way.
+
+The outcome is a separate field on the upload, next to the status:
+
+|Verification|Meaning|
+|---|---|
+|`pending`|Not decided yet.|
+|`verified`|The stored object is the size and the SHA-256 the agent declared.|
+|`mismatch`|Something is at the key and it is not what the agent declared, or the result contradicts the request for the destination.|
+|`missing`|There is nothing at the key.|
+
+The outcome rides the result event as well, as `verification` on the `result.uploads` entry. It is the outcome itself and not a yes or no, because `mismatch` and `missing` are different problems and a consumer of the event cannot go to the database to tell them apart.
+
+Some answers are not answers, and Zentral records none of them. A storage that times out, a key the deployment cannot read, or an object with no stored digest all leave the verification `pending`, with a log entry, and put no `verification` on the event. Recording a transient failure as `missing` would be worse than recording nothing: the object is very likely there, and the row would say for ever that it is not.
+
+Give Zentral `s3:ListBucket` on the bucket. Without it S3 answers 403 rather than 404 for a key that does not exist, and Zentral cannot tell a missing object from one it may not read: the upload stays `pending`, and every result for it writes an error to the log.
+
+**Verification never reopens a one-time job.** That is what the two axes are for: the job was done on that machine when the agent reported, whatever the storage says afterwards. A `mismatch` does not give the agent another destination, and it does not make Zentral ask for the file again.
 
 ### Tolerant batches
 
@@ -1982,6 +2007,8 @@ Each entry in `result.uploads` reports one artifact, and it is what closes that 
 |`error`|Present instead of `key` when the artifact did not go up. `error.reason` is one of `expired`, `http_status`, `network`, `too_large` or `mint_rejected`; `error.attempts` and `error.last_http_status` are informational. A reason this release does not know still closes the upload as failed – the artifact is not coming either way.|
 
 An entry that reports an artifact Zentral never minted a destination for is logged and ignored. It never fails the batch: the run happened, and its result is worth keeping.
+
+The `turbo_result` event carries the entries back, each with a `verification` added when [verification](#verification) reached an answer. The event also carries `run.id`, which is what an upload is keyed on.
 
 Response:
 
--- a/tests/utils/test_storage.py
+++ b/tests/utils/test_storage.py
@@ -2,10 +2,14 @@ import base64
 import hashlib
 from urllib.parse import parse_qs, urlparse
 
+from botocore.exceptions import ClientError
+from botocore.stub import Stubber
+from django.core.files.base import ContentFile
 from django.core.files.storage import storages
-from django.test import SimpleTestCase, override_settings
-from zentral.utils.storage import (file_storage_has_presigned_uploads, file_storage_has_signed_urls,
-                                   generate_presigned_put, select_dist_storage)
+from django.test import SimpleTestCase, TestCase, override_settings
+from zentral.utils.storage import (_stat_client, file_storage_has_presigned_uploads,
+                                   file_storage_has_signed_urls, generate_presigned_put,
+                                   select_dist_storage, sha256_object, stat_object)
 
 
 S3_STORAGE = {"default": {"BACKEND": "storages.backends.s3.S3Storage",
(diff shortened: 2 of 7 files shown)


zentral/utils/storage.py
```diff
@@ -9,6 +24,156 @@ def file_storage_has_signed_urls(storage=None):
...
+    if storage is None:
+        storage = storages["default"]
+    if file_storage_has_presigned_uploads(storage):
+        from botocore.exceptions import ClientError
+        client = _s3_client(storage, "_zentral_stat_client",
+                            connect_timeout=_STAT_CONNECT_TIMEOUT,
+                            read_timeout=_STAT_READ_TIMEOUT,
+                            retries={"mode": "standard", "max_attempts": _STAT_ATTEMPTS})
+        try:
+            response = client.head_object(Bucket=storage.bucket_name, Key=key, ChecksumMode="ENABLED")
```
np5: **issue (rebase-critical):** `Key=key` here is the bare name, and on the fixed #1684 head (`b402614c`) `generate_presigned_put` now signs `storage._normalize_name(clean_name(key))`. After the rebase the two disagree under a storage `location`: the agent writes the object under the prefix, this HEAD looks at the bucket root, S3 answers 404, and the row is recorded `missing`. That is a verdict, not an undecided answer, so it sticks. The dry-run cherry-pick of your two commits onto `2bbc5ae0` applies this file cleanly, which is exactly why nothing will flag it.

A small `_object_key(storage, key)` used by both the presign and this HEAD keeps them on one namespace, plus a Stubber test mirroring `test_generate_presigned_put_respects_the_storage_location` that expects `Key: "zentral/turbo/uploads/test"` under `location="zentral"`.

zentral/contrib/turbo/public_views/results.py
```diff
@@ -83,13 +92,80 @@ def _process(self, entries, batch):
...
+                    logger.warning("Turbo results from %s: echoed key does not match for %r on run %s",
+                                   self.serial_number, artifact, run_id)
+                    row.status = UploadStatus.FAILED
+                    row.error_reason = None
+                else:
+                    row.status = UploadStatus.UPLOADED
+                    row.truncated = bool(reported.get("truncated"))
+            row.completed_at = timezone.now()
+            if row.status == UploadStatus.UPLOADED:
+                verification = verify_upload(row, reported, storage)
```
np5, line 167: **question:** This is a storage round trip per uploaded artifact, inside the results POST, inside its transaction (`ATOMIC_REQUESTS` is on). The ceiling you set is 2 s connect plus 5 s read, one attempt. A results batch holds up to `results_batch_size` entries (default 100), and a machine draining a backlog of `file_export` runs brings two artifacts per entry. During a storage outage every HEAD pays the full connect timeout: fifteen artifacts in one batch is 30 s, gunicorn's default worker timeout, and this deployment does not raise it. The whole POST then 502s and nothing in it lands, tracker updates, compliance statuses and tags included, until the storage is back. The design says this answer "was never urgent", which is the argument for taking it out of the request: a task per row, enqueued here, is the shape the design notes already planned for multipart completion, because an enqueue is not a storage call. Is inline deliberate for this first slice? If so, a per-batch budget would bound the worst case, since the per-call ceiling does not.

zentral/contrib/turbo/public_views/results.py
```diff
@@ -83,13 +92,80 @@ def _process(self, entries, batch):
...
+                if verification is not None:
+                    row.verification = verification
+                    row.verified_at = timezone.now()
+                    uploads_ref[-1] = {**reported,
+                                       "verified": verification == UploadVerification.VERIFIED}
```
np5: **suggestion:** The event gets a boolean, so `missing` and `mismatch` both arrive as `false`. The verdict string costs nothing more (`"verification": verification.value`, next to or instead of the boolean) and a consumer or a probe can then key on the one that means "the file is gone" without a database lookup.

tests/utils/test_storage.py
```diff
@@ -21,3 +38,148 @@ def test_select_dist_storage(self):
...
+
+class StatObjectS3TestCase(SimpleTestCase):
+    """The S3 branch, stubbed against the real service model — so a parameter S3 does not have, or a
+    response member that is not one, fails here instead of in production."""
+
+    def _stubbed(self):
+        storage = storages.create_storage(S3_STORAGE["default"])
+        # the client stat_object caches on the storage, primed with a stub before it can build one
+        client = storage.connection.meta.client
+        storage._zentral_stat_client = client
```
np5, line 116: **nit:** Priming the cache with the storage's own client means `_s3_client(..., connect_timeout=, read_timeout=, retries=)` is never built by any test. The kwargs are valid botocore `Config` keys, but a future typo there would surface in production as an exception that `verify_upload` swallows into "undecided", with only a log line to find it by. One test that builds the real stat client and asserts `meta.config.connect_timeout == 2` and `retries["max_attempts"] == 1` closes that.

docs/apps/turbo.md
```diff
@@ -248,6 +368,52 @@ When `collect_inventory` is enabled, the agent posts a full machine inventory sn
...
+The outcome is a separate field on the upload, next to the status:
+
+|Verification|Meaning|
+|---|---|
+|`pending`|Not decided yet.|
+|`verified`|The stored object is the size and the SHA-256 the agent declared.|
+|`mismatch`|Something is at the key and it is not what the agent declared, or the result contradicts the request for the destination.|
+|`missing`|There is nothing at the key.|
+
+A `verified` or `missing` outcome also rides the result event, as `verified` on the `result.uploads` entry.
```
np5: **nit:** A `mismatch` rides the event too, as `verified: false`, which is what the code does at results.py:163. The sentence names `verified` and `missing` only.

np5 (commented):
Read the two commits that are this PR's own, `0c7a4254` and `bbfdc865`, on top of the base they were written against. Ran ruff (clean) and `mkdocs build --strict` (clean). CI on this head is green, but it ran at 08:00 UTC, before the fixes to #1682 and #1684 existed.

**This branch predates both fix commits, and GitHub says `CONFLICTING`.** `git range-diff` against `turbo-upload-round-trip` shows it carries the pre-rework SHAs of #1682 and of #1684, and lacks `b402614c` and `2bbc5ae0`, so the PR diff here spans the whole stack. A dry-run cherry-pick of your two commits onto `2bbc5ae0` conflicts in one file only, `tests/utils/test_storage.py`, in the import block. #1686 and #1687 sit on the same chain and will need the same rebase after this one.

Two things the rebase has to carry, because the files apply cleanly and nothing will flag them:

- `stat_object` must use the same normalised key the presign now signs. Inline, on `storage.py:147`.
- The CHANGELOG paragraphs were written under the 2026.5 Turbo block. The context they attach to moved to 2026.6 in the #1682 rework, so check where the hunk lands.

On the content itself: one design question and three small things, inline. The question is the synchronous storage round trip inside the results transaction. One operational note for the docs, from your own `test_stat_object_forbidden_is_not_missing`: a bucket policy without `s3:ListBucket` makes S3 answer 403 for a key that does not exist, so on such a deployment a genuinely missing object never becomes `missing`, it stays `pending` with an error log on every result. Worth one sentence next to the bucket rules.

What looked good: the two S3 traps are the right ones and both are pinned against the real service model; the 404-only branch, the composite-checksum guard and the refusal to hash a signing storage's object are all correct calls; closing the uploads before the wire ref is built, with a copied outcome, is the right fix for the aliasing class; and the two-leg comparison with the mint is sound. I will re-anchor whatever is still open once the rebased head is up.

zentral/contrib/turbo/public_views/results.py
np5: Agreed, closing this. In production the check is one HEAD per uploaded artifact, metadata only, with the digest coming back through the S3 checksum side channel. The body read exists for the fallback storage alone, where the ceiling is 10 MiB.

The outage case is also narrower than my arithmetic. An upload that fails during an outage comes back as an error entry, which is never verified. The HEADs that pay the connect timeout are only those of uploads that completed before the outage and whose results land during it, fifteen in one batch, from one machine. The agent retries the batch, and ingest is idempotent on run.at. Inline is the right call for this slice.

np5:
Rebased onto `41454e80`. The two commits are `3f9becec` and `6c92645d`, and the PR diff is now this PR's own work only.

Two conflicts, not one. `tests/utils/test_storage.py` in the import block, as you predicted — the base added `MAX_PENDING_UPLOADS` and my side adds `Stubber`, `ClientError`, `ContentFile`, `TestCase` and `UploadVerification`, so both resolutions are the union. `CHANGELOG.md` also conflicted, and that is the second item you flagged: the `HEAD` side of the 2026.5 Turbo block is now **empty**, because the #1682 rework moved the whole block to 2026.6. My paragraphs are under 2026.6.

**Both rebase-critical items are carried.**

`_object_key(storage, key)` now holds the normalisation, and the presign and the HEAD both call it, so the two cannot drift again. There is a Stubber test under `location="zentral"` that expects `Key: "zentral/turbo/uploads/test"`. `sha256_object()` needs nothing: it goes through `storage.open()`, which normalises for itself.

**Your nit about the stat client found a real bug.** The test that builds it asserted `retries["max_attempts"] == 1` and raised `KeyError`. botocore reads `max_attempts` as the number of *retries* and stores `max_attempts + 1`:

```
{"max_attempts": 1}       -> {"total_max_attempts": 2}
{"max_attempts": 0}       -> {"total_max_attempts": 1}
{"total_max_attempts": 1} -> {"total_max_attempts": 1}
```

So the ceiling allowed **two** attempts while its comment said one, and the worst case was twice the number I gave you in the design note. A bare `Config(retries=...)` keeps what you passed, so nothing shows the rewrite until a client exists. It is `total_max_attempts: 1` now, which is the key botocore keeps and means what it says, and the test asserts on that key.

Your suggestion about the verdict is in, and I took the "instead of" branch: the entry carries `verification` — `verified`, `mismatch` or `missing` — and no boolean. Two fields for one fact would age badly, truthiness is still one comparison, and `verified` never shipped, so nothing depends on it. That also answers the docs nit: the sentence names the field and its values, so there is no longer a case it fails to mention. An undecided verification puts no `verification` on the entry at all.

The `s3:ListBucket` note is in the docs, next to the storage rules, and in the CHANGELOG: without it S3 answers 403 for a key that does not exist, Zentral cannot tell a missing object from one it may not read, and the upload stays `pending` with an error in the log for every result.

Your own question about the synchronous round trip: I read it as closed by your follow-up, and I agree with the reasoning. An upload that fails during an outage comes back as an error entry, which is never verified.

Full suite green, 8647 tests. 100% patch coverage on every production line against `41454e80`. `ruff` and `mkdocs build --strict` clean.

#1686 and #1687 still sit on the old chain and need the same rebase.

np5:
Checked `6c92645d` on `41454e80`. The base rework is a pure reorganisation: its tree is identical to `2bbc5ae0`, the head I checked on #1684. The PR diff is now the two commits only.

All four threads are fixed, and I resolved them. Verified in the code: `_object_key` is the one place the key is built, the presign and the HEAD both call it, and the `location` test expects the prefixed key. The stat client is built by a test with its own ceiling, and that test found the botocore reading of `max_attempts` as retries; `total_max_attempts: 1` is the right key and the comment now says why. The event carries `verification` with the three outcomes and nothing when undecided, the docs and the changelog name the field and its values, the changelog paragraphs are under 2026.6, and the `s3:ListBucket` note is in. CI is green, patch coverage is full, ruff and the strict docs build are clean.

Nothing open from my side. Ready to merge behind #1684.

PR #1686, opened 2026-09-03
turbo: multipart uploads, and the server closes them

np5:
Stacked on #1685 — review that one first. A large artifact goes up in parts now, and Zentral closes the upload rather than the agent.

## `storage: the multipart control plane, and the declaration it rests on`

`create_multipart_upload()` declares `ChecksumAlgorithm: CRC64NVME`, and **that declaration is the one line of this design that must not be dropped.** Without it S3 accepts a *wrong* whole-object checksum at completion, and `HeadObject` afterwards reports a perfectly correct one because S3 computed its own — the object looks verified and is not. `ChecksumType: FULL_OBJECT` and `MpuObjectSize` genuinely are optional; do not read their optionality as license to drop the algorithm.

CRC-64/NVME rather than SHA-256 because S3 offers a full-object checksum on a multipart upload only for the CRC algorithms: only those compose from part checksums into a checksum of the whole object.

`ListParts` paginates. Reading a truncated first page as the whole list would assemble a partial object and call it done — and Zentral's own geometry cannot fill a page, which is exactly why a truncated answer must not be mistaken for a complete one.

The presigned part goes through the SigV4 client, like every presign here. A part carries no checksum on either storage, so its length is the entire contract, and the legacy V2 presign signs no header at all. Measured again while writing this: a raw client presigns V2 in `us-east-1`, so this is not hypothetical.

**Two client ceilings, and a correction to #1685's.** A call inside a device request keeps seconds; the assembly gets ten minutes, since both storages document that it can take several. And `max_attempts` in botocore counts **retries**, not attempts:

```
max_attempts=0 -> {'mode': 'standard', 'total_max_attempts': 1}
max_attempts=1 -> {'mode': 'standard', 'total_max_attempts': 2}
```

#1685's review had already fixed its stat client, with the better spelling: `total_max_attempts: 1`, the key botocore actually keeps, which cannot be misread the way `max_attempts` can. This PR adopts it for the assembly client too, and the tests assert that key. The client is renamed `_request_client`, because this PR gives it two more callers — starting and aborting a multipart upload — and `_stat_client` would have named only one of the three.

## `turbo: multipart, and the server closes it`

The mint answers `mode: "multipart"` with the geometry and one signed URL per part. **The geometry comes back with the URLs**, which is the whole reason it is not published or cached anywhere: the agent sends the same body either way and reads the mode off the answer, so there is nothing to predict, no part size to cache, and no stale-configuration two-step.

**The threshold is two part sizes**, 128 MB. At one there is no resumability to buy: a `part_size + 1` artifact is a two-part upload whose second part is one byte, so losing the first part costs the retransmission that restarting a single `PUT` would have — the whole multipart bill paid to protect a one-byte tail. At two, the first failure loses at most half the transfer.

**A multipart artifact is refused without `crc64nvme`.** `sha256` is composite-only on a multipart upload, so the whole-object CRC is the only thing that can validate the assembled object; without it completion would accept whatever arrived. Nothing uses the CRC on a single `PUT`, and a single `PUT` is not refused for want of it.

**Completion is the server's.** The agent posts the three keys that identify the upload and gets a `202`; Zentral lists the parts from the storage — never from the wire — and completes with its own credentials. That is the one place the agent would have needed storage-specific knowledge: collecting `ETag`s, building a completion body, and on S3 finding an error inside a `200 OK`. It runs in Celery because completion is documented as taking minutes, and the agent has no use for the answer anyway. Idempotent, so the call can be retried freely.

The task maps each answer onto the verification axis. A checksum that does not match the bytes is terminal — `assembly_failed` — because no retry improves it and retrying keeps the parts alive and billed. `NoSuchUpload` is ambiguous: the storage says it both for an upload that completed already and for one whose parts were aborted, and the object itself is the only thing that tells them apart. Everything else is transient and gets backoff.

**Result ingest does not HEAD a multipart row** — completion already validated the whole object. What it does instead is notice that a completion is still owed (the agent's call can have been lost, and a result is often the first thing to arrive) and drop the parts when the agent reports giving up.

A resume re-signs only the parts the agent names, against the row's **own** part size, fixed when the upload was created. It is refused when it declares a different `size` or `sha256`: those parts are the bytes the row describes, and re-signing against a new geometry would hand out parts that cannot line up with what is already in flight. I found that one reviewing my own first pass, which had no such check. The presented `upload_id` is checked against the row rather than believed.

Two small changes to shared code the `202` needed: the base POST view grows a `response_status`, and the `turbo_request` event posts on any 2xx rather than a literal `200` — otherwise an endpoint that only accepts the work would emit no event at all.

## Deployment

**The `AbortIncompleteMultipartUpload` lifecycle rule on the bucket is now load-bearing.** Zentral drops the parts when an agent reports giving up, but nothing else removes the parts of an upload the agent never reports, and parts are stored and billed until something does. It is in the docs and the changelog, but it is a deployment step and not only code.

## Not in this PR

**A sweep for multipart rows still `pending`** — raised in review, and a real gap. Only two things ask for a completion: the agent's `uploads/complete/` call and the result ingest. After the task's five retries are spent, or if both enqueues are lost, nothing asks again: the row stays `pending` for ever, counts against `MAX_PENDING_UPLOADS`, and its parts age out on the lifecycle rule leaving no object. The fix is a periodic re-enqueue and not a verdict on timeout — a storage that was unavailable is not a failed assembly, which #1685 settled — and this app already has the shape for it in `cleanup_turbo_machine_job_statuses`. Recorded in the plan as its own piece of work.

GCS multipart — its XML API is documented as compatible and the design accommodates it, but it is a second server-side implementation through a different client. And the console views to list and download an artifact.

## Tests

Full suite green — 8704 tests. 100% coverage on every line of the three modules this touches (`zentral/utils/storage.py`, `zentral/contrib/turbo/uploads.py`, the new `tasks.py`), verified from `coverage json` rather than from the diff, since a new file does not show up in a diff-hunk intersection. 71 new tests, `mkdocs build --strict` clean, all in-page anchors resolve.

Every S3 call is stubbed with `botocore.stub.Stubber` against the real service model, so a parameter S3 does not have — or a response member that is not one — fails here rather than in production. That matters more than usual for this PR: the declaration everything rests on is one S3 accepts silently when it is absent.

### DIFF
--- a/docs/apps/turbo.md
+++ b/docs/apps/turbo.md
@@ -383,6 +383,22 @@ Where the bytes go depends on the storage backend:
 
 The agent sees the same `mode: "put"` shape in both cases, and does not know which storage is behind it.
 
+#### Multipart
+
+A large artifact on an S3 storage backend goes up in parts. Zentral chooses that, not the agent: the mint response says `mode: "multipart"` and carries the geometry — the size of a part, and one signed URL for each part. The agent sends the same request either way and reads the mode off the answer, so there is nothing for it to predict and no part size for it to cache.
+
+Resumability is the reason. A 400 MiB `PUT` that fails at 95 % starts again from zero, where parts retry one at a time. An agent whose part URLs expired asks again with the `upload_id` and the numbers of the parts it still needs, and Zentral signs only those, against the part size the upload was created with.
+
+The limit is two part sizes of 64 MiB, so 128 MiB. At one part size there is nothing to gain: an artifact one byte over would be a two-part upload with a one-byte second part, and losing the first part costs the same as starting the single `PUT` again. Multipart begins to pay when a resume can save a full part.
+
+**Zentral completes the upload, not the agent.** When every part is up, the agent posts the three keys that identify the upload to [`/public/turbo/uploads/complete/`](#publicturbouploadscomplete) and gets a `202`. Zentral then lists the parts from the storage — it never takes a part list from a request — and completes the upload with its own credentials. This is the one place the agent would have needed to know something about the storage: completion means collecting entity tags, building a request body, and, on S3, finding an error inside a `200` answer.
+
+The completion runs in the background because both storages document that it can take several minutes. The agent has no use for the answer: it sent the bytes, and whether the storage assembles them is recorded on the upload's own verification field. The call is idempotent, so the agent can repeat it, and Zentral asks for a completion again by itself when a result arrives for an upload that is still waiting.
+
+A multipart upload is verified by its completion and not by a later question to the storage. The whole-object CRC-64/NVME the agent declared when it asked for the destination is supplied at completion, and the storage computes the assembled object's own and refuses a mismatch. A failure there is `assembly_failed`. This is why the config asks for that digest, and why an artifact large enough for multipart is refused without it: the SHA-256 cannot do the same work on a multipart upload.
+
+When the agent gives up on a multipart upload and says so in its result, Zentral drops the parts. Configure the `AbortIncompleteMultipartUpload` lifecycle rule on the bucket as well: it is the only thing that removes the parts of an upload the agent never reports, and parts are stored, and billed, until something removes them.
+
 [Config](#configuration) publishes `upload_max_size` and `upload_digests`. The agent can then refuse a file that is too large before it collects it, and compute the digests the storage wants in the same pass as the SHA-256.
 
 The result closes the upload. It carries a `run.id`, and one entry in `result.uploads` for each declared artifact. The entry reports the upload, or the reason it failed. Zentral compares the key the agent echoes with the key it minted, and refuses an echo that does not match: the key never comes from the wire.
@@ -404,9 +420,10 @@ The outcome is a separate field on the upload, next to the status:
 |Verification|Meaning|
 |---|---|
 |`pending`|Not decided yet.|
-|`verified`|The stored object is the size and the SHA-256 the agent declared.|
+|`verified`|The stored object is the size and the SHA-256 the agent declared, or a multipart upload assembled with the CRC-64/NVME it declared.|
 |`mismatch`|Something is at the key and it is not what the agent declared, or the result contradicts the request for the destination.|
 |`missing`|There is nothing at the key.|
+|`assembly_failed`|A [multipart](#multipart) upload could not be assembled. The parts do not make the object the agent described.|
 
 The outcome rides the result event as well, as `verification` on the `result.uploads` entry. It is the outcome itself and not a yes or no, because `mismatch` and `missing` are different problems and a consumer of the event cannot go to the database to tell them apart.
 
@@ -2157,7 +2174,9 @@ Request:
 |`artifact`|**Required.** The artifact name, as the job kind declares it.|
 |`size`|**Required.** The exact size of the file, in bytes.|
 |`sha256`|**Required.** The hex SHA-256 of the file. This is the identity of the artifact for Zentral.|
-|`digests`|Optional. The storage digests the config asked for, keyed by algorithm.|
+|`digests`|Optional. The storage digests the config asked for, keyed by algorithm. Required for an artifact large enough for [multipart](#multipart), which nothing else can validate.|
+|`upload_id`|Resume only. The `upload_id` of the multipart upload to continue. Zentral compares it with the one it holds.|
+|`missing_parts`|Resume only. The numbers of the parts that still need a URL. Send it with `upload_id`, never alone.|
 
 Response:
 
@@ -2179,6 +2198,30 @@ The agent sends the file in one `PUT` to `url`, with every header in `headers`, 
 
 `key` goes back to Zentral in the result. `expires_at` is when the signature stops being accepted; the signature is verified when the request starts, so an upload that begins in time can finish.
 
+For an artifact large enough for [multipart](#multipart), the answer carries the geometry and one URL per part instead of a single `url`:
+
+```json
+{
+    "mode": "multipart",
+    "key": "turbo/uploads/C02ZXXXXXXXX/b1d1a1f0-.../7/sysdiagnose_C02ZXXXXXXXX_20260820-091541.tar.gz",
+    "upload_id": "2~kEXAMPLEuploadIDexample",
+    "part_size": 67108864,
+    "parts": [
+        {"n": 1, "url": "https://bucket.s3.amazonaws.com/...&partNumber=1", "headers": {"Content-Length": "67108864"}},
+        {"n": 2, "url": "https://bucket.s3.amazonaws.com/...&partNumber=2", "headers": {"Content-Length": "31457280"}}
+    ],
+    "expires_at": "2026-08-20T10:15:41+00:00"
+}
+```
+
+|Attribute|Description|
+|---|---|
+|`part_size`|The size of every part but the last. The agent slices on it.|
+|`parts[].n`|The part number, which the storage needs and which a resume names.|
+|`parts[].headers`|The exact length of that part, signed. A part carries no checksum: the whole object is validated at completion.|
+
+The agent sends each part with one `PUT`, then tells Zentral it is done. See [`/public/turbo/uploads/complete/`](#publicturbouploadscomplete). A resume asks this endpoint again with `upload_id` and `missing_parts`, and gets back only the parts it named — signed against the `part_size` the upload already has, so they line up with what is in flight.
+
 The same request again, for the same run and the same artifact, returns the same `key` with a new signature. That is the retry: it replaces the object instead of adding a second one. Zentral counts the requests and stops at five.
 
 Errors:
@@ -2191,12 +2234,19 @@ Errors:
 |`400`|`{"error": "invalid_size"}`|`size` is absent, or not a positive integer.|
 |`400`|`{"error": "invalid_sha256"}`|`sha256` is absent, or not 64 hexadecimal characters.|
 |`400`|`{"error": "invalid_digests"}`|`digests` is not an object, or one of its values is not a base 64 digest.|
+|`400`|`{"error": "missing_digest"}`|The artifact is large enough for [multipart](#multipart) and `digests.crc64nvme` is absent. Nothing else can validate the assembled object.|
+|`400`|`{"error": "invalid_upload_id"}`|`upload_id` is present and is not a string.|
+|`400`|`{"error": "unknown_upload_id"}`|`upload_id` is not the one of the multipart upload Zentral holds for that run and artifact.|
+|`400`|`{"error": "invalid_missing_parts"}`|`missing_parts` is not a non-empty list of part numbers.|
+|`400`|`{"error": "missing_upload_id"}`|`missing_parts` is present without `upload_id`. A resume names the upload it resumes.|
+|`400`|`{"error": "artifact_changed"}`|A resume declares a different `size` or `sha256` than the multipart upload in flight. That is a different artifact, and it needs its own upload.|
 |`400`|`{"error": "unknown_artifact"}`|The job does not declare that artifact. Terminal – no retry changes it.|
 |`400`|`{"error": "attempts_exhausted"}`|Five destinations were already minted for that run and artifact.|
 |`404`|`{"error": "unknown_schedule"}`|The schedule is unknown, belongs to another configuration, is out of scope, or is outside its delivery window.|
 |`410`|`{"error": "gate_closed"}`|A result for that one-time job already came back. The job is done on this machine, and an artifact uploaded now could never be referenced.|
 |`410`|`{"error": "already_uploaded"}`|The result already reported that artifact as uploaded.|
 |`413`|`{"error": "too_large"}`|`size` is above the limit of the job kind, or above the limit of the deployment.|
+|`503`|`{"error": "storage_unavailable"}`|The storage did not answer when Zentral started the [multipart](#multipart) upload. Ask again later.|
 |`429`|`{"error": "too_many_pending"}`|The machine holds too many destinations on that schedule that no result has closed. Report the runs already made; a request for a destination the machine already has is never refused for this reason.|
 
 #### The hosted destination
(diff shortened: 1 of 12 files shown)


zentral/contrib/turbo/tasks.py
```diff
@@ -0,0 +1,125 @@
...
+                       UploadVerification.ASSEMBLY_FAILED)
+    raise task.retry(exc=error, countdown=60 * 2 ** task.request.retries)
+
+
+def _record(upload, result, verification):
+    # the verification axis only. The status axis closed when the agent reported, and an assembly that
+    # failed does not reopen a one-time job any more than a mismatch does.
+    upload.verification = verification
+    upload.verified_at = timezone.now()
+    upload.save()
```
np5: **issue:** This `save()` writes every field of a row the task loaded minutes earlier, and the results ingest writes the same row in between. The sequence is the normal one: the agent calls `uploads/complete/`, the task loads the row with `status=pending`, `ListParts` and `CompleteMultipartUpload` run for the several minutes the docs allow, and the agent posts its result meanwhile. The ingest sets `status=uploaded`, `completed_at` and `truncated`, and commits. The task then finishes and saves its stale copy: `status` goes back to `pending`, `completed_at` and `truncated` are lost, and only `verification` is new. The result was acknowledged, so the agent never resends it, and the row stays `pending` for good. It counts against `MAX_PENDING_UPLOADS`, and the console will show a report that did arrive as one that did not.

The tests cannot see it: the storage calls are mocked, so the task has no window.

`update_fields` on both writers closes it. Here `upload.save(update_fields=["verification", "verified_at"])`; in `_close_uploads` the status fields only (`status`, `error_reason`, `truncated`, `completed_at`, plus the verification pair on the single-PUT path, which decides inline). The reverse race, the ingest overwriting a verdict the task committed during its own short window, heals itself through the re-enqueue and the `NoSuchUpload` branch, and the same change removes it too.

zentral/contrib/turbo/public_views/uploads.py
```diff
@@ -102,14 +153,17 @@ def do_post(self, data):
         upload.sha256 = sha256
         upload.crc64nvme = digests.get("crc64nvme", "")
         upload.crc32c = digests.get("crc32c", "")
+        if upload.mode == UploadMode.MULTIPART and not upload.upload_id:
+            start_multipart_upload(upload, artifact, storage)
```
np5: **issue:** `create_multipart_upload` is a network call to S3 inside the mint, and it is the only storage call on a device endpoint that nothing wraps. A `ClientError`, a connect timeout or a read timeout here is an unhandled exception: Django answers with an HTML 500, on the endpoint this PR's own description says must never serve one, and `ATOMIC_REQUESTS` rolls the row back while S3 may already hold the new upload id, an orphan the lifecycle rule has to sweep. Everything else in this plane maps a storage failure to a JSON answer the agent can act on.

Catching `ClientError` and `BotoCoreError` around this call and raising a `WireError` with a retry-later status (429 is already "ask again" on this endpoint, or a 503) keeps the contract. The agent then retries the mint, which is the right outcome for a storage that did not answer.

zentral/contrib/turbo/tasks.py
```diff
@@ -0,0 +1,125 @@
...
+    if code == "NoSuchUpload":
+        # the upload id is gone, which is what the storage says BOTH for one that completed already
+        # and for one whose parts were aborted. The object itself is the only thing that tells them
+        # apart, and a second completion call is exactly how the first case arrives here.
+        try:
+            stored = stat_object(upload.key, storage=storage)
+        except Exception:
+            logger.exception("Could not stat turbo upload %s", upload.pk)
+            raise task.retry(exc=error, countdown=60 * 2 ** task.request.retries)
+        if stored is not None and stored["size"] == upload.size:
```
np5: **suggestion:** The object at the key is taken as the completed upload on its size alone. `stat_object` already returns the `ChecksumCRC64NVME` the HEAD reports with `ChecksumMode=ENABLED`, and the row holds the CRC the agent declared, so `stored["crc64nvme"] == upload.crc64nvme` is an exact match for free where the size is a coincidence. Fall back to the size only when the HEAD carries no checksum.

docs/apps/turbo.md
```diff
@@ -2191,6 +2208,12 @@ Errors:
 |`400`|`{"error": "invalid_size"}`|`size` is absent, or not a positive integer.|
 |`400`|`{"error": "invalid_sha256"}`|`sha256` is absent, or not 64 hexadecimal characters.|
 |`400`|`{"error": "invalid_digests"}`|`digests` is not an object, or one of its values is not a base 64 digest.|
+|`400`|`{"error": "missing_digest"}`|The artifact is large enough for [multipart](#multipart) and `digests.crc64nvme` is absent. Nothing else can validate the assembled object.|
+|`400`|`{"error": "invalid_upload_id"}`|`upload_id` is present and is not a string.|
```
np5, line 2238: **nit:** The errors name `upload_id` and `missing_parts`, but the request table above does not list them, and the response example is the single-PUT shape only. A `mode: "multipart"` example with `upload_id`, `part_size` and `parts[].n / url / headers`, plus the two resume fields in the request table, would give the agent team the shape from the reference rather than from the prose.

docs/apps/turbo.md
```diff
@@ -383,6 +383,22 @@ Where the bytes go depends on the storage backend:
 
 The agent sees the same `mode: "put"` shape in both cases, and does not know which storage is behind it.
 
+#### Multipart
+
+A large artifact on an S3 storage backend goes up in parts. Zentral chooses that, not the agent: the mint response says `mode: "multipart"` and carries the geometry — the size of a part, and one signed URL for each part. The agent sends the same request either way and reads the mode off the answer, so there is nothing for it to predict and no part size for it to cache.
+
+Resumability is the reason. A 400 MB `PUT` that fails at 95 % starts again from zero, where parts retry one at a time. An agent whose part URLs expired asks again with the `upload_id` and the numbers of the parts it still needs, and Zentral signs only those, against the part size the upload was created with.
+
+The limit is two part sizes, so 128 MB. At one part size there is nothing to gain: an artifact one byte over would be a two-part upload with a one-byte second part, and losing the first part costs the same as starting the single `PUT` again. Multipart begins to pay when a resume can save a full part.
```
np5: **nit:** "128 MB" and "400 MB" here, where the constants are 64 × 2²⁰ and the changelog says MiB. The rest of this page uses MiB for the 10 MiB limit.

np5 (commented):
Read the two commits on the rebased base, which is the reviewed #1685 head. Ran ruff (clean) and `mkdocs build --strict` (clean). CI on `669940fc` was still running when I posted this; the previous head was green.

Because the Stubber tests validate parameters against the service model and not S3's behaviour, I checked the semantics this design rests on against the AWS documentation: part-level checksums are optional when the upload uses a full-object checksum type, so a part that carries only its length is legal; the completion validates the whole-object value and answers `BadDigest` on a mismatch; an `x-amz-mp-object-size` mismatch is `InvalidRequest`, which your terminal set already holds; and the docs state that a completion checksum is accepted but not validated when the algorithm was not declared at creation, which is the "one line that must not be dropped". The control plane is right.

Two things to fix before merge, both in the plumbing rather than the design:

- `zentral/contrib/turbo/tasks.py`: the completion task saves the whole row minutes after loading it, and the results ingest writes the same row in between. The normal sequence leaves the row `pending` for good after its result was acknowledged. `update_fields` on both writers.
- `zentral/contrib/turbo/public_views/uploads.py`: `create_multipart_upload` is the one storage call on a device endpoint that nothing wraps, so a storage that does not answer becomes an HTML 500 on the mint.

Smaller: the `NoSuchUpload` branch can compare the CRC the HEAD returns instead of the size alone; the endpoint reference lacks the multipart response shape and the resume request fields; two unit slips in the docs. The PR description still says both clients pass `max_attempts` of `0`, while the code passes `total_max_attempts: 1`, which is the right spelling and the one the tests assert.

One question for the "not in this PR" list: after the task's five retries, or a lost enqueue, nothing asks for a completion again except another result or another call from the agent. A periodic sweep of multipart rows still `pending` would be the backstop, the way the lifecycle rule is for the parts.

What looked good: the threshold argument, the geometry riding the response, completion on the server with the part list taken from the storage, the ambiguity of `NoSuchUpload` resolved by looking at the object, the resume refusing a changed artifact, and the two client ceilings with the retries spelled as a total.

zentral/contrib/turbo/tasks.py
np5: Fixed in 9ec0e27b, and you had the sequence exactly right. I confirmed it by reverting the fix and running the new test: `AssertionError: 'pending' != UploadStatus.UPLOADED`.

`update_fields` on both writers, each naming only the axis it owns — the task writes `verification` / `verified_at` / `updated_at`, `_close_uploads` writes the status fields plus the verification pair on the single-PUT path where it decides inline.

On the test being blind to it: `test_a_result_that_lands_during_the_assembly_survives_it` gives the storage call a real window, by having the `complete_multipart_upload` mock close the status axis while the completion is notionally in flight. It fails on the old code, which is the only reason to trust it.

zentral/contrib/turbo/public_views/uploads.py
np5: Fixed in 9ec0e27b. `ClientError` and `BotoCoreError` around the call, and a `WireError("storage_unavailable", status=503)`.

I went with 503 rather than 429 because 429 already carries `too_many_pending` on this endpoint, and that one is about the agent — it has too many rows open — where this is about the storage. The wire doc classes `429 / 5xx / network` together as retry-later, so the agent behaviour is the same either way, and the code says which happened. Documented in the mint error table.

Two tests: a connect failure and a refusal, both asserting the 503 and the JSON body.

zentral/contrib/turbo/tasks.py
np5: Adopted in 9ec0e27b, as `_is_the_completed_object()`: the CRC decides when the HEAD carries one and the row has one, and the size is the fallback otherwise.

Two tests for the two directions — an object of the right size whose CRC differs is no longer taken for the completed upload, and a matching CRC decides even when the size disagrees.

docs/apps/turbo.md
np5: Added in 9ec0e27b: a `mode: "multipart"` response example with `upload_id`, `part_size` and `parts[]`, a table for `part_size` / `parts[].n` / `parts[].headers` — including that a part carries no checksum and why — and `upload_id` / `missing_parts` in the request table, marked resume-only. Plus a line on what a resume gets back and what it is signed against.

docs/apps/turbo.md
np5: Fixed in 9ec0e27b — 400 MiB and "two part sizes of 64 MiB, so 128 MiB", so the page agrees with the constants and with the changelog.

np5:
All five addressed in 9ec0e27b, and the PR description is corrected — it still claimed both clients pass `max_attempts: 0`, which was my spelling before the rebase picked up yours.

**The lost update was the real one.** Your sequence was right, and I only trust the new test because it fails on the old code: reverting the `update_fields` gives `AssertionError: 'pending' != UploadStatus.UPLOADED`. Worth noting it was not reachable before this PR — the single-PUT path decides its verification inline, inside the same request that writes the status, so there was no second writer and no window. Multipart introduced both.

**On the sweep.** You are right that nothing asks again once the five retries are spent or both enqueues are lost, and the row then stays `pending` for ever while its parts age out on the lifecycle rule — leaving neither an object nor a verdict, and a row that counts against `MAX_PENDING_UPLOADS` for good.

I have kept it out of this PR and recorded it in the plan, for one reason worth stating: the fix is a periodic **re-enqueue**, not a verdict on timeout. A storage that was unavailable is not a failed assembly, which #1685 settled for the HEAD path and which holds here for the same reason — so the sweep must not be tempted to write `assembly_failed` when it gives up. That makes it a small piece of work with a rule of its own, and this app already has the shape for it in `cleanup_turbo_machine_job_statuses`. Bounded by age, and capped per pass.

Full suite green at 8704, and zero missing lines in every module this touches.

np5:
Checked `9ec0e27b`. All five threads are fixed, and I resolved them.

Verified in the code: the task writes only the verification pair and the ingest only the status axis, both through `update_fields`, and the new test closes the status axis inside the mocked completion, which is the exact window the docs allow. Starting the multipart upload answers `503 storage_unavailable` on a storage that does not answer or refuses, and the half-built row that persists is a normal pending row the retry mint finishes. The `NoSuchUpload` branch identifies the object by its CRC with the size as fallback. The reference carries the multipart response, the two resume fields and MiB units, and the description now tells the retries story as the code has it.

The test run on this head was still in progress when I checked; the previous head was green. Ruff and the strict docs build are clean.

Nothing open from my side. One note for the list of things not in this PR: a completion that exhausted its retries, or whose enqueue was lost, is only asked for again by another result or another call from the agent. A periodic sweep of multipart rows still pending would be the backstop.

zentral/contrib/turbo/tasks.py
np5: Now in `bbcb618e` — the `update_fields` on both writers, and the test that fails without them. The review commit was folded into `turbo: multipart, and the server closes it`, so the `9ec0e27b` above is no longer on the branch; the fold changed no content.

The reasoning moved into that commit message rather than staying a fix, since after folding there was never a bug: two writers on two axes, with the minutes-long completion as the reason. It also records that the single-PUT path never needed this, which is what made the hazard new.

zentral/contrib/turbo/public_views/uploads.py
np5: Now in `bbcb618e` — the `503 storage_unavailable` around the multipart create. The review commit was folded into `turbo: multipart, and the server closes it`, so the `9ec0e27b` above is no longer on the branch; the fold changed no content.

zentral/contrib/turbo/tasks.py
np5: Now in `bbcb618e` — `_is_the_completed_object`, comparing the whole-object CRC with the size as fallback. The review commit was folded into `turbo: multipart, and the server closes it`, so the `9ec0e27b` above is no longer on the branch; the fold changed no content.

docs/apps/turbo.md
np5: Now in `bbcb618e` — the `mode: "multipart"` response example, the `part_size` / `parts[].n` / `parts[].headers` table, and the two resume fields in the request table. The review commit was folded into `turbo: multipart, and the server closes it`, so the `9ec0e27b` above is no longer on the branch; the fold changed no content.

docs/apps/turbo.md
np5: Now in `bbcb618e` — 400 MiB, and "two part sizes of 64 MiB, so 128 MiB". The review commit was folded into `turbo: multipart, and the server closes it`, so the `9ec0e27b` above is no longer on the branch; the fold changed no content.

np5:
Folded the review commit into the two it belonged to, so the branch is back to:

```
bbcb618e  turbo: multipart, and the server closes it
8a368975  storage: the multipart control plane, and the declaration it rests on
```

All of it went into the second one — the fixes touched six files, none of them `zentral/utils/storage.py` or `tests/utils/test_storage.py`, so there was nothing to split.

**The tree is unchanged by the fold**: `git diff 9ec0e27b bbcb618e` is empty, which is the check worth reporting, since a force-push over a branch you have already read is otherwise something you have to take on trust. 849 turbo + utils tests pass on it, and CI is re-running.

The commit message absorbed the reasoning rather than keeping it as a fix, because after folding there was never a bug to fix. Three things earned a place in it: why each writer of the row names the fields it owns (two axes, and a completion that runs for minutes), the `503` on the mint, and the whole-object CRC as what tells the two `NoSuchUpload` cases apart.

`9ec0e27b` no longer exists on the branch, so my earlier replies cite a commit you cannot check out. I have noted the surviving SHA on each thread.

PR #1687, opened 2026-09-03
turbo: see and fetch what a job collected

np5:
Stacked on #1686 — review that one first. This is where the artifacts become visible: the page of an enrolled machine lists what its jobs collected, and hands over a file to the people who may read one.

## Two actions, differing by group

They share one `applies_to` and are deliberately not the same act:

|Action|Group|What it authorizes|
|---|---|---|
|`viewJobUpload`|Admin, User, **Viewer**|See that a job collected an artifact, and what became of it.|
|`downloadJobUpload`|Admin, User|Fetch it.|

Seeing that a sysdiagnose exists is an ordinary view. Reading it hands over the data of the person who uses the machine. The distinction the design prose already drew, made structural rather than left to policy — a deployment that never writes one still gets it.

The USER-only roster guard in `tests/server_accounts/test_pbac_engine.py` caught this, correctly: every USER-but-not-VIEWER action has to be declared there. The decision is now recorded next to the MDM secret readers rather than being a quiet asymmetry.

**An artifact the user may not see is absent from the page**, not merely unlinked — so its name and its size do not leak either. Two decisions per row, batched into one evaluation the way the scheduling rows already are: the rows differ by machine, by configuration and by job, so a single answer for the page would be wrong.

## Two containers, because they answer different questions

The resource is the upload, and it is a member of both:

 * **`Inventory::Machine`** — whose data this is, and it brings meta business unit scoping with it.
 * **`Turbo::Configuration`** — whose collection produced it.

That second one is what lets a single policy cover the whole lifecycle:

```
forbid (
  principal,
  action in [Turbo::Action::"createOneTimeJob", Turbo::Action::"updateOneTimeJob",
             Turbo::Action::"deleteOneTimeJob", Turbo::Action::"viewJobUpload",
             Turbo::Action::"downloadJobUpload"],
  resource in Turbo::Configuration::"…"
);
```

With only the machine parent, that policy would silently cover the scheduling and miss the artifacts. Both chains have a test, and so does `resource.job.kind` / `resource.artifact`.

**And the limit of the first chain, which I would rather write down than have someone find.** The meta business units of a machine come from its inventory **snapshots**, not from the enrollment secret — so a machine that has enrolled and never posted inventory is in none of them as far as a policy is concerned. My first version of the MBU test failed for exactly that reason. There is now a test for each case, and the docs say it: MBU scoping alone is not the boundary, which is precisely why the configuration parent exists.

I checked the schema before building on it:

```
JobUpload memberOfTypes: ['Inventory::Machine', 'Turbo::Configuration']
JobUpload attrs:         ['artifact', 'job']
```

## The rest

`resolve_upload_schedules()` resolves the schedule polymorphically — the row is keyed like the wire and holds no foreign key to either scheduling model — and one lookup yields the configuration and the job, which happen to be the two things both the page and a decision need. An upload whose schedule has since been deleted is dropped: the row survives, because the artifact is still in the bucket, but there is no configuration to name and no job to read, so there is nothing to decide against.

The download signs `ResponseContentDisposition`. The key already ends in the artifact's filename, so a browser saves the right name from the path alone — but a small `manifest.json` would open in the tab instead of being saved. A `404` for a missing object is expected rather than exceptional: the expiry rule removes objects out from under rows that outlive them.

The machine page itself is still gated by the global `turbo.view_enrolledmachine` permission. That is defence in depth and not a boundary; scoping the machine views by MBU is separate work, and out of scope here.

## Tests

Full suite green — 8703 tests. 100% coverage on every line this touches, checked from `coverage json` per module as well as by diff intersection. 21 new tests, `mkdocs build --strict` clean, all in-page anchors resolve.


zentral/contrib/turbo/views/enrolled_machines.py
```diff
@@ -70,9 +78,55 @@ def get_context_data(self, **kwargs):
...
+
+    view_enrolledmachine gates the page and this gates the file, which are two different questions:
+    the first is whether the machine page exists for you at all, the second is whether the data
+    inside a sysdiagnose is yours to read. The typed action decides the second.
+    """
+    permission_required = "turbo.view_enrolledmachine"
+
+    @cached_property
+    def _file_storage(self):
+        return select_dist_storage()
```
np5: **issue:** The download reads from `select_dist_storage()`, and every write in this plane goes to `storages["default"]`: the mint presigns against it, the hosted PUT saves to it, the verification HEAD and the multipart completion read it. `select_dist_storage` exists for the MDM artifacts, and the SaaS configures a separate `dist` storage for them, so on those deployments the presigned GET points at a bucket the artifact was never written to, and every download answers a 404 from S3 after the redirect. On a deployment without `dist` the two storages coincide, which is why the tests pass. `storages["default"]` here, and the same for the `exists()` and `open()` fallback.

zentral/utils/storage.py
```diff
@@ -311,6 +311,17 @@ def abort_multipart_upload(key, upload_id, storage):
...
 
+def generate_presigned_get(key, filename, storage):
+    """A signed GET that saves the object instead of rendering it.
+
+    The key already ends in the artifact's filename, so a browser saves the right name from the path
+    alone — but a small manifest.json would open in the tab rather than be saved, and the disposition
+    is what settles that. It is signed with the URL, so it cannot be stripped off on the way.
+    """
+    return storage.url(key, parameters={
+        "ResponseContentDisposition": f'attachment; filename="{filename}"'})
```
np5, line 329: **issue:** `file_storage_has_signed_urls` is true for the GCS backends too, so this is the branch a GCP deployment takes, and it raises there. `ZentralGoogleCloudStorage.url(self, name)` takes no `parameters` argument, so the call is a `TypeError`, a 500 on every download. Even through django-storages' own `GoogleCloudStorage.url(name, parameters)`, the dict is passed to `generate_signed_url` as keyword arguments, which spells this option `response_disposition`, so it fails there as well. Only the S3 spelling is tested.

Two ways: give the GCS subclass the `parameters` passthrough and map the option per backend (`ResponseContentDisposition` on S3, `response_disposition` on GCS), or hand the disposition to S3 only and fall back to a plain `storage.url(key)` on GCS, accepting that a small manifest opens in the tab there. A test with the GCS backend class, like the ones in `tests/utils/test_storage.py`, would have caught it.

zentral/contrib/turbo/pbac.py
```diff
@@ -58,6 +95,15 @@ def get_one_time_job_resource(one_time_job) -> Resource:
                     attrs={"job": get_job_resource(one_time_job.job)})
 
 
+def get_job_upload_resource(upload, configuration, job) -> Resource:
+    return Resource(
+        "JobUpload", str(upload.pk), get_namespace(),
+        [get_meta_machine_resource(MetaMachine(upload.serial_number)),
```
np5: **suggestion:** A new `MetaMachine` per row, so `meta_business_units` runs its query once per upload, and on the machine page every row has the same serial. `get_meta_machine_resource` caches on the machine object, so one `MetaMachine` per distinct serial, built by `authorize_job_upload_rows` and passed in, turns N queries into one for the page.

zentral/contrib/turbo/views/enrolled_machines.py
```diff
@@ -70,9 +78,55 @@ def get_context_data(self, **kwargs):
...
         ctx["selected_kind"] = kind
+        # the artifacts this machine has been asked for, newest first. Filtered by policy rather than
+        # merely unlinked: an upload the user may not see is absent from the page.
+        uploads = (JobUpload.objects.filter(serial_number=serial_number)
+                                    .order_by("-created_at"))
+        if kind:
+            schedules = resolve_upload_schedules(uploads)
+            uploads = [upload for upload in uploads
+                       if (schedules.get(upload.schedule_pk) or (None, None))[1]
+                       and schedules[upload.schedule_pk][1].kind == kind]
```
np5: **nit:** With a `kind` filter the schedules are resolved twice, here and again inside `authorize_job_upload_rows`, and the comprehension is hard to read. Filtering the authorized rows afterwards on `upload.job.kind`, which `authorize_job_upload_rows` sets, needs neither.

zentral/contrib/turbo/pbac.py
```diff
@@ -228,6 +332,44 @@ def offerable_jobs(user_obj, configuration, queryset):
     engine.authorize_requests(requests)
     return [job for job, pbac_request in zip(jobs, requests) if pbac_request.is_authorized]
 
+def check_download_job_upload(request, upload, configuration, job):
```
np5, line 338: **nit:** One blank line before this definition where the module has two everywhere else. Ruff does not flag it, its blank-line rules are preview-only, which is why it slipped through.

docs/apps/turbo.md
```diff
@@ -397,7 +397,22 @@ The completion runs in the background because both storages document that it can
...
-When the agent gives up on a multipart upload and says so in its result, Zentral drops the parts. Configure the `AbortIncompleteMultipartUpload` lifecycle rule on the bucket as well: it is the only thing that removes the parts of an upload the agent never reports, and parts are stored, and billed, until something removes them.
+When the agent gives up on a multipart upload and says so in its result, Zentral drops the parts. That report can be a day behind, though, and it may never arrive at all — so the bucket needs a lifecycle rule as well. See [Bucket lifecycle](#bucket-lifecycle).
+
+#### Bucket lifecycle
+
+Zentral writes the artifacts and never removes them, and a bucket collects other things that only the bucket itself can clean up. Set these rules on the bucket if you manage your own deployment.
+
+|Rule|Scope|Why|
+|---|---|---|
+|`Expiration`|`turbo/uploads/`|Zentral keeps no expiry of its own, so nothing removes a collected artifact. They hold the data of the person who uses the machine, and they are collected to answer one question: keep them for as long as answering it takes. 14 days is a good default.|
```
np5, line 408: **nit:** The keys live under the storage's `location` when one is set, so the prefix of this rule is `<location>/turbo/uploads/` there. Worth one clause, since a rule on the bare prefix never matches and nothing reports that the artifacts are not expiring.

np5 (commented):
Read the two commits on the rebased base, whose tree is identical to the #1686 head I checked. Ran ruff (one finding, the unused import in `test_pbac_schema.py` that is already on main) and `mkdocs build --strict` (clean). The test run on this head was still in progress when I posted.

The authorization design is right and well pinned: two actions on one resource, differing by group, with the roster guard recording why; the two parents, each with a test showing which policy shape it serves; the MBU limit written down instead of discovered; rows absent rather than unlinked. The findings are in the download path.

- `zentral/contrib/turbo/views/enrolled_machines.py`: the download reads from the `dist` storage while everything else in the plane writes to `default`. The SaaS has a separate `dist` storage, so downloads there point at the wrong bucket.
- `zentral/utils/storage.py`: the signed GET with a disposition calls `url()` with a `parameters` argument the GCS subclass does not take, and a spelling its client does not know. GCP deployments take exactly this branch and get a 500.

Smaller: one `MetaMachine` per row on a page where every row is the same machine; the `kind` filter resolves the schedules twice; a blank line; and the lifecycle rule's prefix under a storage `location`.

What looked good beyond the design: `resolve_upload_schedules` giving one lookup for both the page and the decision, the deleted-schedule case dropping the row rather than guessing, and the lifecycle docs covering the three rules with the reasons for each.

zentral/contrib/turbo/views/enrolled_machines.py
np5: Fixed in df96d5c3 — you are right, and the reasoning about why the tests could not see it is right too.

`storages["default"]` now, for the redirect and for the `exists()`/`open()` fallback. I had copied the MDM download pattern without checking which storage this plane actually writes to.

`test_the_download_reads_the_storage_the_plane_writes_to` configures a separate `dist` storage. I checked it fails without the fix (`302 != 200`) rather than passing for free.

zentral/utils/storage.py
np5: Fixed in df96d5c3. I verified both halves before choosing:

```
ZentralGoogleCloudStorage.url:            (self, name)
django-storages GoogleCloudStorage.url:   (self, name, parameters=None)
generate_signed_url disposition params:   ["response_disposition", "response_type"]
```

So the subclass raises on the argument and the plain one would raise on the option name.

I took your second option: the disposition is signed on S3 and nowhere else, and a GCS deployment gets a plain signed URL with a manifest that opens in the tab. Mapping the option per backend means teaching `ZentralGoogleCloudStorage.url` a `parameters` passthrough it overrides away, for one caller — not worth it against a cosmetic difference on one backend.

`test_a_storage_that_cannot_take_parameters_gets_a_plain_url` uses the GCS backend class and asserts `url()` is called with the key alone. It fails without the fix.

zentral/contrib/turbo/pbac.py
np5: Done in df96d5c3. `authorize_job_upload_rows` builds one `MetaMachine` per distinct serial and passes it into both requests, so the parents are resolved once for the page. `get_job_upload_resource` and the request classes take the machine rather than making one.

`check_download_job_upload` builds its own, which is right there: one row, one machine.

zentral/contrib/turbo/views/enrolled_machines.py
np5: Done in df96d5c3. The rows are authorized first and the filter runs on `upload.job.kind`, which `authorize_job_upload_rows` already set — one resolution instead of two, and the comprehension is gone.

zentral/contrib/turbo/pbac.py
np5: Fixed in df96d5c3.

docs/apps/turbo.md
np5: Added in df96d5c3: the section now says the keys sit under the storage `location` when one is set, so the rule needs `<location>/turbo/uploads/` there — and that a rule on the bare prefix matches nothing, with nothing reporting that the artifacts are not being removed. That last part is what makes it worth saying.

np5:
Checked `df96d5c3`. All six threads are fixed.

Verified in the code: the download reads `storages["default"]` for the redirect and for the fallback, and the split-storage test is the one that fails without it. The disposition is signed on S3 only, and the GCS subclass test pins that its `url()` is called with the key alone. One `MetaMachine` per serial, the `kind` filter after the decision, the blank lines, and the `location` clause in the lifecycle docs are all in. Ruff shows only the unused import that is already on main, and the strict docs build is clean. The test run on this head was still in progress when I checked; the previous head was green.

One note for the GCS single-PUT work, not for this PR: the disposition branch keys on `file_storage_has_presigned_uploads`. The moment GCS joins that predicate, the S3 spelling goes to GCS again. Key it on the S3 classes, or map the option per backend, when that lands.

Nothing open from my side. Ready to merge behind #1686.

PR #1690, opened 2026-09-04
mdm: add a command to re-poll the app installs stuck at AwaitingConfi…

sebastian-zentral:
## Why

#1572 fixed the poll schedules of the store app and enterprise app install checks, but left the target artifacts that were already stuck at `AwaitingConfirmation` where they were. Nothing revisits them: the install decision in `artifacts.py` only acts on an `Uninstalled` or force-installed artifact version, and only the two install commands ever queue a `ManagedApplicationList` or `InstalledApplicationList` check. On an affected device the app is installed, but its artifact row still reads AwaitingConfirmation, indefinitely.

A forced install (#1654) resolves it, but re-sends the install command, bumps the retry and install counters, and takes one call per device and artifact. A plain re-poll is enough: where the app did install, the first response reports `Managed` and the status flips to `Installed`.

## What

`recheck_app_installs` finds the app target artifacts at AwaitingConfirmation that have no live check, queues the check the install would have queued, linked to the artifact version so the acknowledgement updates the status, and notifies each target once over APNs. Both channels are covered.

A check that is queued, in flight, or answered `NotNow` is sent or re-sent at the next connection, so it counts as live and the artifact is left alone. Without the NotNow case a device in a login-window state would get a duplicate check on every run.

```
python server/manage.py recheck_app_installs --dry-run
python server/manage.py recheck_app_installs --serial-number C02XXXXXXXXX
python server/manage.py recheck_app_installs --no-notification
```

The reusable pieces live in `zentral/contrib/mdm/app_install_checks.py`: the detection queryset, the check builder per artifact type, the queueing with a target compatibility check, and the notification. The management command only formats the output. It reports a skipped row with the reason, and ends with a count.

## Notes for the reviewer

- No `retries` value is passed, like the install commands, so a re-poll walks the full new backoff schedule.
- A store app whose asset has no bundle ID is skipped with "Asset without bundle ID". That is the #1550 case: run `refresh_apps_books_asset_metadata` first, then rerun.
- No docs change, following the sibling commands, of which only `mdmcerts` is documented.

## Tests

`tests.mdm.test_app_install_checks` and `tests.mdm.management_commands.test_recheck_app_installs`, 31 tests. Detection covers the status and type filters, the three pending-command shapes, the user channel and the serial filter. The command tests cover every flag, the one-notification-per-device behaviour and the skipped output. Run together with the two poll command modules: 70 tests OK. ruff clean.

### DIFF
--- a/tests/mdm/management_commands/test_recheck_app_installs.py
+++ b/tests/mdm/management_commands/test_recheck_app_installs.py
@@ -1,0 +1,135 @@
+from io import StringIO
+from unittest.mock import patch
+from django.core.management import call_command
+from django.test import TestCase
+from django.utils.crypto import get_random_string
+from zentral.contrib.inventory.models import MetaBusinessUnit
+from zentral.contrib.mdm.models import Artifact, DeviceArtifact, DeviceCommand, TargetArtifact
+from ..utils import force_artifact, force_dep_enrollment_session
+
+
+class RecheckAppInstallsCommandTest(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.mbu = MetaBusinessUnit.objects.create(name=get_random_string(12))
+        cls.mbu.create_enrollment_business_unit()
+        cls.dep_enrollment_session, _, _ = force_dep_enrollment_session(
+            cls.mbu, authenticated=True, completed=True, realm_user=True
+        )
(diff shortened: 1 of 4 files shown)


sebastian-zentral:
Closing for now, because it is a one time thing.

PR #1693, opened 2026-09-11
Bump PyJWT to 2.14.0, and do not use the HMAC algorithms an OpenID Connect issuer advertises

np5:
Two commits.

### Bump PyJWT to 2.14.0

A security release. It closes twelve advisories against 2.13.0, one of them critical (GHSA-ffc3-869f-jxw9): a public key with modified PEM armour, with different whitespace or CRLF line endings, moves past the guard in `HMACAlgorithm.prepare_key`. The key is accepted as an HMAC secret, and an attacker can forge an HS256 token. The release also hardens the other variants of the same attack (DER, BOM prefix, JWK and JWKS containers, arrays, encoded material), and the empty HMAC key that an `oct` JWK supplied.

That family of attacks is not possible here: `verify_jws` gives `jwt.decode` the key object from the JWK, and not bytes.

These fixes apply to Zentral, because the token and the JWKS come from outside:

- `PyJWKClient` does not follow a redirect to a different key source anymore (GHSA-9v7f-9g4p-ffgj).
- A header with many levels does not cause an uncaught `RecursionError` in `jwt.decode` anymore (GHSA-8wjv-2p76-3863).
- One incorrect RSA key does not stop the parsing of the full JWK Set anymore (GHSA-w6j9-cwv2-h6wq).

`PyJWKClient.fetch_data` gets the JWKS with its own opener now, to install the handler that refuses the redirects, and not with `urllib.request.urlopen`. The OIDC tests replace the opener.

The dependency metadata of 2.13.0 and 2.14.0 is the same, so no other pin moves.

### Do not use the HMAC algorithms an OpenID Connect issuer advertises

`verify_jws` made its list of permitted algorithms from `id_token_signing_alg_values_supported`, and removed only `none`. Some issuers advertise HS256 with RS256.

The key given to `jwt.decode` is the key that `PyJWKClient` reads from the issuer JWKS, a cryptography key object. A token with `alg=HS256` and a `kid` for such a key went to `HMACAlgorithm.prepare_key`, where `force_bytes` raises `TypeError`, because the key is not a string and not bytes. `TypeError` is not a `PyJWTError`, so it moved past the `except` block: the realm login callback and the OIDC API token issuer gave a 500 error, and did not report an invalid token.

An ID token signed with an HMAC algorithm uses the client secret, so a key from the JWKS cannot verify it. The HMAC algorithms are removed from the list with `none` now, and PyJWT refuses the token before it prepares the key. An issuer that advertises only HMAC algorithms leaves an empty list, and all of its tokens are invalid. Zentral could not verify those tokens with a JWKS key.

Two tests use a forged HS256 token signed with the RSA public key, against a configuration that advertises `["RS256", "HS256"]`, and against one that advertises only `["HS256"]`. With the filter removed, the two tests give the `TypeError`.

### Tests

The full suite is green (8406 tests). flake8 and ruff are clean. The added lines in `zentral/utils/oidc.py` are fully covered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -171,6 +171,8 @@ The probe caches expire after 300 seconds now. A notifier notification was the o
 
 A store worker verifies the configuration of its store every 300 seconds now. Before, only a notifier notification made the worker apply a change, and the notifier does not guarantee delivery. A worker that did not receive the notification kept the event filters and the credentials from its start, and an operator had to restart it. The worker now stops when it finds a change to its store, and starts again with the configuration in the database.
 
+Zentral does not use the HMAC algorithms that an OpenID Connect issuer advertises anymore. It verifies an ID token with a key from the issuer JWKS, and the HMAC algorithms use the client secret. An ID token signed with one of them gives an invalid token error now, and not a 500 error.
+
 Fixed the `zentral_audit` events of the Santa enrollments and the Santa enrolled machines, which were not linked to the object they report. A model that declares the parents of an object replaced the link to the object itself, and the events of that object did not reach its page. A model keeps control of its own key when it declares one, because the MDM commands are linked by UUID and not by primary key.
 
 Fixed the version of an Osquery query in the events published when a user changed the query from the web console. The version was a database expression, and not a number.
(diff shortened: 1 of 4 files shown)


soysauceconspiracy (approved):
LGTM

PR #1694, opened 2026-09-11
inventory: export only the objects the current snapshots reach

np5:
## What

The full inventory export (`POST /api/inventory/full_export/`, `export_full_inventory` command) contains the current snapshot of each machine and source, and the objects that these snapshots reference. A snapshot that a newer snapshot replaced, and an object that only such a snapshot references, are not exported anymore.

Before this change, the `machine` table had the current snapshots only, but every other table was dumped whole. The link tables carried the rows of every retained snapshot.

## How

- Each entry of the query list declares how its rows connect to the current snapshots, and the SQL is built from that: a filter on the current snapshot ids for the link tables, on the ids the current snapshots carry for the tables behind a foreign key, on the ids the current link rows carry for the tables behind a link table, and one more level for the apps, the programs and the meta business units.
- The certificates use a recursive query: the certificates of the snapshots, the signers of the app instances and of the profiles, and the chains above them.
- All the tables are read in one `REPEATABLE READ` transaction on the read-only database, so they see the same snapshots. The isolation level is only set when the export opens the transaction itself.
- The unused execute argument is removed. The queries have no placeholder, so psycopg2 never merged it.

## Tests

- A machine with two commits: only the rows of the second snapshot are exported, for every table. The certificate chain of the first snapshot is gone, and the shared root certificate stays.
- A referential closure test: every `*_id` column that points at an exported table resolves inside the export. The columns that are not foreign keys, and the one that points at a table that is not exported, are listed with a reason. A new `*_id` column fails the test until it is classified.
- The isolation level, in a `TransactionTestCase`, and the part rolling.

CHANGELOG entry under *Backward incompatibilities*. Docs updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,6 +122,11 @@ An import of a standard Osquery pack fails if one of its queries has SQL that Ze
 
 `PATCH` gives a `405 Method Not Allowed` on the Monolith catalog, condition, enrollment and manifest enrollment package endpoints. Use `PUT`, and send all the attributes of the object.
 
+#### 🧨 Inventory full export
+
+The full inventory export contains the current snapshot of each machine and source, and the objects that these snapshots reference. A snapshot that a newer snapshot replaced, and an object that only such a snapshot references, are not exported anymore. All the tables of an export are read in one transaction, so they are consistent with each other.
+
+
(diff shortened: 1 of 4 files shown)


soysauceconspiracy (approved):
LGTM

PR #1695, opened 2026-09-11
inventory: pick the tables of the full export, and describe it with a manifest

np5:
## What

- `POST /api/inventory/full_export/` takes an optional `tables` list, and the `export_full_inventory` command takes a repeatable `--table` option. The tables are exported in their canonical order, one time each. An unknown table, or an empty list, gives a `400` that lists the valid tables.
- The task result carries a manifest: the tables, with their row counts, their columns and their files, and the files of the archive, with their table, row count, size and SHA-256 digest. The archive contains the same manifest as `manifest.json`. The manifest is the contract of the multi-file downloads and of the Parquet export that come next.
- The `zentral_machine_disks_<index>.jsonl` files are named `zentral_machine_disk_<index>.jsonl`, like the files of the other link tables.

## Fix on the way

The archive was named after a timestamp with a one second resolution, and `do_full_export` returned the name it asked for, not the name the storage used. Two exports started in the same second got the same name, and the second task result pointed at the file of the first export. The archive is named after the export id now, `full_inventory_export-<timestamp>-<random suffix>.zip`, and the task result carries the name the storage used.

## How

- The export loop is split into a table iterator, which runs the selected queries in the one transaction, and a JSONL writer, which hashes and counts what it writes. The columns of a table are read from the description of the server-side cursor after the first fetch, so they are known for a table without a row.
- `FullExportSerializer` validates `tables`. The view, the task and the command pass the list through.

## Tests

- The manifest against the archive: the size, digest and row count of every file, the columns against the rows, `manifest.json` equal to the task result manifest, the file name with the export id.
- The `tables` option: canonical order and duplicates, unknown table, empty list, `null`, at the export, the API, the task and the command.
- A table without a row: no file, columns known.

CHANGELOG: a feature entry under *Inventory*, and the rename under *Backward incompatibilities*. Docs: the table list, the manifest, the retention sentence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ The documentation is published at https://www.zentral.com/docs/zentral/ now, and
 
 The inventory JMESPath compliance check publishes the same `zentral_audit` events as the other objects now, from the web console and from the API.
 
+The full inventory export can be limited to a list of tables, with the `tables` attribute of the API request or the `--table` option of the `export_full_inventory` command. The task result carries a manifest with the tables, their row counts and their columns, and the files of the archive with their size and SHA-256 digest. The archive contains the manifest as `manifest.json`.
+
 
 #### MDM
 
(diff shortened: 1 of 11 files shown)


soysauceconspiracy (approved):
LGTM

PR #1696, opened 2026-09-11
base: serve the files of a task result that has a manifest

np5:
## What

The base app can serve a task result with several files. Such a result carries a manifest with the `location` of the directory of the files in the storage, and the `files` of the directory, keyed by their name in the directory.

- `GET /api/task_result/<task_id>/`: the `download_url` attribute is present for such a result too. The manifest is echoed without the `location`.
- `GET /api/task_result/<task_id>/download/`: the manifest, with a `download_url` for each file. With an S3 or GCS storage, each file also has its `url` in the storage, and the `expires_at` time of the URL when the URL expires.
- `GET /api/task_result/<task_id>/download/?file=<key>`: one file. The key is looked up in the manifest first, then the object is resolved as `location + key`, then the response is a redirect to the signed URL, or the file itself. An unknown key, or a result without a `location`, gives a `404`. Nothing from the query string reaches the storage without the check.
- The tasks pages list the files, each with a download link.

A result with a `filepath` is served as before. A result with a manifest but without a `location` has no download, which is the case of the JSONL full export of #1695.

No task produces such a result yet. The Parquet inventory export, next in the series, is the first one. The contract is written down in the planning doc of the series.

## How

- `zentral/utils/storage.py`: `file_storage_signed_url_expiration` gives the validity of a signed URL, from the S3 `querystring_expire` seconds or the GCS `expiration`, and `None` when the storage does not sign its URLs.
- `server/base/api_views.py`: the two branches of the download view, and one file response helper shared with the single-file case.
- `server/templates/accounts/tasks/_result.html`: a table of the files on the task page, a dropdown on the tasks list.

## Tests

- The status response with a manifest, with and without a `location`.
- The manifest response: the download URLs, the signed URLs with and without an expiration time, the ownership rules.
- One file: served, redirected, unknown key with an object that exists but is not in the manifest, unknown key with a signing storage, missing object, `file` on a single-file result, other user.
- The expiration helper for the default storage, S3 with and without `querystring_auth`, S3 without an expiration, GCS.
- The two tasks pages.
- Each new test was checked against a broken version of the code, and one test was strengthened by that check: the unknown-key test passed for the wrong reason until the object existed in the storage.

Docs: the two task result sections of the core page. CHANGELOG: one entry under *Core*.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 The PBAC Schema browser documents each action now. It shows what the action authorizes, the context attributes a `when` clause can read, and the values an attribute accepts when the set is closed — for example the artifact types of `MDM::Action::"forceInstallArtifact"`. An optional attribute carries a `?` after its name, which tells you if a `has` guard is necessary. The principals and the resources show the types they are members of, so you can see what a `resource in` clause can name: a `forceCleanSync` policy can be scoped to a machine, or to one of its meta business units.
 
+A task that produces several files has a manifest in its result. The download endpoint of the task gives the manifest with a download URL for each file, and the `file` parameter downloads one file. With an S3 or GCS storage, the manifest also gives the URL of each file in the storage, and the expiration time of the URL when it expires. The tasks pages list the files.
+
 
 #### Documentation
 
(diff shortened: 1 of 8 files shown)


soysauceconspiracy (approved):
LGTM

PR #1697, opened 2026-09-11
inventory: export the full inventory as Parquet files

np5:
## What

`export_format`, `JSONL` by default or `PARQUET`, on `POST /api/inventory/full_export/`, and `--format` on the `export_full_inventory` command.

With `PARQUET`, each table is a set of [Apache Parquet](https://parquet.apache.org/) files in a directory of the storage, `exports/inventory/<export id>/<table>/<table>-<index>.parquet`, and the manifest is written last as `manifest.json`. The task result carries the manifest with the `location` of the directory, the typed columns of each table, and the files with their row count, size and SHA-256 digest. The task result endpoints of #1696 serve the files: the manifest with a download URL for each file, one file with `?file=<key>`, and the URL in the storage with its expiration time on S3 and GCS.

## How

- `zentral/utils/parquet.py`: the PostgreSQL type OID to Arrow type map, the schema of a cursor description, the record batches, and the file roller. An unknown type fails the export. The naive UTC timestamps are `timestamp[us, tz=UTC]`. `json` and `jsonb` values are the JSON text: `jsonb` reaches the cursor as text, because Django registers the identity loader, `json` as parsed values. `inet` and `uuid` values are text.
- `full_export.py`: the table iterator yields the cursor description, and the JSONL and the Parquet writers share it. Each Parquet file is written to a local temporary file through the hashing sink and saved to the storage, then the temporary file is removed: the local footprint is one file of 128 MiB at most, and no archive is built. The export fails if the storage saves a file under another name.
- A table without a row has one file without rows, so its schema is in the storage and a `*.parquet` glob on the directory matches. The JSONL format has no file for such a table, as before.

## Trade-off

The files are saved from inside the export transaction, so the `REPEATABLE READ` transaction on the read-only database lasts for the queries and for the uploads, one round trip per file of 128 MiB. The JSONL export uploads its archive after the transaction, at the price of one local copy of the whole export. A long transaction on a read replica can be cancelled by `max_standby_streaming_delay` on a streaming replica, and holds back vacuum. Something to watch on the first export of a large inventory.
- The command prints the manifest and one line per file.

## Sizes

| Setting | Value |
|---|---|
| Rows per record batch, one row group each | 65,536 |
| File size at which a new file starts | 128 MiB |
| Compression | zstd, with the page index |

## Tests

- `tests/inventory/full_export_parquet_schemas.json` is the schema contract: the test compares the schema of every table with the file, and `ZENTRAL_UPDATE_PARQUET_SCHEMAS=1` rewrites the file when a change is intended.
- The Parquet export against the storage: the size, digest and row count of every file, the columns against the file schemas, the manifest object, the current snapshots only, typed timestamps and JSON text.
- A table without a row, the file rolling, the unknown format, the file saved under another name.
- The type map, the record batches and the roller, without a database. Every mapped type that the export uses has a value in the tests, the `inet` column included.
- The API, the task and the command options.
- One end to end test: the export runs, its result becomes a task result, and the file is downloaded through the task result endpoint and read back with pyarrow.

Docs: the Parquet section of the inventory page, with the type table and reading recipes for DuckDB, Athena and Snowflake. CHANGELOG: one entry under *Inventory*.


soysauceconspiracy (approved):
LGTM

PR #1698, opened 2026-09-12
inventory: read the machine exports with an explicit fetch, on the read-only database

np5:
Two commits.

## 1. Read the export cursors with an explicit fetch

`export_machine_snapshots`, and the machine applications exports through `_export_machine_csv_zip`, set `cursor.itersize = window_size` on a chunked cursor, then iterate the cursor. The assignment does nothing: `django.db.backends.utils.CursorWrapper` has no `__setattr__`, so the value stays on the wrapper, and the psycopg2 named cursor keeps its default window of 2000 rows. The `window_size` parameter of the two functions had no effect.

The two exports share one `chunked_query` helper now, in `zentral/contrib/inventory/utils/db.py`. It opens the transaction a server-side cursor needs, runs the query, and gives the columns and the rows, `cursor.fetchmany(window_size)` rows for each round trip. The exports read 5000 rows for each round trip, the window the code always declared. The archives do not change.

The two call sites also lose their `columns = None` preamble, and the test of it for each row: the helper reads the description after the first fetch, like `iter_tables` does.

#1695 removed the same `itersize` line from the full export, because the new code fetches explicitly. The full export keeps `iter_tables`: it runs many queries in one transaction, and it needs the types of the columns for the Parquet schema.

## 2. Read the machine exports on the read-only database

The two exports are the heaviest scans of the product. They walk the current snapshots and their application tables with a wide `ORDER BY` that spills. They take no lock, and they write nothing: the archive reaches the storage after the cursor is closed. `chunked_query` runs them on the database that `get_read_only_database` names, like the full export.

Where `POSTGRES_RO_HOST` is not set, `get_read_only_database` gives the default database, and nothing changes. Where it is set, note that a long read on a replica can be cancelled when it conflicts with recovery. The three inventory exports have the same behaviour now.

## Tests

- `test_export_machine_snapshots_window_size` and `test_export_machine_csv_zip_window_size` export 5 rows with `window_size=2`, and pin the fetches to `[(2, 2), (2, 2), (2, 1), (2, 0)]`. A recording `chunked_cursor` wraps `fetchmany` on the real cursor.
- `test_chunked_query_reads_the_read_only_database` pins the alias that the helper asks `connections` for.
- Each test was checked against a broken version of the code. The `itersize` assignment records `[]`, because the psycopg2 cursor iterates below the wrapper. A hardcoded window records `[(2000, 5), (2000, 0)]`. `connections["default"]` records `["default"]`.

No CHANGELOG entry: the exports write the same files, and no caller gives a `window_size`.

### DIFF
--- a/tests/inventory/test_exports.py
+++ b/tests/inventory/test_exports.py
@@ -10,7 +10,7 @@ from pathlib import Path
 import zipfile
 import pyarrow.parquet as pq
 from django.core.files.storage import default_storage
-from django.db import connection
+from django.db import connection, connections
 from django.test import TestCase, TransactionTestCase
 from django.utils.crypto import get_random_string
 from zentral.contrib.inventory.models import MachineSnapshot, MachineSnapshotCommit, MetaBusinessUnit, Source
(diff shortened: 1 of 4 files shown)


soysauceconspiracy (approved):
LGTM

PR #1699, opened 2026-09-12
Drop the Terraform export

np5:
Drop the Terraform export from the Inventory, MDM, Monolith, Munki, Osquery and Santa apps, and the shared code in `zentral.utils.terraform`.

The "Download Terraform config" button gave a zip archive with one Terraform configuration file per group of objects, and a script with one `terraform import` command per object. The Terraform HTTP state backend in `zentral.core.terraform` is a different feature. It stays.

## Why

The archive did not describe a Zentral completely, and an apply could remove configuration.

* It covered 31 of the 57 resources of the Terraform provider. It had no resource for the MDM enrollments, declarations, apps and issuers, for the Monolith repositories, for the probes, for the stores, or for any Turbo object.
* It dropped attributes that the provider supports: the default location, the DDM distribution of the legacy profiles and the software update enforcements of an MDM blueprint, the PRK reveal rotation delay of a FileVault configuration, and the CEL expression and the custom URL of a Santa rule. Each of those attributes has a default value in the provider schema, so an apply after the import set the attribute to that default. A FileVault configuration made in the web console has a PRK reveal rotation delay of 60 minutes. An apply set it to 0, and stopped the rotation.
* The archive had no variable declaration for the API token, and none for the secret attributes. `terraform plan` stopped on an undeclared variable until you wrote one.
* The MDM archive carried the source of every configuration profile, and the download asked for the blueprint view permission only. `DownloadProfileView` asks for the artifact version view permission for one profile.
* Each view read all the objects of its app in one request, with no limit, wrote the profiles to the temporary directory of the worker, and kept the complete archive in memory.

Terraform 1.5 gives the replacement. Write an `import` block, then run `terraform plan -generate-config-out=<file>`. The provider imports each of its 57 resources, and Terraform writes the configuration from the provider schema, so the result stays correct when the provider gets a new attribute.

## Changes

* 7 commits, one per app plus the shared code.
* 39 files, 1973 deletions, 23 insertions.
* No migration, no API endpoint, no permission change. The removed views answer a browser session only.
* The index pages of Monolith, Munki and Osquery have the same header as the Santa index page now.

## Tests

Full suite with coverage: 8416 tests, no failure.


soysauceconspiracy (approved):
LGTM

PR #1702, opened 2026-09-14
pbac: answer a request with no context yet, and reach every entity a policy reads

np5:
The PBAC ground work of zentralopensource/zentral#1682, on its own, plus the cedarpy bump it asks for.

The engine change alters authorization for each module, and the Turbo work that motivated it does not change its correctness. It can be reviewed and merged before that work is ready. #1682 keeps its other two commits, and is rebased on this one.

| | Commit | Files |
|---|---|---|
| **1** | `pbac: answer a request with no context yet, and reach every entity a policy reads` | 8 |
| **2** | `Bump cedarpy to 4.12.0` | 1 |

---

# 1 — `pbac: partial evaluation, and entities that a policy can actually reach`

Three additions to the engine. No action that exists today uses any of them, which is why the four bugs below were invisible.

## Partial evaluation, so a context attribute can stay required

A `Request` carries `unknown_context`. A caller sets it when it cannot fill the context in yet — a view that decides whether to offer an action, before the user picks the object that would complete it. `Engine.authorize_request` sends those to cedarpy's `is_authorized_partial`. It holds back the part of a policy that reads the missing context, instead of failing on a context that does not have it.

The routing is on the request, so a caller that can receive either kind gets the correct evaluator without a choice to make. A policy writer never learns which one ran.

The alternative was to declare each such attribute optional, and to write `!(context has x) || …` in each policy that reads one. That fails **asymmetrically**. A `permit` without the guard is skipped, and the action disappears for the roles it grants. A `forbid` without the guard is skipped, and still decides correctly. Only the first kind is wrong, so the mistake is easy to make and hard to see.

## `NoDecision` is two answers, and only one of them is "could be permitted"

Cedar returns `NoDecision` when a policy needs the missing context. It returns the same `NoDecision` when it could not read the request at all: a bad character in an entity id, a list of entities it refuses to parse. cedarpy states this — *"When input errors occur, `decision` is `NoDecision` and `residuals` is empty."*

The two must not get the same answer. The preview grants on `Allow`, or on `NoDecision` with no errors. A rule of "anything but `Deny`" turns each input cedarpy rejects into a grant, where the full evaluator denies it. #1705 closed the one path that made this reachable, a machine entity id built from a raw serial number. But `Entity.__str__` still writes an id into Cedar source text with no escape, so an entity id that holds free text opens it again.

All three evaluators now log the Cedar diagnostics. Cedar skips a policy that errors and answers with the others. A `forbid` that reads an attribute or an entity we did not send is therefore skipped, and the request is allowed. Two of the four bugs below have that shape, and the log is what makes the next one visible.

## A context is all or nothing

Cedar accepts a context that is absent. It errors on a context that is present but missing a key, and the policy that reads it is then skipped: a `permit` wrongly denies, and a `forbid` is dropped. There is no way to say "this one key is not known yet", so `Request` refuses a request that claims one, instead of dropping the context it was given.

`{}` is not the same as absent. An empty context is a record with no attributes, and it errors in *both* evaluators. This is why a preview serializes its context as `None`. `authorize_requests` answers the previews one by one, because partial evaluation has no batch form.

## A preview answer is not a decision

`authorize_request_preview` writes to `Request.is_authorized`, which is the attribute that each enforcement path reads. Those paths have no other way to separate "could be permitted" from "is permitted", and cedarpy says a partial result must not be a final decision. So `PBACViewMixin` and `PBACPermission` refuse a request with an unknown context, and the full evaluators refuse one as well — they read the absent context as an empty record, and would deny it.

## Four latent bugs, each of which failed silently

An attribute can be typed as an entity, and an entity can be named in a context. Both were expressible before. Neither worked.

* `register_entity_type` auto-registered parents, but not the types that an attribute refers to. Cedar then answers `failed to resolve type`, which fails the **whole** schema — each policy save in the deployment, not only the action that declared it. The walk runs at the first registration: an attribute added to an `EntityType` after that is not picked up.
* `register_action` registered the types of the principals and the resources of an action, but not the ones named in its **context**. Same failure, same blast radius, through the other door.
* `_serialize_entity` copied attributes unchanged. An entity-valued attribute reached Cedar as a reference to something absent: `resource.x.y` evaluated to `entity … does not exist`, and a `forbid` **failed open**, with only a diagnostic to show for it.
* The entities sent to Cedar were taken from the principal, the action and the resource. An entity named in a **context** never reached Cedar either.

`pbac.types.iter_entity_types` is the one walk over the `TypeRef` union that registration uses. `format_type` and `cedar._attr_to_json` each already had their own, and each raises on a type it does not know. This one now does the same: a container that the walk misses is an entity type missing from the schema, which is the first bug in the list.

## What it costs

`_serialize_value` walks each entity attribute and each context on every request. Today all of them are primitives, so the walk rebuilds a dict where the code made a shallow copy before. This is a small constant cost on the authorization path, and it is what makes an entity-valued declaration enforceable. `_serialize_requests` renders a request and collects its entities in one pass, so a request is walked once, not twice.

A preview costs more than a decision. For each policy it cannot settle, `is_authorized_partial` builds and returns the part that is left, and this code reads only the decision. Measured on one request against a growing policy set: 49 µs against 118 µs at 10 policies, 243 µs against 1704 µs at 200, 554 µs against 4259 µs at 500. A page that asks one preview pays this once. A page that asks one per row pays it per row, and `Engine.authorize_requests` does not stop that — the first consumer should cache the answer on the user object, as `has_legacy_perm` already does, because a preview depends only on the principal, the action and the resource.

## What changes for a deployment today

No action declared in `main` types a context attribute as an entity, and no caller sets `unknown_context`. Each request keeps the path it takes today, and only the tests reach the preview.

Two changes are visible. A Cedar diagnostic is now logged as an error, so a deployment that already has a policy that errors during evaluation will start to say so — that is the point of the log, and the first deploy is where it surfaces. And `_serialize_value` raises on a value type that Cedar has none for, instead of passing it to cedarpy. Each context built in the tree holds `str` and `int` only, so nothing today reaches either.

---

# 2 — `Bump cedarpy to 4.12.0`

cedarpy 4.12.0 moves the Cedar Policy engine from v4.8.2 to v4.12.0, and the Cedar language from 4.4 to 4.5. The language addition is the extended `has` operator in the JSON policy format, which desugars the way the Cedar text format already did. Each policy written today still reads unchanged.

The engine breaking changes in that range are in experimental features that cedarpy does not enable — `tpe`, `protobuf` and `tolerant-ast` — and in the `pst` module, which nothing here calls. The `partial-eval` surface behind `authorize_request_preview` is unchanged.

4.12.0 also cuts the peak memory of validation of a policy against a schema with a large action set, and removes redundant per-request work under concurrent load. Each policy save here validates against the whole action set.

**This commit is separable.** `is_authorized_partial` exists on 4.8.6, and it returns the same decision as 4.12.0 for each case that commit 1 tests. Commit 1 does not need the bump, and either commit can be reverted without the other.

**The caveat this version writes down:** cedarpy documents `is_authorized_partial` as an experimental surface. It rides the `cedar-policy` crate's `partial-eval` feature, which Cedar ships outside its semver guarantee and can change in any release. The same feature backed `is_authorized_partial` on 4.8.6, so the bump does not introduce this. But commit 1 makes it the engine of a code path that Zentral has, so the next bump must be read with the partial evaluation tests in hand.

Security, from cedarpy's notes: `pyo3` 0.27.2 → 0.29.2 closes GHSA-36hh-v3qg-5jq4 and GHSA-chgr-c6px-7xpp, and `anyhow` 1.0.95 → 1.0.104 clears RUSTSEC-2026-0190. cedarpy reports none of them reachable from its API.

## Tests

The full suite passes on cedarpy 4.12.0. Each added line is covered.


sebastian-zentral (dismissed):

soysauceconspiracy (approved):
LGTM

PR #1703, opened 2026-09-14
santa: Scoped client modes

np5:
The entity attributes that this pull request uses are merged (#1702). This pull request is rebased on `main`.

A Santa configuration gives the same client mode to all its machines. This pull request adds the **scoped client mode**: an entry on a configuration that gives a different mode to some of the machines.

## What an entry is

* the scope fields of a rule: serial numbers, primary users and tags, each one with an exclusion;
* a client mode;
* its own block notification button, with an `Inherit` source that keeps the button of the configuration.

Zentral resolves one mode for each machine when it answers the preflight. There is no Santa change, no configuration profile, and no scoping in the agent. A machine gets its mode at its next full synchronization.

If more than one entry is in scope, Zentral sorts them and applies the first one:

1. the narrowest scope: serial numbers, then primary users, then tags, then no scope at all. Only the narrowest field of an entry counts;
2. `Lockdown` before `Monitor`;
3. the first name in alphabetical order.

An exception is written for one machine or for one user, and a policy for a population. This order lets the exception win.

## The cost at the preflight

The preflight is the hottest Santa endpoint, so the cost is worth stating. A configuration **with** entries pays two more queries for each preflight: one for the entries in scope, and one for the prefetch of their tags, which `precedence_rank()` reads. The filtered join carries only the tags that matched, so it cannot answer "is this entry scoped on tags at all".

A configuration **without** entries pays nothing: the sync views annotate the enrolled machine with an `Exists` subquery on a query they already make. The annotations live in `EnrolledMachine.objects.for_sync()`, which the views and the tests both call, so a test machine cannot drift from the one a view builds.

The annotations are an optimisation and not a contract. An enrollment builds the machine with `update_or_create()` and without them, and the method then asks: one query for the entries, and one for the tags of the machine. It reads the tags after the enrollment applies the tags of the enrollment secret, so the first preflight of a new machine already matches an entry scoped on one of them.

The scope match and the ranking stay in Python, so the predicate is one function that a test can call. `tests/santa/test_scoped_client_mode.py` has the first `assertNumQueries` tests of the Santa module, for the four cases.

## PBAC

Four typed actions, and no Django permission:

| Action | Resource | Groups |
|---|---|---|
| `createScopedClientMode` | `Santa::Configuration` | Admin |
| `viewScopedClientMode` | `Santa::ScopedClientMode` | Admin, User, Viewer |
| `updateScopedClientMode` | `Santa::ScopedClientMode` | Admin |
| `deleteScopedClientMode` | `Santa::ScopedClientMode` | Admin |

These are the first Santa resource types – `forceCleanSync` uses the inventory ones. The configuration is the parent of an entry, so one policy covers every entry of a configuration. `resource in [A, B]` is a parse error, so a role that manages two configurations needs one policy for each of them.

**The configuration is the boundary, and the actions carry no other attribute.** A role that can write the entries of a configuration can give any client mode to any of its machines, and the documentation says so.

An earlier revision gave the entry a `mode` attribute and the create action a `clientMode` context, to keep the Lockdown entries for another role. It is removed:

* the danger of a mode is relative to the mode of the configuration. A Monitor entry on a Lockdown configuration removes protection, and a Lockdown entry on a Monitor configuration adds it. `Santa::Configuration` carries only its name, so a policy cannot read the mode of the configuration and cannot ask the question. The gate protected the safe direction on a Monitor configuration, and permitted the dangerous one on a Lockdown configuration;
* it was not enforceable on the update, which decided on the mode that the entry had, and not on the mode that the form would give it. A role that could write Monitor entries could change one into a Lockdown entry;
* it also needed a new attribute on `PBACViewMixin`, to let the create form open on a partial decision. That commit is gone, and the shared PBAC code is untouched.

Gating on the **scope** of an entry – the serial numbers, the primary users and the tags – is the useful version of this, because it bounds the machines that the role can reach. It needs the first set-valued attributes in the schema, and it is a separate change.

The URL carries the configuration and the queryset filters on it, so `configuration` is not a form field and an entry cannot be reparented. A reparenting write is authorized against two containers, and a decision against one of them is an escalation.

## Second commit: the blocked path regex in Lockdown mode

The configuration form rejected a blocked path regex in Lockdown mode. The second commit removes the rule.

`SNTPolicyProcessor.mm` evaluates the blocked path regex, then the allowed path regex, then the client mode, and each stage returns immediately. `fileIsScopeAllowed` returns `Not a Mach-O` for a file that is not a Mach-O file, so that file is allowed in every mode, Lockdown included, and the mode fallback never sees it. In Lockdown mode the blocked path regex is therefore the only way to block such a file by path, and it also carves exceptions out of the allowed path regex. The rule removed a capability, and it did not prevent a mistake.

The rule read the client mode of the configuration, which a scoped client mode now changes for some of the machines.

The change permits what was rejected before, so no existing configuration breaks. It is a feature, not a backward incompatibility.

## Not in this pull request

The REST API, and therefore goztl and the Terraform provider. The scoped path regexes. The `requires_approval` flag of the approval engine. The preview of the mode of one machine in the web console, which is the thing an operator still cannot see.

## One edit to a merged migration

`0047_event_detail_source` carries the help text of the event detail source that #1700 changed after the last `makemigrations --check`. The model and the migration did not agree on `main`. The migration was never deployed, so the fix goes in the file instead of a new migration.

## Tests

`tests/santa/test_scoped_client_mode.py` (scope, resolution, preflight, event detail, query counts, serialization), `tests/santa/test_scoped_client_mode_pbac.py` (the console, with raw Cedar policies), and the form tests in `tests/santa/test_forms.py`. 100% patch coverage, verified before the push.

The tests build a real enrolled machine, and read it back through `for_sync()` when they want the annotations. Seven of them post to `santa_public:preflight`, on both paths: the machine the view reads with the annotations, and the machine an enrollment builds without them.


sebastian-zentral (dismissed):

soysauceconspiracy (approved):
LGTM

PR #1704, opened 2026-09-14
mdm: Apply the DEP enrollment maximum OS version to the required update

np5:
`verify_os_version` read the maximum required OS version of the DEP enrollment only to decide to look for an update. It did not pass the maximum to the lookup, so Zentral asked the device to install the most recent update available for it, and not the most recent update below the maximum.

An enrollment with a macOS maximum of `27` asked a device to install macOS 27, which the enrollment page shows as excluded (`min ≤ OS version < max`). The declarative path already passes the maximum.


sebastian-zentral (approved):

soysauceconspiracy (approved):
LGTM

PR #1705, opened 2026-09-15
inventory: use the url safe serial number as the machine entity id

np5:
A serial number comes from a device, and Zentral stores it as it arrives. The machine resource used it as the Cedar entity id, and that id reaches Cedar as source text — `Inventory::Machine::"<serial>"`. A serial that holds a quote, a backslash, a control character or an invisible space made a reference Cedar refused to parse.

The result was a denial before Cedar read a single policy:

* The machine could not be tagged, and nothing said why. The same applies to a forced install of an artifact and to a forced clean sync, which also take the machine resource.
* A policy that named such a machine could not be saved, because the same text failed validation.

This is live in a release: `zentral/contrib/inventory/pbac.py` has not changed since `v2026.5.7`.

## The change

The resource uses `get_urlsafe_serial_number()` now. That is the serial itself when the serial needs no encoding — which covers every Apple serial number — and a leading dot plus url safe base64 otherwise.

Two serials cannot share an id. A serial that needs encoding always starts with a dot, and a serial that already starts with one is encoded too.

The form is already the machine URL, and every link to a machine page uses it, so the id a policy names is the one in the address bar.

## What it costs

A policy that names a machine whose serial number needs encoding has to use the new id, which the URL of the machine page shows. A policy that names a machine with a plain serial number does not change.

The machine resource is the only entity id built from text a device sends. Every other one is a primary key or a literal in the code.

## Why this form

The measurements below cover 147 serials: all of ASCII, the invisible space characters, accents, combining marks, emoji, CJK, and the edges — empty, `.`, `~`, 4000 characters.

| | raw serial (today) | escaped in the policy text | url safe serial |
|---|---|---|---|
| quote, backslash, control characters | denied | ok | ok |
| NBSP, zero width space, ideographic space | denied | denied | ok |
| accent, combining mark, emoji, CJK | ok | broken | ok |
| collisions | — | — | none |

Escaping the serial in the source text is not enough. Cedar refuses the invisible space characters whatever the escape, and escaping every non ASCII character breaks the accents and the emoji that work today.

The url safe form is also the only one that works for a partial evaluation. cedarpy accepts a structured `{"type": ..., "id": ...}` for `is_authorized` and `is_authorized_batch`, which would fix those two, but `is_authorized_partial` takes source text only.

## Tests

The full suite passes: 8451 tests. Reverting the one line fails four of the new tests, two of them because the policy itself cannot be saved.


soysauceconspiracy (approved):
LGTM

PR #1707, opened 2026-09-16
santa: Scoped path regexes

np5:
The entity attributes and the abstract base that this pull request uses are merged (#1703). This pull request is rebased on `main`.

A Santa configuration has one allow path regex and one block path regex for all its machines. This pull request adds the **scoped path regex**: an entry on a configuration that adds a pattern for some of the machines. It is PR 4 of the scoped configuration items series, and it reuses the abstract base of #1703 without changing it.

## What an entry is

* the scope fields of a rule: serial numbers, primary users and tags, each one with an exclusion;
* a policy, `Allow` or `Block`;
* an ICU pattern.

Zentral composes one pattern for each policy when it answers the preflight, because Santa accepts only one. There is no precedence: every entry in scope is included, and the pattern of the configuration is a part that always matches, so it keeps applying.

```
^(?:(?:the configuration)|(?:the first entry)|(?:the second entry))
```

## Three properties of the composition

**Anchored once, around the whole alternation.** Santa prepends `^` to a pattern that does not have one, so `^(?:a)|(?:b)` would leave `b` unanchored. Each part loses its own `^` and is wrapped.

**Ordered by name.** Santa removes every decision cache when the pattern changes, and the entries query has no order of its own — Postgres returns the rows in whatever order the plan gives, which changes when a row is updated. An unordered read would send a different string on each preflight and flush the caches of the whole fleet, which is the bug PR #1606 fixed by another route.

**A part must not match an empty path.** `(?:)`, `|`, `.*` and `(?:abc)?` all do, and one of them in the union makes the union match every path, because an empty match is a match at position 0. The validator refuses a pattern whose compiled form matches `""`, which also covers the `^`-only spelling, and the composition drops a part that strips to nothing.

## Validation

The path regexes have no server side validation today: never compiled, never checked, in the form, the serializer or the model. Composing them makes that untenable, because one bad fragment takes the whole union with it.

Each pattern is compiled when it is saved. A **capture group** is refused, because each entry is wrapped in the alternation, where a capture group renumbers the ones after it and a repeated named group makes ICU fail — `re.compile(p).groups == 0` refuses both in one check. An **inline flag group without a scope** is refused with its own message: `(?i)abc` compiles alone but not when it is wrapped, and ICU accepts it in both positions, so `re.compile()` cannot be the one that reports it. Write `(?i:abc)`.

The composed pattern is compiled again before it is sent, and **the fallback is the pattern of the configuration, not a pattern that cannot match.** Nothing validated the path regexes of a configuration until now, and ICU accepts syntax that Python does not, so a configuration can hold a pattern that Santa enforces today and Zentral cannot compose — `(?i)/downloads/` is the realistic one, because it compiles alone and fails only once it is wrapped. Sending a pattern that cannot match would end a block that works today, and the sync state has priority over the configuration profile, so nothing would take over. The configuration keeps exactly what it enforces, without the scoped entries, and Zentral logs the error.

A path regex of a configuration is validated when it changes now, so a new pattern is refused at the form and in the API. A pattern stored before this change does not block an edit of another field, because it is still enforced.

Python and ICU do not agree at the margins. Python refuses `\p{…}`, `\X`, `\R` and `\Q…\E`; `[[:alpha:]]` is worse, because Python accepts it as a set in a set, which is not the POSIX class ICU applies. This is a sanity check and not a guarantee, and the documentation says so.

## The Santa agent

Verified against `northpolesec/santa` at `8b0cfe9c`, 72 commits after the last time these facts were read.

`b1f00013` (2026-09-02) sends the sync patterns through `expressionForPattern:`, so the agent anchors them itself and the difference between the session that receives a pattern and the sessions after a restart is closed. **It is in no release**: the newest tag is `2026.7`, from 2026-08-04. So the anchoring here is a fix for every agent in the field and a no-op once that ships.

The evaluation order is unchanged: the block regex, then the allow regex, then the client mode. `fileIsScopeAllowed` returns `Not a Mach-O` after it reads the allow pattern, so a block entry is still the only way to block a script by its path, in every mode.

## Backward incompatibility

The patterns of a configuration are composed too, so a configuration with no entry also gets `^(?:(?:its pattern))`. The enforcement after a restart of the agent does not change, because the agent already anchored it there, but the pattern was a search in the session that received it and it is a prefix in every session now. **A pattern that must match a path segment in the middle needs a leading `.*`.** The machines remove their decision caches one time, when they receive the new pattern.

## The first commit: one query instead of two

`precedence_rank()` only asks whether an entry is scoped on tags at all, and the join in `for_machine()` carries the tags that matched, not all of them — so the answer came from a prefetch, a second query for one boolean. `with_tag_presence()` annotates it with an `Exists` in the SELECT that already reads the entries.

A scoped path regex never ranks, so it does not read the tags at all. The result is that a configuration using both pays the same two queries per preflight that a configuration using scoped client modes pays today: the optimisation pays for the feature.

| configuration | before | after |
|---|---|---|
| no entry | +0 | +0 |
| scoped client modes | +2 | **+1** |
| both | – | **+2** |

Measured on the endpoint: a steady state preflight is 25 queries and an enrollment preflight is 63.

## PBAC

Four typed actions, and no Django permission: `createScopedPathRegex` on `Santa::Configuration`, and `view`/`update`/`deleteScopedPathRegex` on `Santa::ScopedPathRegex`, which has the configuration as its parent. Writes are in `AdminActions`, the view action is in `AdminActions`, `UserActions` and `ViewerActions`.

**The actions carry no attribute.** The design document gives the create action a `policy` context, from the revision that gave `ScopedClientMode` a `mode`. It was removed in #1703 for a reason that holds here without changes: the gate is not enforceable on the update, because `policy` is a form field, so a role that could write Block entries could make one an Allow entry. The shared PBAC code is untouched.

## Not in this pull request

The REST API, and therefore goztl and the Terraform provider — PR 3 covers both scoped models in one pass. The per machine preview of the composed patterns. A warning when an allow entry is shadowed by a block entry: regex containment is undecidable, and no customer has asked for it.

## Tests

`tests/santa/test_scoped_path_regex.py` (composition, the order between two preflights, scope, preflight, queries, serialization), `tests/santa/test_scoped_path_regex_pbac.py` (the console, with raw Cedar policies), and the form tests in `tests/santa/test_forms.py`.

The preflight tests go through `santa_public:preflight` on **both** paths: the machine the view reads with its annotations, and the machine an enrollment builds without them. That second path is what 500'd in #1703, where every test used a stand-in.

100% patch coverage, verified before the push by intersecting the missing lines of `coverage json` with the hunks of `git diff -U0`.


docs/content/apps/santa.md
```diff
@@ -234,6 +234,90 @@ permit (
...
+
+Three properties of the result:
+
+* **It is anchored at the start of the path.** Santa adds a `^` to a pattern that does not have one, so an unanchored alternation makes only the first alternative an anchored one. Use a leading `.*` to match a path segment in the middle, for example `.*/Downloads/`.
+* **The entries are in the alphabetical order of their names.** Santa removes all its decision caches when the pattern changes, so the order must be the same between two preflights.
+* **If no pattern applies, Zentral sends a pattern that cannot match.** Santa keeps the pattern it has if the key is not in the response, and it uses the pattern of the configuration profile if the key is empty.
+
+### Accepted patterns
+
+Santa uses the ICU regular expression syntax. Zentral compiles each pattern with Python, which is not the same syntax at the margins. Python does not accept `\p{…}`, `\X`, `\R` and `\Q…\E`. `[[:alpha:]]` is more dangerous: Python accepts it as a set in a set, which is not the POSIX class that ICU applies.
```
sebastian-zentral, line 265: What is this? An explaination? Would it not be better to have some examples? Or a clear reference?

sebastian-zentral (approved):

PR #1708, opened 2026-09-16
santa: REST API for the scoped client modes and the scoped path regexes

np5:
The scoped client modes (#1703) and the scoped path regexes (#1707) are merged, and the console is the only way to manage them. This pull request adds the REST API for both. It adds no PBAC action, no model change and no migration.

```
/api/santa/scoped_client_modes/            GET (list), POST
/api/santa/scoped_client_modes/<int:pk>/   GET, PUT, DELETE
/api/santa/scoped_path_regexes/            GET (list), POST
/api/santa/scoped_path_regexes/<int:pk>/   GET, PUT, DELETE
```

It carries two other things, in their own commits. Both came out of the review of the API, and both touch code outside it. Say the word and I split them out.

## The list endpoints take a required configuration

The view action takes the **entry** as its resource, so a correct list is one decision for each row. The comment above `viewEnrolledMachine` says why that is usually refused: a page cut before the decisions drops an entry the caller is allowed to see, and compiling a Cedar residual into a queryset filter is its own piece of work.

That reasoning is about the fleet. The entries of one configuration are another scale, and the endpoint carries that bound: `configuration_id` is required, and the decisions are taken before the page is cut, so `count` and `next` describe what the caller can see.

**Nothing caps the entries of a configuration.** No model constraint, no validator, no form limit. What keeps the set small is the way the entries are used: a temporary exception updates an entry instead of adding one. Measured on the endpoint with `?limit=1`, which returns one row whatever the count: 3.8 ms at 10 entries, 8.7 ms at 100, 66 ms at 1000, 416 ms at 5000. Cedar alone accounts for 139 ms of that last figure, so an ORM rewrite halves the constant and leaves the slope. Taking the decisions in SQL needs a partial evaluation residual compiled into a queryset filter, and a type aware partial evaluation inside Cedar's semver to build it on; cedarpy 4.12.0 exposes only `is_authorized_partial`, which is outside that guarantee. The comment records this.

**An entry the caller cannot see is absent from the results, and not a 403.** An entry that is refused and an entry that does not exist read the same way. A 403 would say that the configuration has entries.

The **detail** endpoints do the opposite, and cannot do otherwise: the entry is the resource, so `get_object_or_404` runs before the decision and a primary key that is not stored is a 404 while a stored one the caller may not see is a 403. A caller with a token learns which primary keys exist, and nothing else about them. That is what every PBAC detail endpoint of the project does, and the console with it. Two tests state it now, because neither answer had one.

`configuration_id` is declared with a `ModelChoiceFilter` on the `DjangoFilterBackend` the base class already wires, like `RuleFilter` in the same file. An ID that names no configuration is a 400, where the first revision of this branch answered 200 with an empty page.

## A create reads the configuration, then decides, then validates the rest

The configuration the body names is the **resource** of the decision, so it has to be a real one before the decision can be taken. A body with no configuration, or with one that does not exist, is a 400 whether the caller was authorized or not. `forceCleanSync` does the same with its sync type.

Everything else is validated after the decision. A caller the engine refuses gets a 403 and learns nothing about the names a configuration already holds.

## An update carries what a check reads with something else

The three scope pairs are required on an update, and the event detail of a client mode with them. The item validator declares them, next to the checks that read them, and the update serializer requires whatever it declares — so the two cannot drift when a scope dimension is added. A field no check reads with another one, `description` and `event_detail_text`, keeps its stored value when the body leaves it out, like every other `fields = "__all__"` serializer of the project. `configuration`, `name` and the client mode or the pattern are required anyway, because the model gives them no default.

The first revision of this branch merged the body over the stored entry instead. That fed stored values to the validators, and they have no skip-if-unchanged guard, unlike `ConfigurationValidator`: an entry stored with an `event_detail_source` of `VOTING_PORTAL` could not be edited at all once the voting realm of its configuration was cleared, because a `PUT` that only changed the name answered 400 on an attribute the caller never sent. Requiring those attributes removes the merge, and a caller escapes a stored value that stopped being valid by naming another one.

The cost is a longer body: a `PUT` carries eleven attributes where a create needs three, so a client that wants to rename an entry reads it first. That is free to decide now, because goztl does not cover these models yet and a Terraform provider sends full state on every apply.

**The configuration is compared, not dropped.** A different one in the body is a 400, so an entry cannot change configuration — delete it and create it again. The comparison is on resolved objects, the way `JobScopeSerializerMixin` does it in turbo, so a primary key written `1.0` or `"01"` resolves instead of failing a string comparison.

`PATCH` is a 405.

## Two changes outside the API

**`santa: Let Django validate the name of a scoped configuration item`.** The configuration was excluded from the form, so Django dropped the unique check of `(configuration, name)` and the item validator carried it by hand — which then forced the serializers to empty `Meta.validators`, so that the API would answer on the `name` field like the form rather than on `non_field_errors` like DRF. Two mechanisms to compensate for one excluded field.

The configuration is a **disabled hidden field** of the form now. A disabled field keeps its initial value whatever the body carries, so an entry still cannot be reparented, and Django validates the unique together itself. The check by hand goes, `Meta.validators` goes with it, and DRF builds the validator of a multi field key the way it builds every other one — so a constraint added later is checked instead of silently dropped.

State the cost: **a duplicate name answers on `non_field_errors` over the API and on `__all__` on the form**, so the console shows it at the top of the form and not under the `name` input. The models that answer on the `name` field elsewhere in the API declare the name unique on its own, which is a different check.

One trap this opened: a Django template resolves a key before an attribute, so `form.configuration` became the bound field once the field joined the form. The cancel link of the two form templates failed loudly and the tests caught it. The breadcrumb failed **silently** — an empty `href` with a hidden input as its link text — because nothing in the app asserts rendered template content. Both read `view.configuration` now, verified by rendering the create form.

**`santa: Refuse a PATCH on the rules endpoint`.** Zentral does full updates only, and the audited base classes drop `PATCH` from `http_method_names`. A URL configuration walk over every contrib app found one endpoint that still answered one: `/api/santa/rules/<pk>/`, a plain `generics.RetrieveUpdateDestroyAPIView`. It was not a partial update either — `RuleSerializer.validate()` reads `data["target"]`, DRF skips a field a partial body leaves out, and the `KeyError` was a **500**.

It gives a 405 now, under a backward incompatibility in the CHANGELOG. A test walks the URL configuration and asserts that no API view has `patch` in `http_method_names`, so the restriction no longer depends on the base class a view inherits. I verified the guard fails by re-enabling `PATCH` and watching it name the offending class.

## Not in this pull request

goztl, and therefore the Terraform provider — two repositories, in that order, because goztl has to be released before the provider can depend on it. The per machine preview of the composed patterns. The `requires_approval` flag of the approval engine.

## Tests

`tests/santa/test_api_scoped_items.py`: 49 cases over the two models, 90 tests, a mixin for what they share and one case each for the client mode and the pattern. Every test drives a **service account with a raw Cedar policy** and never a superuser — a superuser is authorized by PBAC and by the Django model permissions alike, so a suite that used one would pass on an endpoint that no other principal can reach.

Three things the tests needed to be told properly:

* **Postgres does not order the names the way Python sorts them.** The database collates `en_US.utf8`, which weighs the letter before the case, so a Python sort of the default random names disagrees with it in 38 runs out of 40. The paging test builds lowercase names, which agree in 40 out of 40 in that collation and under `C` alike, and sorts them itself. Comparing the pages to the endpoint's own unpaged answer, as the first revision did, passes with the `order_by` deleted.
* **The audit event of an update carries the tags before and after**, because `prev_value` is serialized before the write and the new value after it. The update test changes the tags to hold that.
* **A refusal does not need a state assertion.** `ATOMIC_REQUESTS` is on and DRF's exception handler calls `set_rollback()`, so a 4xx rolls the request back. Injecting a write before a `PermissionDenied` in `perform_create` gives 403 with no row.

Verified before the push: the full suite the way CI runs it (8717 tests), each commit of the branch green on its own, `makemigrations --check` with no change, flake8 and ruff clean on every file this pull request touches, `hugo --panicOnWarning` clean, and 100% patch coverage from the missing lines of `coverage json` against the hunks of `git diff -U0`.


soysauceconspiracy (approved):
LGTM

PR #1709, opened 2026-09-16
santa: Remove the ambiguity from an update body

np5:
## The gap

Almost every attribute of a Santa configuration and of a Santa rule has a model default or is blank, so DRF makes it optional and a `PUT` that leaves one out keeps the stored value. `ConfigurationSerializer.validate()` and `RuleSerializer.validate()` only read the body, so they never saw what the write produced. Four requests were accepted:

* `PUT {"event_detail_url": ""}` on a configuration whose stored source is `CUSTOM` wrote a `CUSTOM` configuration with no URL.
* `PUT {"voting_realm": null}` on a `VOTING_PORTAL` configuration wrote one with no portal.
* A rule body that named an exclusion, and said nothing about the inclusion, wrote a rule that includes and excludes the same serial number, primary user or tag.
* A rule body that changed the policy wrote a rule that keeps a `custom_msg`, a `custom_url` or a `cel_expr` that the new policy does not accept.

The web console cannot reach any of them: the fields of `ConfigurationForm` and of `UpdateRuleForm` carry every attribute into `cleaned_data`.

The cause is one hole: in a `fields = "__all__"` serializer, an attribute that is not in the body is absent from `validated_data`, so "not sent" and "set to empty" look the same to a check and different to the write. This pull request removes the ambiguity from the bodies that are accepted. It does not rebuild the body from the instance.

## The rule clears what the body leaves out

The nine attributes the checks read carry a default now: `cel_expr`, `custom_msg` and `custom_url` are `""`, and the six scope attributes are `[]`. The body always answers for them, so a `PUT` that leaves one out clears it. That is what a full update means, and it is what the console already does.

`validate()` is unchanged. It reads the body, as before, and the body is now complete. The whole change is an `extra_kwargs` block.

`description` keeps its stored value. No check reads it with another attribute, so a body that does not name it is not ambiguous.

The rules written the old way were dead, and not dangerous: a scope that includes and excludes the same value never matches, and a policy that does not block has no notification to carry the message. But they were silently dead.

## The configuration writes the whole button, or none of it

One of `event_detail_source`, `event_detail_url` and `event_detail_text` on its own cannot say what the block notification button has to be. An update takes the three together, or takes none of them, and names the ones that are missing. A create is not concerned: there is no stored value to read the body against, and an attribute the body leaves out is the model default.

The source then decides the rest:

| source | URL | label |
|---|---|---|
| `CUSTOM` | required | kept |
| `VOTING_PORTAL` | cleared, the portal computes its own | kept |
| `LOCAL`, `NONE` | cleared | cleared |

`VOTING_PORTAL` also needs a voting realm with the user portal. The label is kept for `VOTING_PORTAL` because `get_event_detail()` reads it there, with `DEFAULT_EVENT_DETAIL_TEXT` as its fallback — which is also why the label is not required for `CUSTOM`.

**The voting realm stays out of the group.** It is an attribute of the configuration and not of the button, and the console manages it on its own. State the cost: the pair is then read across the body and the instance. Whichever side the body names, the other one comes from the stored configuration. `PUT {"event_detail_source": "VOTING_PORTAL", …}` reads the stored realm, and `PUT {"voting_realm": null}` reads the stored source. A body that names neither side makes no claim about the button, and nothing is checked.

The alternative is to put the realm in the group, which removes the two instance reads and makes every request that manages the button send the realm as well. That is a bet on what a client sends. Keeping the realm out holds for a client that sends none of these attributes, one that sends the three, and one that sends the three and the realm.

## The console

The form carries every attribute, so the group rule never fires there. It clears the URL and the label the same way, which it did not do before: the console could leave a URL behind on a configuration that had moved to `LOCAL`. One test states this.

## Tests

Twenty-one tests, over the serializers, the endpoints and the configuration form.

`APIViewsTestCase.force_rule()` built a rule the API refuses, `ALLOWLIST` with a custom message, so thirteen tests were passing on a state that no validated path produces. It creates the rule without the message now, and three assertions change: two lose that key, and `test_update_rule` states that the excluded tags its body does not name are cleared.

Verified before the push: the full suite the way CI runs it (8740 tests), each commit green on its own, flake8 and ruff clean on every file this pull request touches, `hugo --panicOnWarning` clean, and 100% patch coverage from the missing lines of `coverage json` against the hunks of `git diff -U0`.

## Not in this pull request

goztl and the Terraform provider. The provider has no attribute for the button today, so it sends none of the three and this pull request does not change what its `apply` does. When it manages them, it sends the three together, which is what the group rule asks for.


soysauceconspiracy (approved):
LGTM

PR #1718, opened 2026-09-19
santa: The first scope field that matches a machine decides

np5:
The scope of a Santa rule is decided by the first field that matches the machine. The scoped client modes and the scoped path regexes follow the same rule, and two path regex entries with the same pattern are two statements about one path, with one winner per machine.

This is the first pull request of three. The next one allows several rules per target with a winner per machine, and the last one adds a page per machine that explains every decision.

## The model

A rule has three scope fields, each one with an exclusion. A field **matches** a machine when it contains its serial number, its primary user, or one of its tags. Zentral looks at the fields from the narrowest to the widest, and the first field that matches the machine decides:

1. Serial numbers. In `serial_numbers`, the rule is in scope. In `excluded_serial_numbers`, it is not.
2. Primary users, the same way.
3. Tags. A machine can carry a tag of each field: the exclusion wins.
4. If no field matches the machine, the rule is in scope only when it has no scope field. Such a rule is for every machine that its exclusions do not match.

The level that decided is the **match rank** of the rule for the machine: serial number, primary user, tag, everyone. This pull request computes it, and the next one uses it to pick a winner between several rules for one target. An exception is written for one machine or one person, and a rule for a population, so the rank lets the exception win.

| Rule | Machine | Decided by | In scope |
| --- | --- | --- | --- |
| serial numbers `S1`, tags `vip` | `S1`, no tag | the serial number | yes |
| serial numbers `S1`, tags `vip` | `S2`, tag `vip` | the tag | yes |
| serial numbers `S1`, tags `vip` | `S3`, no tag | nothing, the rule has scope fields | no |
| tags `fleet`, excluded serial numbers `S1` | `S1`, tag `fleet` | the excluded serial number | no |
| serial numbers `S1`, excluded tags `contractors` | `S1`, tag `contractors` | the serial number, the tags are not looked at | yes |
| tags `vip`, excluded tags `contractors` | `S6`, tags `vip` and `contractors` | the tags, the exclusion wins | no |
| no scope field, excluded tags `contractors` | `S4`, tag `vip` | nothing, the rule has no scope field | yes |
| primary users `alice` | `S5`, no primary user | nothing, the rule has scope fields | no |
| tags `fleet`, excluded primary users `bob` | `S5`, tag `fleet`, no primary user | the tag | yes |

## What changes on a fleet

Before, every scope field that was set had to match, and an exclusion in any field removed the rule. The documentation has said "the rule is in scope if any serial number, primary user or tag is a match" since 2021: the AND came with the commit that added the exclusions, 07fa1bbb, and nothing documented it. Three kinds of rules change their reach:

* A rule with two scope fields set, for example serial numbers and tags, reaches the union of the two lists now, not the machines in both.
* A rule with a scope field and a wider exclusion, for example serial numbers and excluded tags, keeps the listed machines now, whatever their tags.
* A rule with `excluded_primary_users` reaches the machines that report no primary user now. Before, such a machine got no rule with that field set.

The scoped client modes and path regexes shipped in the unreleased 2026.6, so their change costs nothing. For the rules, this query lists the ones to review on a tenant before the deploy, and the same query with `santa_scopedclientmode` and `santa_scopedpathregex` lists the entries:

```sql
select id from santa_rule
where (cardinality(serial_numbers) > 0)::int
    + (cardinality(excluded_serial_numbers) > 0)::int
    + (cardinality(primary_users) > 0)::int
    + (cardinality(excluded_primary_users) > 0)::int
    + exists (select 1 from santa_rule_tags where rule_id = santa_rule.id)::int
    + exists (select 1 from santa_rule_excluded_tags where rule_id = santa_rule.id)::int >= 2;
```

A rule with two fields at the same level, tags and excluded tags, behaves as before.

## The scoped items

* The scoped client mode resolution ranks an entry by the level that decided for the machine, not by the narrowest field of the entry. An entry with serial numbers and tags was a serial number entry for every machine it reached. It is one for a listed machine, and a tag entry for a machine it reaches by a tag. Lockdown, then the name, break the ties as before.
* Two path regex entries with the same pattern are two statements about one path, like two rules for one binary. For each pattern, one entry wins for the machine: the lowest match rank, then Block before Allow. Before, a pattern under both policies reached the client in its two regexes, and the client checks the block one first, so an Allow entry written as an exception to a Block entry was dead. A leading `^` does not make another pattern, the composition removes it. The composed pattern keeps the alphabetical order of the names.

## Implementation

* The rule download query computes the match rank in one `case` expression, whose `when` order is the model. A rule is in scope when the rank is not null. A null primary user parameter and an empty tag array make their clauses false, so a machine without a primary user or without tags needs no other statement. The rank is returned by the query for the server and popped before the rule reaches the client: the wire format does not change.
* `for_machine()` builds the same `case` as a `Case`/`When` annotation, with `Exists` subqueries for the tag fields, because an OR across the many-to-many join duplicates the rows. `precedence_rank()` and `with_tag_presence()` are gone.
* `resolve_scoped_path_regexes()` picks the winner per pattern before the composition.

## Tests

* `tests/santa/test_scope_resolution.py` is the scope table above as data, run against `_iter_new_rules()` and against `for_machine()` by the same module. The two implementations cannot drift without a row failing on one of them.
* The rule engine tests carry the match rank, and the three that pinned the old behaviour now pin the new one: a machine without a primary user gets a rule with `excluded_primary_users`, a listed serial number survives a wider exclusion.
* The scoped client mode and path regex tests cover the level that decided, the winner per pattern, Block over Allow at equal match, the `^` normalisation, and the same pattern under one policy composed once. The preflight query counts are unchanged.

## Docs

The Rule scope section describes the four rules and the table. The scoped client modes and the scoped path regexes point at it, and their resolution sections carry the level that decided and the winner per pattern. `docker compose -f docker-compose.docs.yml run --rm docs hugo --panicOnWarning` passes.


soysauceconspiracy (approved):
LGTM

PR #1719, opened 2026-09-19
Bump anyio to 4.14.2

np5:
Bumps `anyio` from 4.6.2.post1 to 4.14.2, which closes the two open Dependabot alerts.

### The advisories

**GHSA-82r6-8w77-94w6, critical.** `TLSStream.wrap()` encoded an internationalized host name with IDNA 2003, through the standard library, instead of IDNA 2008, so the host name could be matched against the wrong certificate. 4.14.2 resolves the name with `idna2008_resolve()` before it hands it to `wrap_bio()`.

**GHSA-5p39-cfhj-2xmp, medium.** `to_process.run_sync()` blocked for ever when the worker function wrote enough data to fill the undrained stderr pipe. The worker now sends `sys.stderr` to `os.devnull`, which is what the documentation already said.

### What this reaches here

`anyio` is not imported in this repository. It arrives through `httpx`, and only on the async path: httpcore's anyio backend calls `anyio.streams.tls.TLSStream.wrap()` with the server host name.

- The APNs client in `zentral/contrib/mdm/apns.py` uses `httpx.Client`, which is synchronous and goes through the stdlib `ssl` module. The TLS advisory does not touch it.
- The async path is the Intune client in `ee/zentral/contrib/intune/api_client.py`, through `httpx.AsyncClient` and the Microsoft Graph SDK. Those host names are ASCII, and IDNA 2003 and IDNA 2008 agree on ASCII, so the wrong-certificate match is not reachable today.
- The process pool advisory is not reachable at all: nothing here calls `anyio.to_process`.

### Dependency metadata

No other pin moves. anyio 4.12.0 dropped `sniffio` as a direct dependency, but `sniffio` stays pinned because `httpx` 0.27.2 requires it. 4.14.2 asks for `typing_extensions` only below Python 3.13, and for `idna>=2.8`, which `idna==3.15` satisfies. 4.13.0 dropped Python 3.9; Zentral runs 3.14.

### Why not the most recent release

4.15.1 requires `typing_extensions>=4.16.0` below Python 3.15, so on Python 3.14 it would move a second pin, and 4.15.0 rebuilt the `anyio` and `anyio.abc` modules to import their submodules lazily, which 4.15.1 had to patch again. 4.14.2 is the version the two advisories name.

### Verified locally

The image builds, `pip check` reports no broken requirements, and the installed set is the constraints file with no difference. The test suite was not run locally; CI covers it.


soysauceconspiracy (approved):
LGTM

PR #1720, opened 2026-09-19
santa: One rule per target and policy, and a winner per machine

np5:
Second pull request of three, after #1718. A target can have several rules in a configuration, and each machine gets one of them.

## The winner per target

A target has one rule per policy at most. Each rule is a statement about a population. For one machine, the rule download keeps the first rule of the target in this order:

1. The lowest match rank, from #1718: serial number, primary user, tag, everyone.
2. The strictest policy.

| Order | Policy | Position |
| --- | --- | --- |
| 1 | `BLOCKLIST` | The strictest. |
| 2 | `SILENT_BLOCKLIST` | The same decision, without the notification. The loud block first, so a machine under both gets the message. |
| 3 | `CEL` | Decides at run time, and can block: below an unconditional block, above an unconditional allow. |
| 4 | `ALLOWLIST` | |
| 5 | `ALLOWLIST_COMPILER` | Allows the target and what the process writes: the most permissive. |

The `(configuration, target, policy)` constraint is what makes the order total: the candidates for one target all have different policies, so the second key never ties and no third key is needed.

Four levels of intent in two rules, for one Team ID: block the fleet, allow `vip`, but not the `vip` machines that are `contractors`, but yes for `S1`, a vip contractor.

| Rule | Policy | Fields |
| --- | --- | --- |
| A | `BLOCKLIST` | none, every machine |
| B | `ALLOWLIST` | tags `vip`, serial numbers `S1`, excluded tags `contractors` |

| Machine | B | Winner | Result |
| --- | --- | --- | --- |
| tag `fleet` | not matched | A | blocked |
| tag `vip` | matched by the tag | B, the narrower match | runs |
| tags `vip` and `contractors` | out at the tag level | A | blocked |
| `S1`, tags `vip` and `contractors` | matched by the serial number | B, the narrower match | runs |

This is the case #1718 opened: in Lockdown, an exclusion alone leaves the machine without a rule, so the exception has to be an allow rule for the population, on the same target.

## Voting rules

A target with a voting rule takes no other rule, and a target with a rule takes no vote. The second half exists: the ballot box refuses the vote. The first half is new in the form and the API. A partial unique index keeps one voting rule per target, the voting upsert conflicts on it, and it skips a target that has a non-voting rule, so a stale voting rule cannot survive a rule added later. Without the exclusion, a voting rule scoped to the voters' primary users would win by rank over an admin block scoped to a tag.

## The scoped items

The same shape of constraint for the two kinds of entries. It makes their order total too, and the name leaves the sort keys:

| Model | Constraint | Two candidates for a machine always differ on |
| --- | --- | --- |
| Scoped client mode | `(configuration, client_mode)` | the mode, so Lockdown over Monitor decides |
| Scoped path regex | `(configuration, regex, policy)` | the policy, so Block over Allow decides |

A leading `^` is removed from a path regex when the entry is saved, because the composition removes it: `^/a/` and `/a/` are one pattern. The migration removes it from the stored rows before it adds the constraint. A bare `^` stays, and is refused as before.

The cost of the mode constraint is one block notification button per mode per configuration. Kiosks and employees with different buttons in Lockdown are two configurations, which they usually are already.

## Rulesets

A ruleset rule is keyed by its target and its policy. A policy change in the ruleset creates a rule and deletes the old one, with a `created` and a `deleted` event instead of an `updated` one. A conflict with a rule outside the ruleset is per policy, and a voting rule on the target is a conflict.

## On a fleet

The rule constraints hold for the existing rows: there was one rule per target. The migration refuses a configuration with two scoped client modes of the same mode, or two scoped path regexes with the same pattern and policy. It also limits a path regex to 512 characters, and refuses a longer stored pattern. Both kinds shipped in the unreleased 2026.6, so the checks are cheap:

```sql
select configuration_id, client_mode, count(*)
from santa_scopedclientmode group by 1, 2 having count(*) > 1;

select configuration_id, ltrim(regex, '^'), policy, count(*)
from santa_scopedpathregex group by 1, 2, 3 having count(*) > 1;

select id, configuration_id, length(regex)
from santa_scopedpathregex where length(regex) > 512;
```

## Tests

* The rule engine: the two-rule table above, the strictest policy at equal match with the exclusion as the exception, and a winner change that the ledger sends as a policy change.
* The form and the API: a second rule with the same policy refused with the policy in the message, another policy accepted, a rule on a target with a voting rule refused, and a policy change on the update form to a policy another rule has.
* The rulesets: a policy change is a created and a deleted rule, two policies in one ruleset, a conflict per policy.
* The voting upsert: no voting rule on a target with a non-voting rule, and the voting rule once the rule is gone.
* The scoped items: the two constraints in the forms, the API and the models, the `^` removed and a bare `^` still refused, the name gone from the mode resolution.

## After the first review

* The target page offered an "Add rule" link only to the configurations with no rule on the target, so the second rule of the headline example had no link. It now leaves out only the configurations with a voting rule on the target: the form says which policies are taken.
* The five pick-rule pages kept one rule per identifier in a dictionary over an unordered queryset, so with two rules per target one was dropped and "Update rule" pointed at either. They carry the list now, strictest policy first, with an "Update `<policy>` rule" link per rule and a "Create rule" link next to them.
* The pattern of a scoped path regex is part of a unique index now, and a Postgres btree row holds about 2,700 bytes, so a very long pattern would have been a 500 at save time. The field is a `CharField` of 512 characters: 4 bytes per character at most keep every pattern under the limit with the other columns of the index. The form and the API refuse a longer one, the migration alters the column.

## Docs

A Rule resolution section with the policy table and the two-rule example. The scoped client modes and the scoped path regexes carry their constraint. The ruleset section carries the new key. The changelog has the feature and two backward incompatibility entries.


soysauceconspiracy (approved):
LGTM

PR #1724, opened 2026-09-21
santa: The Rules and the Path regexes tabs of a machine

np5:
PR 3b of the scope and precedence series, after #1723. The machine page becomes a set of tabs, and the two new ones answer the question the page was built for: which rules does this machine have, and which does the next synchronization send.

## One statement for the candidates

The rule download ranks the rules of a configuration for one machine, keeps the winner of each target, and diffs it against the ledger. The page needs the first half of that, and a second implementation of it would be a second answer to "what does this machine get".

`CANDIDATE_RULES_SQL` is that first half, read by both. The download keeps its `DISTINCT ON`; the page ranks with a window function, so it keeps the losers and counts the rules the configuration has for each target. The download tests did not change, which is the evidence that the extraction is faithful.

`MachineRule.objects.rows_for_machine()` merges the candidates with the ledger and gives one row per target:

| State | Meaning | The next synchronization |
| --- | --- | --- |
| On device | The device confirmed the rule of the server, same policy and same version. | sends nothing |
| Not yet on device | The server has a rule the device does not have, or has with another policy or version. A rule sent during the current session stays here until the postflight: Santa writes its database at the end of the download. | sends the rule |
| Still on device | The device has a rule the server does not have anymore. | sends the removal |
| Skipped | The only rules in scope have the `CEL` policy, and the Santa version of the machine cannot evaluate one. | sends nothing |

A clean session is the third case where the ledger and the next batch disagree: the client rebuilds its database from that session alone, so the download ignores the rows it confirmed before and sends every rule of the server again. `rows_for_machine()` mirrors that, or a rule on its way back would read as `On device` for the length of a session.

A skipped rule sorts last in the ranking, so the rule that decides is the one the download picks, and the page says why the CEL rule is not it instead of hiding it.

**The trust test**, in five shapes: the rows in `Not yet on device` and `Still on device` are exactly the identifiers a real `rule_download()` returns.

The state column answers for a synchronization that **starts now**. One that is already running has sent part of it: a rule or a removal sent in an earlier batch stays in its state and is not sent again, because the device holds nothing new for it until the postflight. The table says so, and the tests each start a session before they compare.

## The Rules tab

One row per target. The columns are the target type, the identifier, the policy and the version of the server, what decided for the machine, the policy and the version of the device, and the state. The two policy columns read the same way, so a version that differs is as visible as a policy that differs.

What decided is `Serial number`, `Primary user`, `Tag`, or `All machines` for a rule with no scope field — the wording the scope of an entry already uses. Only a row with no rule on the server decides nothing and gives `-`: a rule the device still has, or one that is skipped. The policy links to the rule that decided, and a target with more than one rule in the configuration gives their number with a link to all of them.

The filters — target type, policy, voting, state — and the identifier search carry the count of each value, over the rows the other filters keep.

## The Path regexes tab

One row for each part of the two patterns Zentral sends: the patterns of the configuration itself, with no name and `Configuration` as what decided, and the entry that won each pattern. The rows are in the order Zentral composes the patterns in, so the table reads like the composition. A policy filter and a free text field on the name or the pattern narrow it.

The entries that lose, and the ones the machine is out of, are not in it: the tab answers what the machine gets, not how every entry was resolved. That is also why this tab needs no `decided_by` annotation, and `ScopedConfigurationItemQuerySet` is untouched.

## What the two tabs give, and to whom

Both tabs answer "what does this machine get", so they draw one line: **what the machine enforces is the state of the machine, and the identity of the object behind it is the configuration.**

`viewEnrolledMachine` therefore gives the policy and the version of every rule, and every pattern, the way the Overview of #1723 gives the client mode. `view_rule` is only needed for the two links of the Rules tab, and `viewScopedPathRegex` only to name an entry and link to it — an entry you cannot view has no name and no link, and its pattern is there.

`ConfigurationRulesView` needs `view_rule` for the same policies and versions, one configuration at a time. This is one machine at a time, and it is that machine's state.

## What it costs

* The table of that tab is not paginated, and it cannot be without a change: the rows of the configuration are positional, first in each policy group because that is where the composition puts them, so a paginator would strand them on an arbitrary page. The patterns in scope for one machine are bounded by the single regex Santa has to compile from them.
* The page no longer shows the composed patterns themselves, so a policy with no part at all — the machine gets a pattern that cannot match — and the rare composition that fails to compile only appear in the logs.

## Tests

* The states, as a matrix of candidate and ledger combinations, staged rows and a removal staged over a staged rule included.
* The trust test: new rules, nothing to send, a removal with a policy change, a skipped rule, and a clean session that puts a confirmed rule back on its way.
* A skipped rule does not decide, and a recent client gets the same rule.
* The tabs: each one reachable, PBAC allowed and denied, the machine that is not enrolled.
* The Rules tab: the columns, the two links, every filter and its counts, the identifier search, pagination, and the two rows that decide nothing.
* The Path regexes tab: the pattern of the configuration alone, the winner of each pattern, the policy filter with its counts, the search on a name and on a pattern, and an entry whose name is hidden without the permission while its pattern stays.

The full suite passes, and every line of the patch is covered.


sebastian-zentral (dismissed):

soysauceconspiracy (approved):
LGTM

